PR #8, opened 2019-09-25
Feature/non local build structure

ivcosla:
This PR aims to move every binary build as well as integration scripts outside of the project root directory in order to have a better workspace and make developers less likely to add by error some undesired binaries to git history.


jdknives:
Superseded by #41

PR #11, opened 2019-10-02
Remove transport from discovery on transport deregister

### LINKED ISSUE #10: Add ability to delete entries in transport discovery.
**Feature description**

Currently, one can only update statuses of entries in transport discovery and cannot actually delete entries.

This causes issues when restarting transports. When we remove transports from visor, we only remove them locally. On visor startup, it polls transport discovery to determine saved transports. Hence, locally deleted transports are revived.

**Implementation Details:**

- A transport between nodes A and B can only be deleted by A or B.

**Tasks:**

- [ ] Add endpoint to transport discovery to delete transport.
- [ ] Update transport discovery client.
- [ ] Update `(transport.Manager).DeleteTransport` to call transport discovery's delete transport endpoint.

ivcosla:
Fixes #10 10

 Changes:	
- Added `DeleteTransport` to client as well as to the mock	

How to test this PR:
Either `make test` or integration test on `skywire-services` using the branch of the same name.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -35,4 +35,4 @@ require (
 )
 
 // Uncomment for tests with alternate branches of 'dmsg'
-// replace github.com/SkycoinProject/dmsg => ../dmsg
+replace github.com/SkycoinProject/dmsg => ../dmsg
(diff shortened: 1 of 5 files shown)


ivcosla:
Note that I implemented it as specified in the issue, however, at the moment (transport.Manager).DeleteTransport is not being called anywhere in the code other than by `skywire-cli` on the `DeleteTransport` RPC method.

nkryuchkov:
Replaced by #54

PR #46, opened 2019-11-06
Invalidate Hypervisor session after password change.

### LINKED ISSUE #18: Invalidate the Hypervisor session after changing the password
**Feature description**

Describe the feature
If the Hypervisor is ever going to be used remotely by various users, changing the password should act as a way for instantly blocking access to unauthorized users at any time. Invalidate all sessions of the user should be part of that.

After changing the password o fan account, unauthorized users would still have access as long as they keep an earlier session open.

All sessions of a particular account should be invalidated just after changing its password.

**Additional context**
This is not critical, as system administrators can delete all sessions by restating the Hypervisor, but would be more convenient.

**Possible implementation**

evanlinjin:
Fixes #18

### DIFF
--- a/cmd/skywire-visor/commands/root.go
+++ b/cmd/skywire-visor/commands/root.go
@@ -1,5 +1,6 @@
 package commands
 
+// NOTE: "net/http/pprof" is used for profiling.
 import (
 	"bufio"
 	"context"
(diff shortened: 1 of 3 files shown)


jdknives:
@evanlinjin please remove the binary

PR #51, opened 2019-11-07
[WIP] Fix app2, router2 tests

### LINKED ISSUE #50: Overall testing of app2 and router2
Need to fix tests for the app2 and router2 modules. Need to do a perform a thorough testing of these modules

Darkren:
Fixes #50 

 Changes:	
- Fixed packages:
-- app
-- router
-- visor
-- pty
-- hypervisor
-- setup
- Fixed tests:
-- visor
-- visor RPC
- Temporarily commented out tests:
-- router
-- setup node


pkg/router/route_group.go
```diff
@@ -46,64 +81,134 @@ type RouteGroup struct {
 	done    chan struct{}
 	once    sync.Once
 
-	rt routing.Table
-
-	lastSent int64
+	readTimer  *time.Timer
```
evanlinjin: For the timeouts, we can use https://golang.org/src/net/pipe.go as @nkryuchkov has pointed out here: https://discordapp.com/channels/466450852098605059/466450852098605061/652522206290771969

pkg/router/route_group.go
```diff
@@ -46,64 +81,134 @@ type RouteGroup struct {
...
 }
 
 // Read reads the next packet payload of a RouteGroup.
-// The Router, via transport.Manager, is responsible for reading incoming packets and pushing it to the appropriate RouteGroup via (*RouteGroup).readCh.
-// To help with implementing the read logic, within the dmsg repo, we have ioutil.BufRead, just in case the read buffer is short.
+// The Router, via transport.Manager, is responsible for reading incoming packets and pushing it
+// to the appropriate RouteGroup via (*RouteGroup).readCh.
+// To help with implementing the read logic, within the dmsg repo, we have ioutil.BufRead,
+// just in case the read buffer is short.
+// TODO: too long, simplify
```
evanlinjin: Yes, the logic here needs simplifications, and we need to use the `bytes.Buffer` just in case we have short reads.

pkg/router/route_group.go
nkryuchkov: @evanlinjin I agree, just pushed :)

pkg/router/route_group.go
nkryuchkov: Commented it out to ensure the code without it works as expected. Now it does, `bytes.Buffer` is back.

PR #53, opened 2019-11-07
Reconnect proxy if yamux session failed

### LINKED ISSUE #52: Proxy client does not work if yamux connection fails
If an intermediary node is restarted or if a network error happens, the proxy client does not reconnect and stops working. We need to implement reconnection logic.

nkryuchkov:
Fixes #52

Changes:	
- [x] Implement reconnection 
- [x] Adjust broken unit tests for the new implementation

How to test this PR:
Run e2e tests in `skywire-services`.

### DIFF
--- a/internal/therealproxy/client.go
+++ b/internal/therealproxy/client.go
@@ -4,9 +4,14 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"time"
 
 	"github.com/SkycoinProject/skycoin/src/util/logging"
 	"github.com/hashicorp/yamux"
+
+	"github.com/SkycoinProject/skywire-mainnet/internal/netutil"
+	"github.com/SkycoinProject/skywire-mainnet/pkg/app"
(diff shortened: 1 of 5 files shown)


internal/therealproxy/client.go
```diff
@@ -16,20 +21,110 @@ var Log = logging.MustGetLogger("therealproxy")
...
+			}
+		}
+	}()
 }
 
-// ListenAndServe start tcp listener on addr and proxies incoming
+// ListenAndServe starts tcp listener on addr and proxies incoming
 // connection to a remote proxy server.
+// TODO: get rid of it
 func (c *Client) ListenAndServe(addr string) error {
```
Darkren: I think we should get rid of this. Since we're passing listener to the client's contructor and have a separate `Serve` func, this `ListenAndServe` makes things confusing

nkryuchkov: @Darkren Good catch. But not sure if it's worth for milestone 1 because the new app is going to be integrated into milestone 2, so the proxy client will require modifications anyway.

PR #54, opened 2019-11-07
Remove transport from discovery on transport deregister

nkryuchkov:
Replaces #11
To be merged simultaneously with https://github.com/SkycoinPro/skywire-services/pull/31

### DIFF
--- a/pkg/transport/manager.go
+++ b/pkg/transport/manager.go
@@ -214,7 +214,7 @@ func (tm *Manager) saveTransport(remote cipher.PubKey, netName string) (*Managed
 	return mTp, nil
 }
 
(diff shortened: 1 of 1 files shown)


pkg/transport/manager.go
```diff
@@ -224,7 +225,15 @@ func (tm *Manager) DeleteTransport(id uuid.UUID) {
 	if tp, ok := tm.tps[id]; ok {
 		tp.Close()
 		delete(tm.tps, id)
-		tm.Logger.Infof("Unregistered transport %s", id)
+		tm.Logger.Infof("Unregistered transport %s from manager", id)
```
evanlinjin: Update the method's comment to include: "De-register transport".

PR #41, opened 2019-10-23
Feature/temp build dirs

jdknives:
Fixes #	

- creates temporary build directories
- removes Makefile goals for Docker environment and `/integration` directory since that has been moved to skywire-services.

Note: This is a copy from ivcoslas earlier PR without the merge conflicts
He wrote originally: This PR aims to move every binary build as well as integration scripts outside of the project root directory in order to have a better workspace and make developers less likely to add by error some undesired binaries to git history.

PR #61, opened 2019-11-11
Tmp build dir

jdknives:
Fixes #41  

 Changes:	
- removes integration dir because it was moved to Skywire-services
- added support for moving build binaries to tmp directory outside of the skywire workspace


evanlinjin:
@jdknives Is this still needed>?

PR #63, opened 2019-11-11
Fix wrong use of middleware context in hypervisor.

### LINKED ISSUE #58: It is not possible to delete transports and routes using the hypervisor API
**Describe the bug**
In the master branch, when calling the `DELETE /api/nodes/{pk}/transports/{tid}` and `DELETE /api/nodes/{pk}/routes/{routeId}` API endpoints for deleting transports and routes, the operation fails.

**Environment information:**

- OS: Linux (Ubuntu 18.04.1)

**Steps to Reproduce**

1. Start some nodes using the `make integration-run-generic` command of the `skywire-services` repository.
2. Create transports and routes.
3. Try to delete the transports and routes using the `DELETE /api/nodes/{pk}/transports/{tid}` and `DELETE /api/nodes/{pk}/routes/{routeId}` API endpoints.

**Actual behavior**
When trying to delete the transports and routes, the operation always fails and the response is something similar to `{"error":"can not find app of name  from node 024ec47420176680816e0406250e7156465e4531f5b26057c9f6297bb0303558c7"}`

**Expected behavior**
The transports and routes should be deleted.

**Additional context**

**Possible implementation**

evanlinjin:
Fixes #58 

 Changes:	
- Fix wrong use of middleware context in hypervisor.

### DIFF
--- a/pkg/hypervisor/hypervisor.go
+++ b/pkg/hypervisor/hypervisor.go
@@ -650,7 +650,7 @@ func (m *Node) appCtx(w http.ResponseWriter, r *http.Request) (*httpCtx, bool) {
 }
 
 func (m *Node) tpCtx(w http.ResponseWriter, r *http.Request) (*httpCtx, bool) {
-	ctx, ok := m.appCtx(w, r)
+	ctx, ok := m.nodeCtx(w, r)
 	if !ok {
 		return nil, false
 	}


Senyoret1:
@evanlinjin This appears to be working well, I was able to remove transports without issues. However, there is a problem with the routes: until now I have been removing the routes by calling `DELETE /api/nodes/{pk}/routes/{routeId}` sending as the `routeId` param the value named `key` returned by the `GET /api/nodes/{pk}/routes` API endpoint, but that appears to not be working anymore. When I do that, the `DELETE /api/nodes/{pk}/routes/{routeId}` API endpoint returns `{"error":"invalid UUID length: 0"}`, so I think it is now expecting a different value, the problem is that I don't know were to get that value from, as the `GET /api/nodes/{pk}/routes` API endpoint does not return it, or at least not in a evident way.

Maybe a new issue should be filed for this problem, but I don't know what the sollution should be, if adding the param to the `GET /api/nodes/{pk}/routes` API endpoint, or making the `DELETE /api/nodes/{pk}/routes/{routeId}` acept the `key` param returned by `GET /api/nodes/{pk}/routes`, or something else.

PR #70, opened 2019-11-14
change from slice to map

### LINKED ISSUE #67: Use maps instead of slices where lookups often happen
1. https://github.com/SkycoinProject/skywire-mainnet/blob/master/pkg/visor/visor.go#L106

Not a part of your work, but what do you think of changing the type of `appsConf` from slice to hashmap? The most often thing we do is a lookup, which would be faster in that case.

_Originally posted by @nkryuchkov in https://github.com/SkycoinProject/skywire-mainnet/pull/62_

pikomonde:
Fixes #67 

Changes:	
- change some of previous declaration from slice of AppConfig into hashmap. 

How to test this PR:
- haven't run the unit test
- should we do benchmarking to know that is it really faster using these map?

### DIFF
--- a/pkg/visor/visor.go
+++ b/pkg/visor/visor.go
@@ -103,7 +103,7 @@ type Node struct {
 
 	appsPath  string
 	localPath string
-	appsConf  []AppConfig
+	appsConf  map[string]AppConfig
 
 	startedMu   sync.RWMutex
 	startedApps map[string]*appBind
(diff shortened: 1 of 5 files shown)


pkg/visor/visor.go
```diff
@@ -89,7 +89,7 @@ type Node struct {
 
 	appsPath  string
 	localPath string
-	appsConf  []AppConfig
+	appsConf  map[int]AppConfig
```
nkryuchkov: What do you think about changing the type to `map[string]AppConfig` where the key would be the app name? This way we could get an app by its name without iterating all apps.

pkg/visor/visor.go
```diff
@@ -401,18 +401,16 @@ func (node *Node) Apps() []*AppState {
...
-			go func(app AppConfig) {
-				if err := node.SpawnApp(&app, startCh); err != nil {
-					node.logger.Warnf("Failed to start app %s: %s", appName, err)
-				}
-			}(app)
-
-			<-startCh
-			return nil
-		}
+	if app, isAppExist := node.appsConf[appName]; isAppExist {
```
nkryuchkov: I would prefer to return early if an app doesn't exist to reduce indentation. We can also rename `isAppExist` to `ok` because it's a common way in Go to check if an element exists in a collection.
```golang
app, ok := node.appsConf[appName]
if !ok {
	return ErrUnknownApp
}

// start app
```

pikomonde: yeah, agree.. I was gonna use ok back then, but hestitate, because of the readiblity for programmer that not familiar with go.. i'll change it in a minute..

nkryuchkov: No problem, thanks

PR #77, opened 2019-11-27
Feature/default stcp listen

### LINKED ISSUE #71: Make visor listen for stcp transports automatically
**Feature description**

Get local IP address and make visor listen for `stcp` transports on that address and port `7777` by default. 

Currently, we have to manipulate the configuration file manually to setup `stcp` transports and set the local IP and port to listen on. All of that should be automated.

jdknives:
Fixes #71 

 Changes:	
-	add function to get local IP address

How to test this PR:

Build, install and run 

`skywire-cli node gen-config`. Check whether the field for local address is correctly set.


cmd/skywire-cli/commands/node/gen-config.go
```diff
@@ -186,3 +189,19 @@ func defaultSkyproxyClientConfig() visor.AppConfig {
 		Port:      routing.Port(skyenv.SkyproxyClientPort),
 	}
 }
+
+func getLocalIPAddress() string {
+	addrs, err := net.InterfaceAddrs()
+	if err != nil {
+		log.WithError(err)
```
nkryuchkov: I would return an error here and not continue function execution in that case.

cmd/skywire-cli/commands/node/gen-config.go
```diff
@@ -186,3 +189,19 @@ func defaultSkyproxyClientConfig() visor.AppConfig {
...
+	}
+
+	for _, a := range addrs {
+		if ipnet, ok := a.(*net.IPNet); ok && !ipnet.IP.IsLoopback() {
+			if ipnet.IP.To4() != nil {
+				return ipnet.IP.String() + ":7777"
```
nkryuchkov: Could you please add a comment with a short desctiption why `":7777"` is used here?

cmd/skywire-cli/commands/node/gen-config.go
```diff
@@ -186,3 +189,19 @@ func defaultSkyproxyClientConfig() visor.AppConfig {
...
+
+func getLocalIPAddress() string {
+	addrs, err := net.InterfaceAddrs()
+	if err != nil {
+		log.WithError(err)
+	}
+
+	for _, a := range addrs {
+		if ipnet, ok := a.(*net.IPNet); ok && !ipnet.IP.IsLoopback() {
+			if ipnet.IP.To4() != nil {
```
nkryuchkov, line 207: I'm not sure if this implementation is correct, but it looks like it does not work with IPv6. Probably this line should look like this:
```suggestion
			if ipnet.IP.To4() != nil || ipnet.IP.To16() != nil {
```
However, the suggestion needs to be tested.

cmd/skywire-cli/commands/node/gen-config.go
```diff
@@ -186,3 +189,19 @@ func defaultSkyproxyClientConfig() visor.AppConfig {
...
+				return ipnet.IP.String() + ":7777"
+			}
+		}
+	}
+	return ""
```
nkryuchkov: I would return an error in case an address cannot be found.

cmd/skywire-cli/commands/node/gen-config.go
```diff
@@ -186,3 +189,19 @@ func defaultSkyproxyClientConfig() visor.AppConfig {
 		Port:      routing.Port(skyenv.SkyproxyClientPort),
 	}
 }
+
+func getLocalIPAddress() string {
```
nkryuchkov: Probably we need to test the function somehow. It looks like it does what we need, but I'm not sure it would work correctly in all cases.

cmd/skywire-cli/commands/node/gen-config.go
jdknives: It is a random port in the 1025-49152 range. It does not clash with any widely used protocol or application according to [Wikipedia](https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers) and it is easy enough to remember.

cmd/skywire-cli/commands/node/gen-config.go
jdknives: It does not matter to me whether we use IPv4 or IPv6 as long as we are using only one of these by default consitently, imo.

cmd/skywire-cli/commands/node/gen-config.go
jdknives: I am pretty sure this will not work in all cases as users may have non-standard network interfaces on their machines and fairly complex setups. This is just meant as a rough measure to cover most of the standard cases automatically. If you see any big areas that we should check for though, please let me know.

nkryuchkov: I think it's fine then.

nkryuchkov (approved):

PR #80, opened 2019-12-17
Implement visor restart from hypervisor

### LINKED ISSUE #64: Endpoint for restarting visor from hypervisor
**Feature description**
The hypervisor needs to be able to restart the visor because it's complicated to restart a big amount of visors manually.

We can create a new visor process with the same args and kill the current one. 

At the visor start:
1. Save the current working directory.
2. Save the command line args: the relative binary path, binary args

At the visor restart:
1. Since args may contain relative paths, either set the working directory to the saved one, or change relative paths in args to absolute ones using the saved working directory.
2. Run a new visor process with the required args.
3. Kill the current process.

When going back to the beginning of `main()`, it might be very complicated to close all open resources.

**Additional context**
Running a new visor and killing the current one may need further study whether there are any pitfalls. E.g. we need to ensure that all resources needed by the new visor are freed by the old one before the start of the new one.

nkryuchkov:
Fixes #64 

 Changes:	
- Implement visor restart from hypervisor. It can be used via `/nodes/{pk}/restart` endpoint.

How to test this PR:

1. Unit tests
- Run added unit tests for `./pkg/restart/`

2. Integration tests
- Run the generic integration environment 
- Run `ps aux | grep visor`
- Check visors' PIDs
- In `shell` pane run `curl http://localhost:8080/api/nodes/024ec47420176680816e0406250e7156465e4531f5b26057c9f6297bb0303558c7/restart` 
- In `NodeA` pane make sure visor was restarted
- Run `ps aux | grep visor`
- Make sure that `NodeA`'s PID differs


pkg/restart/restart.go
```diff
@@ -0,0 +1,151 @@
...
+	// ErrMalformedArgs is returned when executable args are malformed.
+	ErrMalformedArgs = errors.New("malformed args")
+)
+
+// DefaultCheckDelay is a default delay for checking if a new instance is started successfully.
+const DefaultCheckDelay = 5 * time.Second
+
+// Context describes data required for restarting visor.
+type Context struct {
+	log              *logging.Logger
```
evanlinjin: I prefer to use `logrus.FieldLogger` here to allow more possible structures.

pkg/visor/rpc.go
```diff
@@ -390,3 +393,16 @@ func (r *RPC) Loops(_ *struct{}, out *[]LoopInfo) error {
...
 	return nil
 }
+
+/*
+	<<< VISOR MANAGEMENT >>>
+*/
+
+// Restart restarts visor.
+func (r *RPC) Restart(_ *struct{}, _ *struct{}) error {
+	if r.node.restartCtx == nil {
```
evanlinjin, line 415: Should we protect this with a `sync.Mutex`? As it states in the `(*restart.Context).Restart()` method comment: 
```golang
// Should not be called from a goroutine.
```

evanlinjin (commented):
Good work!

pkg/visor/rpc.go
nkryuchkov: @evanlinjin You're right, thanks.

Darkren (approved):
Great job!

PR #87, opened 2019-12-24
Add keep-alive packet propagation between nodes

### LINKED ISSUE #81: [M2] Keep-alive packet not being propagated through all the route
**Describe the bug**
When node receives keep-alive packet, it update activity of the rule associated with the route. It works fine. But the packet is not being propagated forward the route. If we have, say, 3 nodes, where one of them is intermediary, there won't be any trouble. But if we add more intermediary nodes, they won't get this keep-alive packet. This way the route will get broken as soon as the data packets stop going through the network.

**Actual behavior**
Without transmitting data packets, having 2+ intermediary nodes the route gets broken after rule timeout

**Expected behavior**
Keep-alive packets are being handled by all the nodes along the route, therefore updating rule activity and preventing rules from being removed

**Possible implementation**
There's func handling keep-alive packet. We need to just forward the packet down the line, should be easy to implement

Darkren:
Fixes #81 

 Changes:	
- Proper keep-alive packets propagation was added. Now when router receives the keep-alive packet, assuming that we talk about the intermediary node, the packet is being forwarded to the next node along the route. This way all the rules have their activity up to date, preventing unwanted expiry

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -24,4 +24,4 @@ require (
 	golang.org/x/net v0.0.0-20191204025024-5ee1b9f4859a
 )
 
-// replace github.com/SkycoinProject/dmsg => ../dmsg
+//replace github.com/SkycoinProject/dmsg => ../dmsg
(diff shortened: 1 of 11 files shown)


nkryuchkov (approved):
Great job!

PR #88, opened 2019-12-24
[WIP] [M2] Fix transport manager tests

### LINKED ISSUE #82: [M2] Transport manager tests fail
**Describe the bug**
Two transport manager tests are failing randomly. Sometimes they pass, sometimes they don't. The failing tests are:
- `TestNewManager/check_read_write` with error:
image
Which happens here:
image

- `TestNewManager/check_tp_logs` with error:
image
Which happens here:
image
ALSO, this test sometimes fails with the different error:
image
Which happens on these lines:
image

These 2 are probably connected and happen because of a single bug

- Also test sometimes hangs

Darkren:
Fixes #82 

 Changes:	
- Added wait group to sync test cases

### DIFF
--- a/pkg/transport/manager_test.go
+++ b/pkg/transport/manager_test.go
@@ -5,18 +5,21 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"sync"
 	"testing"
 	"time"
 
-	"github.com/SkycoinProject/skycoin/src/util/logging"
+	"github.com/stretchr/testify/assert"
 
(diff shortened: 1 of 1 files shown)


Darkren:
For now everything is the same. Sometimes it fails for different reasons, sometimes it hangs. I tried adding wait group, probably it doesn't have any effect. From what I've found - there's `EOF` coming from DMSG connection which sometimes causes tests to fail. Hanging is probably connected to this too. I see no point in trying to figure this out for now, since we're waiting for the new version anyway

PR #89, opened 2019-12-25
[M2] Fix default route group timeout

### LINKED ISSUE #83: [M2] Default route group I/O timeout
**Describe the bug**
The default I/O timeout for route group is hard coded now into constant. If we try to fully satisfy the `net.Conn`, we should remove the default timeout, since most of client applications expect such  behavior

Darkren:
Fixes #83 

 Changes:	
- Close packets are now being propagated between two route groups in a loop
- Added proper handling of close packets by route group
- Added correct EOF from the route group's `Read`


go.mod
```diff
@@ -24,4 +24,4 @@ require (
 	golang.org/x/net v0.0.0-20191204025024-5ee1b9f4859a
 )
 
-//replace github.com/SkycoinProject/dmsg => ../dmsg
+replace github.com/SkycoinProject/dmsg => ../dmsg
```
nkryuchkov: Could you please comment it out?

nkryuchkov (approved):
Great job

go.mod
Darkren: Done

Darkren:
Some of the subtests of `TestConn` for route group are still failing/hanging

pkg/router/route_group.go
```diff
@@ -219,20 +226,14 @@ func (rg *RouteGroup) Write(p []byte) (n int, err error) {
 	}
 }
 
-func (rg *RouteGroup) writePacketAsync(tp *transport.ManagedTransport, packet routing.Packet) (chan error, func()) {
-	ctx, cancel := context.WithCancel(context.Background())
-
+func (rg *RouteGroup) writePacketAsync(ctx context.Context, tp *transport.ManagedTransport, packet routing.Packet) chan error {
 	errCh := make(chan error)
```
evanlinjin, line 230: Should we initialize this with size of 1 so the `WritePacket` go routine does not hang?

pkg/router/route_group.go
```diff
@@ -259,42 +260,25 @@ func (rg *RouteGroup) tp() (*transport.ManagedTransport, error) {
...
 
-	rules := rg.rt.RulesWithDesc(rg.desc)
-	routeIDs := make([]routing.RouteID, 0, len(rules))
-
-	for _, rule := range rules {
-		routeIDs = append(routeIDs, rule.KeyRouteID())
+	if rg.isRemoteClosed() {
+		// remote already closed, everything is cleaned up,
+		// we just need to close signal channel at this point
+		close(rg.closed)
```
evanlinjin, line 272: How have we guaranteed that we won't have a race condition on line 272?

Darkren: @evanlinjin it's executed within `sync.Once`

pkg/router/route_group.go
Darkren: @evanlinjin I'll think a bit

Darkren: @evanlinjin the only reason for having a err chan here is to be able to do a select between it and the context chan. so, the err chan gets contantly listened to outside of `writePacketAsync` and this way we really want to block on this

nkryuchkov (approved):
Great job! Please don't forget to fix linter errors.

evanlinjin (approved):

Kifen (approved):
Good work!

PR #90, opened 2019-12-26
Fix default visor/hypervisor start

### LINKED ISSUE #22: Multiple errors running the hypervisor and visor on mainnet-milestone2
**Describe the bug**
Using the `mainnet-milestone2` branch, the hypervisor does not run. Also, the visor has problems connecting to the dmsg server.

**Environment information:**

- OS: Linux (Ubuntu 18.04.1)

**Steps to Reproduce**

1. Run `make install` using the `mainnet-milestone2` branch.

2. Run `hypervisor gen-config -o hypervisor-config.json`, to create a default configuration file for the hypervisor.

3. Run `hypervisor`. It will fail with `Failed to parse rpc port from rpc address: parse :7080: missing protocol scheme`.

4. Open the configuration file created in the second step and change the value of `rpc_addr` to `localhost:7080`.

5. Run `hypervisor` again. It will fail with `Failed to parse rpc port from rpc address: strconv.ParseUint: parsing "": invalid syntax`.

6. Open the configuration file created in the second step again and change the value of `rpc_addr` to `http://skycoin.net:7080`.

7. Run `hypervisor` again. The console will start displaying `no dms_servers found: trying again in 1s... error="Get /dmsg-discovery/available_servers: unsupported protocol scheme """`. The problem with the rpc address appears to be solved.

8. Open the configuration file created in the second step again and change the value of `dmsg_discovery` to `https://messaging.discovery.skywire.skycoin.net`.

9. Run `hypervisor` again. The console will start displaying `no dms_servers found: trying again in 1s... error="Get https://messaging.discovery.skywire.skycoin.net/dmsg-discovery/available_servers: dial tcp: lookup messaging.discovery.skywire.skycoin.net: no such host"`.

In a similar way, trying to run a visor with the default config results in `no dms_servers found: trying again in 1s... error="json: cannot unmarshal number into Go value of type disc.HTTPMessage"` being displayed every second in the console.

**Actual behavior**
The errors previously described appear. Also, the hypervisor is not connecting with the visors, as it does in the `master` branch.

**Expected behavior**
The hypervisor and the visor should work like in the `master branch`, with the default config. If it is currently necessary to run a local dmsg server for making the visor and hypervisor work, it would be good to add instructions in the readme, and it could be also useful in other cases.

**Additional context**

**Possible implementation**

Darkren:
Fixes #22 

 Changes:	
- SSH references were removed from Makefile
- Default config now doesn't have SSH apps
- Updated vendor to use latest version of DMSG from M2

How to test this PR:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,9 +31,9 @@ config: ## Generate skywire.json
 
 clean: ## Clean project: remove created binaries and apps
 	-rm -rf ./apps
-	-rm -f ./skywire-visor ./skywire-cli ./setup-node ./hypervisor ./SSH-cli
+	-rm -f ./skywire-visor ./skywire-cli ./setup-node ./hypervisor
 
-install: ## Install `skywire-visor`, `skywire-cli`, `hypervisor`, `SSH-cli`, `dmsgpty`
+install: ## Install `skywire-visor`, `skywire-cli`, `hypervisor`, `dmsgpty`
 	${OPTS} go install ./cmd/skywire-visor ./cmd/skywire-cli ./cmd/setup-node ./cmd/hypervisor ./cmd/therealssh-cli ./cmd/dmsgpty
 
 rerun: stop
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ clean: ## Clean project: remove created binaries and apps
 	-rm -f ./skywire-visor ./skywire-cli ./setup-node ./hypervisor
 
 install: ## Install `skywire-visor`, `skywire-cli`, `hypervisor`, `dmsgpty`
-	${OPTS} go install ./cmd/skywire-visor ./cmd/skywire-cli ./cmd/setup-node ./cmd/hypervisor ./cmd/therealssh-cli ./cmd/dmsgpty
+	${OPTS} go install ./cmd/skywire-visor ./cmd/skywire-cli ./cmd/setup-node ./cmd/hypervisor ./cmd/dmsgpty
 
 rerun: stop
 	${OPTS} go build -race -o ./skywire-visor ./cmd/skywire-visor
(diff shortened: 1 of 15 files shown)


Darkren:
Hypervisor seems to work. Visor does not. The main problem is that it tries to access the current production DMSG discovery which works for the master branch version, but not for M2. API has changed. There's no point in dancing with the API URLs since the functionality might have been changed too, we're trying to connect two different versions here

Darkren:
@Senyoret1 
Added README files with instructions on how to build and run the DMSG server and discovery locally in here https://github.com/SkycoinPro/skywire-services/pull/40 . In order to run visor, you will need to specify the correct address of your DMSG discovery in the default generated visor config. This is a temporary workaround, I don't see a point in mentioning this in the common skywire README. 

The system seems to be running, however hypervisor doesn't see the running visor. There's ongoing work on this, I'll notify on the result

Darkren:
@Senyoret1 actually it seems to work. Please check it out when you have a minute

Senyoret1:
I tested the visor and hypervisor and found errors while trying to make the manager UI work with the `mainnet-milestone2` brach, which I think was deleted. At this moment I'm not familiar with the changes made to the `milestone2`, so maybe it won't be of much help if I test this PR, as it will take me some time just to check how things work in that branch (specially for some changes I think were made to the integration with dmsg) and I won't know exactly what to check (some of the issues I filed should not be valid anymore). Maybe it would be better if I take a look when the `milestone2` is more mature, to start integrating the manager UI with it. However, if you want me to check it anyways, let me know

Darkren:
@Senyoret1 if that's the case, I think there's no point for now. I'll get it merged and when `m2` gets ready I'll double check that everything works

PR #104, opened 2020-01-02
Integrate skywire-peering-daemon to mainnet

Kifen:
Fixes #	

 Changes:	
-	Added method to startup skywire-peering-daemon as an external process, and receive packets 
       from the daemon over named pipes
-    Added option to visor config to disable skywire-peering-daemon by default
-   Updated skywire-mainet's `make build` command to build skywire-peering-daemon and copy its 
     binary to ./skywire-mainnet

How to test this PR:
-    Ensure a skywire-services integration env is set 
-    Clone skywire-peering-daemon to a  directory adjacent to skywire-mainnet & skywire-services.
-    Add option  {"auto_start_spd": true} to `./skywire- 
       services/integration/generic/{nodeC.json/nodeA.json}`  to enable the daemon
-    
       # run skywire-services
       $  make integration-build 
       $  make integration-run-generic


pkg/visor/daemon.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"os"
+	"os/exec"
+	"strconv"
+	"strings"
+
+	"github.com/SkycoinProject/skywire-mainnet/pkg/snet"
+
+	"github.com/SkycoinProject/dmsg/cipher"
+	cp "github.com/SkycoinProject/skycoin/src/cipher"
```
nkryuchkov: I would leave it as `cipher`. `cp` is a bit confusing.

pkg/visor/daemon.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/SkycoinProject/skycoin/src/util/logging"
+	apd "github.com/SkycoinProject/skywire-peering-daemon/src/daemon"
```
nkryuchkov: I think `spd` would look better.

pkg/visor/visor.go
```diff
@@ -282,6 +288,75 @@ func (node *Node) Start() error {
...
+		for {
+			var (
+				packet apd.Packet
+				buff   bytes.Buffer
+			)
+
+			<-c
+			_, err = io.Copy(&buff, stdOut)
+			if err != nil {
+				node.logger.Fatal(err)
```
nkryuchkov: Is it intended behavior that `visor` fails if `daemon` fails to start?

pkg/visor/daemon.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+var logger = logging.MustGetLogger("skywire-peering-daemon")
+
+func execute(binPath, publicKey, namedPipe string) error {
```
nkryuchkov: If the `daemon` should be started from the `visor`, maybe it would be more elegant to have its code in `skywire-mainnet`. What do you think?

nkryuchkov (commented):
Good job. 

I would not commit the `skywire-peering-daemon` executable, we can add it to `.gitignore`

nkryuchkov (commented):
We will also need to comment out the replace with `skywire-peering-daemon` in `go.mod`, just a reminder :) Why do we need `go.mod427131847.tmp`?

pkg/visor/visor.go
```diff
@@ -282,6 +288,75 @@ func (node *Node) Start() error {
...
+	}
+
+	node.logger.Info("Opening named pipe for reading packets from skywire-peering-daemon")
+	stdOut, err := os.OpenFile(namedPipe, os.O_RDONLY, 0600)
+	if err != nil {
+		node.logger.Fatal(err)
+	}
+
+	c := make(chan notify.EventInfo, 5)
+	notify.Watch(namedPipe, c, notify.Write)
```
nkryuchkov: We need to check a possible error.

pkg/visor/daemon.go
Kifen: Both `github.com/SkycoinProject/dmsg/cipher` and 	 `github.com/SkycoinProject/skycoin/src/cipher`
are imported in the same file.

How about `skycoin_cipher `github.com/SkycoinProject/skycoin/src/cipher` ?

pkg/visor/daemon.go
Kifen: Yeah. Changed.

pkg/visor/daemon.go
nkryuchkov: @Kifen Oh, I haven't seen that. Sorry. Maybe `skycoincipher` and `dmsgcipher`? (Golang guidelines forbid the underscore)

pkg/visor/visor.go
Kifen: Checked

pkg/visor/visor.go
Kifen: Not at all. A daemon failing shouldn't also fail the visor, rather the daemon process should be stopped.

nkryuchkov: @Kifen `Fatal()` calls `os.Exit(1)` and exits the process. So I think we should use `Error()` with `return`.

nkryuchkov: @Kifen I would also unlock the mutex with `defer` right after locking to ensure it's unlocked under any condition.

pkg/visor/daemon.go
Kifen: That would work, but the implementation requires that the daemon be in a separate repo.

PR #108, opened 2020-01-07
integrate skywire-peering-daemon to mainnet

Kifen:
Changes:	
-     Added method to startup skywire-peering-daemon as an external process, and receive packets 
      from skywire-peering-daemon over named pipes
-    Added option to visor config to disable skywire-peering-daemon by default
-    Updated skywire-mainet's make build command to build skywire-peering-daemon and copy its 
     binary to ./skywire-mainnet

How to test this PR:
-    Ensure a skywire-services integration env is set 
-    Clone skywire-peering-daemon to a  directory adjacent to skywire-mainnet & skywire-services.
-    Add option  {"auto_start_spd": true} to ./skywire- 
     services/integration/generic/{nodeC.json/nodeA.json}  to enable the daemon
-    
       # run skywire-services
       $  make integration-build 
       $  make integration-run-generic


jdknives:
We can proceed with this PR now that the services for m2 are up and we can test in production.

I setup two machines in the same network with the daemon on `master` and skywire on `milestone2` for this test. This is what I got. 

1. We should change the logging of the daemon, as it clutters the output. 

2. I got: 
 
```
[2020-01-24T22:15:01+01:00] INFO [SPD]: Received packet 0309ca7e679b9bd43a4545463565f149728787b0428e76200786b9f3dce8be9e24: 192.168.2.106:7777
[2020-01-24T22:15:01+01:00] INFO [apd.broadcast]: Sending packet over pipe
panic: assignment to entry in nil map

goroutine 115 [running]:
github.com/SkycoinProject/skywire-mainnet/pkg/visor.(*Node).RunDaemon.func1(0xc00014e960, 0xc0000100f0, 0xc000134830, 0xc0001d4000)
	/Users/rudikleine/go/src/github.com/jdknives/skywire-mainnet/pkg/visor/visor.go:369 +0x39c
created by github.com/SkycoinProject/skywire-mainnet/pkg/visor.(*Node).RunDaemon
	/Users/rudikleine/go/src/github.com/jdknives/skywire-mainnet/pkg/visor/visor.go:347 +0x811
```

Also I think the logging should be consistent (so either [apd.broadcast] or [SPD]. But I may be wrong on that.

Kifen:
@jdknives I'll fix the error and the conflicts. Also, you're right, the logging should be consistent, [SPD] will work. @Darkren @nkryuchkov what do you think?

PR #113, opened 2020-01-09
Remove socket files

### LINKED ISSUE #93: [M2] Socket files are not removed on visor shutdown
**Describe the bug**
Socket files (app serve and dmsgpty) are not removed on visor shutdown. This way if we restart visor, it won't run because of `address already in use`

**Environment information:**
Independent

**Steps to Reproduce**

1. Run skywire visor
2. Shutdown visor
3. Run it again

**Actual behavior**
Visor fails to run because of `address already in use`

**Expected behavior**
Visor runs

**Possible implementation**
Remove socket files on visor shutdown

Kifen:
Fixes #93 	

 Changes:	
-	Added method to unlink socket files.
-      Checks on startup of a node if socket file exists, if it does, remove it

How to test this PR:
-     Run an integration env
-     Stop a node with Ctrl-c or close the env via tmux kill-session -t skywire
-     Restart the env, error bind address already in use doesn't occur


jdknives:
@Kifen binary file uploaded.

jdknives:
Also the replace should be re-commentend.

Darkren (approved):
Great one! But if we use `syscall` here anyway, what do you think about getting rid of `os.Stat` call? We may just call `Unlink` and filter out `ENOENT` error in case file doesn't exist. Just a suggestion

nkryuchkov (changes requested):
Good job! Please fix linter error adding a comment to `UnlinkSockFile`.

Darkren:
@Kifen also shouldn't we remove also the `dmsgpty` socket? I'm not sure it's probably being removed already, but at least we should check if it really is

Darkren (approved):
Good job!

PR #115, opened 2020-01-09
[M2] Integrate the yamux version of DMSG

Darkren:
Fixes #114 

 Changes:	
- New DMSG with yamux is integrated, all calls changed

PR #118, opened 2020-01-14
Rename messaging to dmsg

Darkren:
Fixes https://github.com/SkycoinPro/skywire-services/issues/26

 Changes:	
- Renamed messaging to dmsg
- Tested it with the integration env


cmd/dmsg-server/commands/root.go
```diff
@@ -27,7 +27,7 @@ var (
 	cfgFromStdin bool
 )
 
-// Config is a messaging-server config
+// Config is a DMSG server config
```
nkryuchkov: @evanlinjin I remember you said that lowercase spelling would be correct (`dmsg`). Is `DMSG` appropriate as well?

evanlinjin: @nkryuchkov I've always stuck with all lowercase. To be honest, I'm not sure what is best haha. What convention do you think we should have?

nkryuchkov: @evanlinjin @Darkren Maybe `DMsg`?

evanlinjin (changes requested):
As this is a milestone 2 PR, we should use binaries from the `dmsg` repo (`dmsg-server` and `dmsg-discovery`), as they have been moved there.

https://github.com/evanlinjin/dmsg-1/tree/feature/m2-yamux/cmd

cmd/dmsg-server/commands/root.go
evanlinjin: @nkryuchkov technically correct. But someone needs to go in and change it all haha.

Darkren: @evanlinjin @nkryuchkov should I change it? and what for? I didn't get the resume lol

Darkren:
@evanlinjin this makes this PR blocked by this one #115 . In #115  and its bro PR from services repo I use binaries from the `dmsg` repo. if this is the only problem here I think it would be better to merge this one for now. after finishing and merging #115 everything should be in its place

cmd/dmsg-server/commands/root.go
evanlinjin: @Darkren @nkryuchkov Let's stick with lowercase `dmsg` for now. As that's how it is written everywhere else. We can talk about changes in the future.

Darkren: Done this

nkryuchkov (approved):

PR #119, opened 2020-01-15
WIP: Implemented backoff for transport reconnection

### LINKED ISSUE #68: Implement backoff for transport reconnection
**Feature description**

Currently, when one edge of a transport goes down, the other edge will try to re-establish the transport every three seconds. We should implement a backoff for re-establishing the transport. @evanlinjin suggested using the retry logic in /internal/netutil . Also, we need to improve logging, so that these messages do not clutter stdout:

```[2019-11-13T10:59:10+08:00] WARN [tp:03c775]: failed to redial underlying connection: dial tcp 192.168.0.107:7777: connect: connection refused```

- [x] improve logging
- [x] implement retrial backoff

### PR DESCRIPTION
**Pull request by Kifen**
Fixes #68 	

 Changes:	
-	updated transport connection redial with a backoff implementation

How to test this PR:
Establish a connection between two visors, then shut down one visor

### CODE REVIEW

File: pkg/transport/managed_transport.go [hunk 1]
```diff
@@ -135,28 +136,25 @@ func (mt *ManagedTransport) Serve(readCh chan<- routing.Packet, done <-chan stru
...
-					if err := mt.dial(ctx); err != nil {
-						mt.log.Warnf("failed to redial underlying connection: %v", err)
-					}
-				}
-				mt.connMx.Unlock()
-			}
+	// redial underlying connection
+	err = r.Do(func() error {
+		if mt.logMod() {
+			err = mt.ls.Record(mt.Entry.ID, mt.LogEntry)
```
nkryuchkov: What do you think of writing the following? 
```golang
return mt.ls.Record(mt.Entry.ID, mt.LogEntry)
```

pkg/transport/managed_transport.go
```diff
@@ -135,28 +136,25 @@ func (mt *ManagedTransport) Serve(readCh chan<- routing.Packet, done <-chan stru
...
-				mt.connMx.Lock()
-				if mt.conn == nil {
-					if err := mt.dial(ctx); err != nil {
-						mt.log.Warnf("failed to redial underlying connection: %v", err)
-					}
-				}
-				mt.connMx.Unlock()
-			}
+	// redial underlying connection
+	err = r.Do(func() error {
```
nkryuchkov: What do you think of writing the following? 
```golang
return r.Do(func() error {
```

pkg/transport/managed_transport.go
```diff
@@ -135,28 +136,25 @@ func (mt *ManagedTransport) Serve(readCh chan<- routing.Packet, done <-chan stru
 		}
 	}()
 
-	// Redial loop.
-	for {
-		select {
-		case <-mt.done:
-			return
+	r := netutil.NewRetrier(3*time.Second, 5, 2)
+	var err error
```
nkryuchkov: Could we get rid of this line?

pkg/transport/managed_transport.go
```diff
@@ -135,28 +136,25 @@ func (mt *ManagedTransport) Serve(readCh chan<- routing.Packet, done <-chan stru
 		}
 	}()
 
-	// Redial loop.
-	for {
-		select {
-		case <-mt.done:
-			return
+	r := netutil.NewRetrier(3*time.Second, 5, 2)
```
nkryuchkov: I think it might be great if we could set up the values in the config.

pkg/transport/managed_transport.go
```diff
@@ -135,28 +136,25 @@ func (mt *ManagedTransport) Serve(readCh chan<- routing.Packet, done <-chan stru
...
-				}
-				mt.connMx.Unlock()
-			}
+	// redial underlying connection
+	err = r.Do(func() error {
+		if mt.logMod() {
+			err = mt.ls.Record(mt.Entry.ID, mt.LogEntry)
+			return err
 		}
+		mt.connMx.Lock()
```
nkryuchkov: What do you think of writing in the following way?

```golang
	mt.connMx.Lock()
	defer mt.connMx.Unlock()

 	return mt.dial(ctx)
```

pkg/transport/managed_transport.go
```diff
@@ -135,28 +136,25 @@ func (mt *ManagedTransport) Serve(readCh chan<- routing.Packet, done <-chan stru
...
+		err = mt.dial(ctx)
+		mt.connMx.Unlock()
+		return err
+	})
+
+	if err != nil {
+		mt.log.Warn("failed to redial underlying connection")
+	} else {
```
nkryuchkov: I would get rid of the `else` here returning in the error handling block.

pkg/transport/managed_transport.go
evanlinjin: I think that will be great too. But which values are you referring to?

pkg/transport/managed_transport.go
evanlinjin: Technically avoiding the `defer` statement can increase performance, but only by a minuscule amount. Personally I am okay with both styles.

evanlinjin (changes requested):
`logTicker.C` was to ensure actions were logged in database. We should really keep this.
The exponential backoff should only trigger when we have disconnected from the underlying connection.

pkg/transport/managed_transport.go
nkryuchkov: @evanlinjin I mean args of `NewRetrier`, i.e. backoff, times and factor.

pkg/transport/managed_transport.go
Kifen: Yes, it's concise.

pkg/transport/managed_transport.go
Kifen: Will do

pkg/transport/managed_transport.go
Kifen: @nkryuchkov  @evanlinjin will set the values in config.

cmd/skywire-cli/commands/node/gen-config.go
```diff
@@ -149,6 +149,14 @@ func defaultConfig() *visor.Config {
 
 	conf.AppServerSockFile = "/tmp/visor_" + pk.Hex() + ".sock"
 
+	if conf.Retrier == nil {
+		conf.Retrier = &visor.RetrierConfig{
```
nkryuchkov, line 153: I would create a new variable or function for default config to be able to reuse it. `3-5-2` values are used many times.

pkg/transport/managed_transport.go
```diff
@@ -160,6 +177,20 @@ func (mt *ManagedTransport) Serve(readCh chan<- routing.Packet, done <-chan stru
 	}
 }
 
+// redialConn redials an underlying connection using an exponetial backoff
+func (mt *ManagedTransport) redialConn(ctx context.Context) error {
+	mt.retrier.alive = true
+	err := mt.retrier.r.Do(func() error {
```
nkryuchkov: I think we could just return here.

pkg/visor/visor.go
```diff
@@ -6,6 +6,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/SkycoinProject/skywire-mainnet/internal/netutil"
```
nkryuchkov: Could you please sort imports here and in other files?

Kifen: Sorted

jdknives:
Made obsolete by #148

PR #126, opened 2020-01-23
Hypervisor improvements

### LINKED ISSUE #75: It is not possible to make operations related to the routes with the API
**Describe the bug**
The hypervisor API has various endpoints that allow to work with the routes, like `GET /nodes/{pk}/routes/{rid}`. The problem is that it is not possible to get the value that must be sent as the `{rid}` param, as the hypervisor API does not have an endpoint for getting the ID of the routes.

Calling the `GET /nodes/{pk}/routes` endpoint only returns a `key` property for each route, which worked before when sent as the `{rid}` param, but not anymore, as the API responds `error: "invalid UUID length: 0"`.

**Possible implementation**
2 possible solutions are:
- Add the ID to the `GET /nodes/{pk}/routes` endpoint.
- Make the API endpoints related to the routes work if the `key` property returned by the `GET /nodes/{pk}/routes` endpoint is sent as the `{rid}` param.

### LINKED ISSUE #92: Make /exec endpoint consistent with others
While all endpoints for nodes management look like `/nodes/{pk}/{action}`, the endpoint for command execution looks like `/exec/{pk}`, which is not consistent with others and violates REST.

I think `/nodes/{pk}/exec` would look better.

nkryuchkov:
Closes #75
Closes #92

 Changes:	
1. Fix the #75 bug
2. Refactor the `hypervisor` module
3. Fix #92

How to test this PR:
1. Follow the instructions from #75
2. Run hypervisor tests and linter


Darkren (approved):
Great job!

jdknives (approved):
Works well!

jdknives (approved):
Works.

PR #128, opened 2020-01-24
Fix broken hypervisor after merging

nkryuchkov:
Hypervisor panics on start because of duplicate flag definitions


jdknives (approved):
Good Catch.

PR #130, opened 2020-01-27
Integrate skywire-peering-daemon to mainnet

Kifen:

Fixes #	Supersedes https://github.com/SkycoinProject/skywire-mainnet/pull/108

 Changes: 
-	Added method to run and stop the skywire-peeringdaemon

How to test this PR:
Clone skywire-peering-daemon to a directory adjacent to skywire-mainnet and skywire-services. Set auto_start_spd to true in skywire-services, then run an integration env.


jdknives:
Getting the following error. 
```
[2020-01-27T14:10:38+01:00] INFO [SPD]: Received packet 032488721b4b0146a864bcede123fb62a0c155098c72857eb8b4c4b1141e2cb675: 192.168.2.106:7777
[2020-01-27T14:10:38+01:00] INFO [SPD]: Sending packet over pipe
panic: assignment to entry in nil map

goroutine 100 [running]:
github.com/SkycoinProject/skywire-mainnet/pkg/visor.readPacket(0xc0002b8000, 0xc00015e3c0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/Users/rudikleine/go/src/github.com/SkycoinProject/skywire-mainnet/pkg/visor/peering_daemon.go:98 +0x38f
created by github.com/SkycoinProject/skywire-mainnet/pkg/visor.(*Node).RunDaemon
	/Users/rudikleine/go/src/github.com/SkycoinProject/skywire-mainnet/pkg/visor/visor.go:349 +0x62c
```

pkg/visor/peering_daemon.go
```diff
@@ -0,0 +1,109 @@
...
+	"io"
+	"net"
+	"net/rpc"
+	"os"
+	"os/exec"
+	"sync"
+	"time"
+
+	"github.com/SkycoinProject/dmsg/cipher"
+	skycoin_cipher "github.com/SkycoinProject/skycoin/src/cipher"
```
evanlinjin, line 15: I think we should only import `"github.com/SkycoinProject/dmsg/cipher"` here. If there are missing functions that we need, we should add it in `"github.com/SkycoinProject/dmsg/cipher"`.

pkg/visor/peering_daemon.go
```diff
@@ -0,0 +1,109 @@
...
+func execute(cmd *exec.Cmd) error {
+	cmd.Stdout = os.Stdout
+	if err := cmd.Start(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func client() (RPCClient, error) {
```
evanlinjin: How is this function related to the Peering Daemon?

pkg/visor/peering_daemon.go
```diff
@@ -0,0 +1,109 @@
...
+		return nil, fmt.Errorf("RPC connection failed: %s", err)
+	}
+	if err := conn.SetDeadline(time.Now().Add(rpcConnDuration)); err != nil {
+		return nil, fmt.Errorf("RPC connection failed: %s", err)
+	}
+	return NewRPCClient(rpc.NewClient(conn), RPCPrefix), nil
+}
+
+// transport establshes an stcp transport to a remote visor
+func createTransport(networkType string, packet spd.Packet) (*TransportSummary, error) {
```
evanlinjin: Why is the second input of type `spd.Packet` instead of `cipher.PubKey`?

pkg/visor/visor.go
```diff
@@ -302,6 +306,63 @@ func (node *Node) Start() error {
...
+		return fmt.Errorf("Failed to start daemon as an external process: %s", err)
+	}
+
+	node.logger.Info("Opening named pipe for reading packets from skywire-peering-daemon")
+	stdOut, err := os.OpenFile(namedPipe, os.O_RDONLY, 0600)
+	if err != nil {
+		return err
+	}
+
+	pubKeyTable := node.conf.STCP.PubKeyTable
```
evanlinjin: We may need to check whether the pk table is nil or not.

evanlinjin (changes requested):
Good work. Got some queries and suggestions.

pkg/visor/peering_daemon.go
Kifen: When a visor receives a packet from a peering daemon, it's supposed to automatically establish an `stcp` transport to a visor whose public key is contained in the (received) packet. 

The function doesn't in any way relate to the peering daemon. There's a similar function in `cmd/skywire-cli/commands/visor/root.go` which returns a RPC client needed to establish transport (dmsg/stcp).  Thinking of moving this function to `pkg/httputil` package, so it can be used in the peering_daemon.go file and also by `skywire-cli`

pkg/visor/peering_daemon.go
Kifen: Passing any of `spd.Packet` or `cipher.PubKey` would work, and I would only need to change a line in the function body to suit the passed parameter. But looking at it again I think passing `cipher.Pubkey` is simpler.

pkg/visor/peering_daemon.go
evanlinjin: @Kifen Thanks for the explanation and that does make more sense.

However, just a few notes:
1. `httputil` package already exists under `dmsg/httputil`. Remove `httputil` as it is a duplicate.
2. I feel like this function should be under the same package, just a different file. Consider putting this in `rpc_client.go`.

pkg/visor/peering_daemon.go
evanlinjin: @Kifen Also in the comment it states `stcp transport`, but the first input is `networkType`.

pkg/visor/peering_daemon.go
```diff
@@ -0,0 +1,109 @@
...
+func watchNamedPipe(file string, c chan notify.EventInfo) error {
+	err := notify.Watch(file, c, notify.Write)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func readPacket(stdOut *os.File, c chan notify.EventInfo, m map[cipher.PubKey]string) TransportSummary {
```
evanlinjin: We are under the `visor` package so we need to name this better. Maybe a better name would be `readSPDPacket`?

evanlinjin (changes requested):

pkg/visor/peering_daemon.go
Kifen: True. The `networkType` parameter is redundant.

pkg/visor/peering_daemon.go
Kifen: `readSPDPacket` is better

pkg/visor/peering_daemon.go
Kifen: For point 2, looking at the function again, already `NewRPCClient` exist in `rpc_client.go` and it returns an `rpcClient` required for making rpc calls. What the function `client()` does in peering_daemon.go is create a `net.Conn` type which is then used in the call to `NewRPCClient`.

pkg/visor/visor.go
Kifen: You're right.

jdknives:
Shall I give that a test again @Kifen ?

jdknives (approved):

pkg/visor/peering_daemon.go
evanlinjin: @Kifen What is the progress with this?

Kifen: There are functions available in `SkycoinProject/skycoin/src/cipher/crypto.go` which are missing in `github.com/SkycoinProject/dmsg/cipher`. An example is the `MustPubKeyFromHex`, which I made use of in `visor/peering_daemon.go`.

evanlinjin: It will be good if we can have them in `"github.com/SkycoinProject/dmsg/cipher"` as well. Thanks.

pkg/visor/peering_daemon.go
```diff
@@ -0,0 +1,109 @@
...
+
+// transport establshes an stcp transport to a remote visor
+func createTransport(networkType string, packet spd.Packet) (*TransportSummary, error) {
+	client, err := client()
+	if err != nil {
+		return nil, err
+	}
+
+	logger.Infof("Establishing transport to remote visor")
+	rPK := skycoin_cipher.MustPubKeyFromHex(packet.PublicKey)
```
evanlinjin: This will result in panic if remote sends invalid PK.

Kifen: Right. Will verify it first before any other action

evanlinjin:
Closed for now. Delayed for M3 or M4.

PR #133, opened 2020-01-28
Fix: Ensure transports are deregistered and removed when rm-tp is run.

evanlinjin:
Did you run `make format && make check`?
> Yes.

Fixes #125 
Fixes #127 

Changes:	
* Implemented .redial() member that only attempts to .dial() when transport is still registered in discovery.

* Both .WritePacket() and .Serve() calls .redial() instead of .dial()

* In .DeleteTransport(), we deregister the transport from discovery before closing the physical connection.

How to test this PR:
> TODO.

PR #134, opened 2020-01-28
Fix infinite keep alive loop

### LINKED ISSUE #95: [M2] Check behavior of route groups
We need to check the behavior of route groups. Specifically, let's consider 2 route groups communicating. If visor with one of them comes down, and the second one still tries to communicate what errors will it be experiencing? Will there be errors at all? What should be considered as a correct behavior in this case?

Darkren:
Partially fixes #95 

 Changes:	
- Added correct exit from keep-alive loop on route group closing


evanlinjin (approved):
Looks good. Good work.

Kifen (approved):
Good!

evanlinjin:
The test that is failing in Travis is fixed in #133

PR #135, opened 2020-01-28
Improve route group logging, fix route group closing procedure

### LINKED ISSUE #95: [M2] Check behavior of route groups

Darkren:
Fixes #95	

 Changes:	
- Route group closing procedure was fixed. Now it doesn't get interrupted in case any of the intermediay nodes is down
- Improved logging a bit, logs are more informative now


nkryuchkov (approved):
Good job!

jdknives:
@Darkren could you please fix the linter and pull the changes? I think that may fix the tests.

PR #140, opened 2020-01-29
Version command

### LINKED ISSUE #137: Version command
**Feature description**

It would be generally wise to have a version command that returns the current Skywire version number. We can either add a Version file for that purpose as done in Skywallet project or add is as a variable in visor package itself. Unless there are any reasons not to go with the latter way, we should declare it in visor package as thats the way the wallet does things. 

We should implement:

- CLI command for that
- hypervisor endpoint that returns version number

For now the version number can be 0.0.1 until we release mainnet. 

This is a prerequisite for #129

nkryuchkov:
Yes

Fixes #137 

 Changes:	
- `make build` takes version number from `git describes` and stores it in the compiled binary along with commit hash and build date in RFC3339 format
- output build info on the start of all binaries except `skywire-cli`
- added CLI command for printing build info (`skywire-cli node version`)
- return build info in hypervisor endpoint (`/nodes/{pk}`)
- added a `0.1.0` tag to this repository

How to test this PR:


evanlinjin (approved):
Beautiful work.

Kifen (approved):
Good!

PR #142, opened 2020-01-30
Add build options to missing Makefile targets

nkryuchkov:
Some build targets in Makefile were missed while implementing #137, and build versions are not stored when using them. This PR adds build options for storing versions to all `Makefile` targets.


jdknives (approved):
Works. Well done!

PR #148, opened 2020-02-03
Implemented backoff for transport reconnection

### LINKED ISSUE #68: Implement backoff for transport reconnection

Kifen:
yes
Fixes #68  	

 Changes:	
-	updated transport connection redial with a backoff implementation

How to test this PR:
Establish a connection between two visors, then shut down one visor


pkg/router/router_test.go
```diff
@@ -10,6 +10,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/SkycoinProject/skywire-mainnet/internal/netutil"
```
nkryuchkov: Could you please move this line and line 15 below to corresponding import blocks?

pkg/transport/managed_transport.go
```diff
@@ -248,6 +264,17 @@ func (mt *ManagedTransport) dial(ctx context.Context) error {
 	return mt.setIfConnNil(ctx, tp)
 }
 
+// redialConn redials an underlying connection using an exponetial backoff
```
nkryuchkov: `exponetial` => `exponential`

pkg/transport/manager_test.go
```diff
@@ -9,6 +9,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/SkycoinProject/skywire-mainnet/internal/netutil"
```
nkryuchkov, line 12: Could you please move 12 and 14 to corresponding blocks?

pkg/visor/visor_test.go
```diff
@@ -9,6 +9,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/SkycoinProject/skywire-mainnet/internal/netutil"
```
nkryuchkov: Could you please move it to corresponding block?

pkg/transport/managed_transport.go
```diff
@@ -151,11 +164,14 @@ func (mt *ManagedTransport) Serve(readCh chan<- routing.Packet, done <-chan stru
 				// If there has not been any activity, ensure underlying 'write' tp is still up.
 				mt.connMx.Lock()
 				if mt.conn == nil {
```
nkryuchkov: I would write `if mt.conn == nil && !mt.retrier.alive` to descrease indentation

nkryuchkov (changes requested):

pkg/router/router_test.go
Kifen: done

pkg/transport/managed_transport.go
Kifen: corrected

pkg/transport/manager_test.go
Kifen: done

pkg/visor/visor_test.go
Kifen: done

nkryuchkov (changes requested):
Great work! Please just remove the committed binaries

nkryuchkov (approved):
Actually, I haven't seen that the binaries are in the `vendor` folder, so that's OK. Maybe just running `make dep` would be great.

jdknives:
Sorry again @Kifen . But the issue tackled with this PR will be fixed in #166

Issue #146, opened 2020-01-31
Inconsistent behavior when changing Skysocks config 


jdknives:
Initially described here by @Darkren : https://github.com/SkycoinProject/skywire-mainnet/issues/123

Changing PK of skysocks-client which is STOPPED at this point. Logs:

![](https://user-images.githubusercontent.com/13316082/73166851-bc6c1d80-4107-11ea-82ad-c0172c2f378e.png)

![](https://user-images.githubusercontent.com/13316082/73166860-c130d180-4107-11ea-95e2-d7efca9d7f03.png)

Logs seem ok except for the fact that this request immediately starts the app even if it was closed. not sure if this is intended. but I get the following response from the API:

```
{
    "error": "stop app skysocks-client: unknown app"
}
```

UPD: it outputs Restarting app skysocks-client but that doesn't really happen, at least not in my case

Logging should be consistent and changing the value of config for Skysocks should not automatically restart the app.

jdknives:
Closed by #149

Issue #144, opened 2020-01-31
Skysocks cannot be restarted


jdknives:
**Describe the bug**
Originally reported by @Darkren here: https://github.com/SkycoinProject/skywire-mainnet/issues/123

 Logs of visor hosting skysocks while changing password:

![](https://user-images.githubusercontent.com/13316082/73166306-ab6edc80-4106-11ea-9c10-32b7b2d75ce1.png)

Probably restart happened too soon after socket closed, cause when I tried to start if from hypervisor, skysocks sucessfully started. Not sure

Darkren:
Actually passcode change goes ok sometimes. But sometimes the picture is as above. Looking deep

jdknives:
Closed by #149

PR #149, opened 2020-02-03
Fix several app restarting issues

Darkren:
Fixes:
- #144 
- #146 

 Changes:	
- Improved logging on app restart, change of passcode/PK
- When changing passcode/PK we now check if the app is down. If it is, it won't start


Darkren:
Now working on #144 , I guess it would be more comfortable to do both issues in a single PR since they're highly related

Darkren:
Actually this one is done, but I feel like I found one more related bug. Checking now

nkryuchkov (approved):
Great work

Kifen (approved):
Good!

pkg/app/appserver/proc.go
```diff
@@ -62,12 +64,21 @@ func (p *Proc) Start() error {
...
+		return err
+	}
+
+	<-p.doneC
+
+	return nil
 }
 
 // Wait waits for the application cmd to exit.
 func (p *Proc) Wait() error {
```
evanlinjin, line 108: `Wait()` should only be called once, and `Stop()` will only return if `Wait()` is called before. If this is by design, we should include it in the comments.

If not, an alternative would be to protect `p.cmd.Wait()` with `sync.Mutex`. And use it in both `Stop()` and `Wait()` (instead of having `doneC chan struct{}`).

Let me know if I'm mistaken.

evanlinjin (changes requested):
Good work. Please review my query.

pkg/app/appserver/proc.go
Darkren: @evanlinjin well spotted! I'll fix this stuff

PR #152, opened 2020-02-06
Fix rule expiration

### LINKED ISSUE #147: Cannot use proxy after restart
Originally described by @Darkren here: https://github.com/SkycoinProject/skywire-mainnet/issues/123

I did the following:

1. Started visors A and C on local machine
2. Added transport C->B, A->B
3. Tried changing password and PK for skysocks
4. Tried curling through skysocks - it works
5. Requested visor restart from hypervisor for C where skysocks-client is hosted. It went down and rose back up just great. but then, logs:

Then tried curling through proxy and got this:

Restarting skysocks-client from hypervisor doesn't save the day, CURL request hangs after that

Darkren:
Partially fixes #147 	

 Changes:	
- Now rule activity gets updated automatically only for consume type of rules (to avoid unnecessary mutex acquiring for the second time in a row)
- In all other cases route groups tells router to update rule activity ONLY if the packet was propagated successfully. This way when transport goes down, rules will be garbage collected and the route will disappear after a timeout
- Routing table is now being initialized inside a router constructor. All outer calls to the table are proxied through the router instance
- Timed out rules garbage collecting routine is now being called from the router, GC loop is moved there
- If consume rule gets timed out, router fetches the corresponding route group, closes it and totally removes. This allows outer code to react properly when remote visor gets killed unexpectedly
- Some route ID errors were fixed. At a couple of places packets were composed with the wrong ID which worked well in the freshly started integration env but actually started failing after a new route got established next to the old one
- Skysocks client is refactored a bit. Now we detect if app conn is failed at `DefaultRouteKeepAlive / 2` time tops. Then app conn is considered broken, client gets closed and reinstantiated by redialing app conn and recreating route. Redialing happens with an exponential backoff
- If dialing visor gets restarted, new route is established instantly. If remove visor (or any visor in between) gets restarted, new route is established as soon as dialing visor sees that route is broken (when route times out and route group gets closed)
- Route timeout reduced to 30 seconds, keep-alive internal is bound to be half of it (15 seconds for now)


Darkren:
So, the simplest part is done, moving forward

evanlinjin (commented):
Would `RuleLastAccessed` be a better term than `RuleActivity`?

pkg/router/router.go
```diff
@@ -634,3 +659,72 @@ func (r *router) IntroduceRules(rules routing.EdgeRules) error {
 		}
 	}
 }
+
+// RoutesCount returns count of the routes stored within the routing table.
+func (r *router) RoutesCount() int {
```
evanlinjin, line 708: Would it be easier just to have a function: `(*router).Table() routing.Table` to reduce clutter?

pkg/routing/table.go
```diff
@@ -40,43 +39,26 @@ type Table interface {
 
 	// Count returns the number of RoutingRule entries stored.
 	Count() int
+
+	// CollectGarbage checks all the stored rules, removes and returns ones that timed out.
+	CollectGarbage() []Rule
```
evanlinjin, line 44: What's the reasoning behind returning `Rule` instead of `RouteDescriptor`?

evanlinjin (commented):
Good work so far :)

Darkren:
Tests. All running in the proxy integration env (skysocks-client on C and skysocks on A):
1) curl, `kill -9` visor C, wait for route to timeout, start it again, curl again
2) curl, `kill -9` visor C, restart it after a couple of seconds, immediately curl as soon as it starts
3) repeat 1) for visor A
4) repeat 2) for visor A
5) repeat 1) for visor B
6) repeat 2) for visor B

- STCP:
- [x] 1)
- [x] 2)
- [x] 3)
This one won't work, since visor C is responsible for route establishing. And app actually quits as soon as the route times out. We should add a route re-establishment with the exponential backoff for such situations
- [x] 4)
Same as 3) above
- [x] 5)
- [x] 6)

- DMSG:
- [ ] 1)
Looks exactly the same as 2) described below
- [ ] 2)
When I do this, I get the following logs on the restarted visor C:
```
time="2020-02-25T08:38:34+03:00" level=warning msg="route finder: no route to destination" source="retrier.go:90"
```
Then after a couple of seconds:
```
failed to redial underlying connection (redial loop): settlement handshake failed: context deadline exceeded
Failed to update statuses error="Post http://localhost:9091/statuses: context deadline exceeded"
```
Then transport gets `UP` status but logs show:
```
 WARN [tp:0348c9]: failed to read packet: EOF
```
And this line just keeps spamming

- [ ] 3)
- [ ] 4)
- [ ] 5)
- [ ] 6)

Darkren:
2), 3) and 4) for STCP show strange hard to describe symptoms. Working on it. Probably one fix will do these both

Darkren:
Removed a couple of tests from checklist since they're not actual anymore

pkg/router/router.go
Darkren: @evanlinjin well, I just proxied all calls to the table through router, there's outer call from visor to get this count, so I left it as is to ensure compatibility

pkg/routing/table.go
Darkren: @evanlinjin routing table stores separate rules, not descriptors. descriptors are part of a route group which stands a bit above the table, table itself is rather simple

Darkren:
> Would `RuleLastAccessed` be a better term than `RuleActivity`?

@evanlinjin sorry, can't find it. probably it was there before and now it's gone. if it's still actual, please let me know

Darkren:
1. We should discuss default keep alive timeout, we agreed on lowering them but we'll need to decide on the actual values. Also we should discuss the keep-alive packets interval
2. There're a lot of debug logs here which must be removed. But it's highly related to #171 which needs to be done first. I suggest that we finish #171 , remove debug logs and then merge this one

evanlinjin (approved):

evanlinjin:
@Darkren I think we can merge after we sort out the logging situation.

nkryuchkov (approved):
Great job!

PR #153, opened 2020-02-06
Updating mechanism for mainnet

### LINKED ISSUE #129: Updating mechanism for mainnet
**Feature description**

We will need to implement an updating mechanism for `mainnet` before we switch over to `mainnet` from `testnet`

This is a ticket for discussing the best way to implemnt such a mechanism.

We need to ensure that:

- configuration files persist
- the update can be triggered from `hypervisor`
- the visor connects to hypervisor after update automatically

We need to update:

- [x] visor

- [x] skywire-cli

- [x] application binaries

### LINKED ISSUE #141: Build binaries on new tag creation
Currently, we have only one `0.1.0` tag. GitHub automatically generated a release for it: https://github.com/SkycoinProject/skywire-mainnet/releases. It contains source code but no binary builds, that are required for https://github.com/SkycoinProject/skywire-mainnet/issues/129.

We need to investigate how to perform the above. There are some Go projects on GitHub that distribute binaries in GitHub releases, so having a look at them may be useful.

Here are links about GitHub releases and how to upload a binary asset that may help:
https://help.github.com/en/github/administering-a-repository/about-releases
https://github.com/actions/upload-release-asset
https://github.com/goreleaser/goreleaser

Here are projects having binaries in GitHub releases:
https://github.com/golangci/golangci-lint/releases

nkryuchkov:
NOTE: After merging this PR, all `git` tags and GitHub releases should be removed because they exist for testing purposes.

Fixes #129
Fixes #141

It should be merged simultaneously with https://github.com/SkycoinPro/skywire-services/pull/84.

 Changes:	
- Implement `skywire-visor` and `skywire-cli` binary files update via the `hypervisor`
- Change `GET /visors/{pk}/restart` to `POST /visors/{pk}/restart`
- Setup `goreleaser` for `skywire-visor` and `skywire-cli`

How to test this PR:
1. Comment out the `VERSION := $(shell git describe)` line in `Makefile`
2. Add a new line below it: `VERSION := 0.1.0`
3. Run the generic integration environment
4. Go to `VisorA` pane
5. Make an HTTP `POST` request to `localhost:8080/api/visors/024ec47420176680816e0406250e7156465e4531f5b26057c9f6297bb0303558c7/update`
6. Check if there are no errors in `VisorA` pane and `VisorA` is restarted successfully and works. Apps should also work. Versions of all binaries in their output should be updated.
7. Check if the HTTP response status is `200` and body is `true`
8. Go to `shell` pane
9. Check if `skywire-visor`, `skywire-cli` and `VisorA` apps binaries are updated and work (e.g. using `ls -lah ./bin`)
10. Check if files/directories from logs from `VisorA` saying `Removing file ...` are removed.
11. Note that in fact, an old version of binaries is uploaded to the last GitHub release. GitHub releases need a clean-up.


pkg/util/updater/updater.go
```diff
@@ -0,0 +1,367 @@
...
+	}
+
+	if !valid {
+		return "", fmt.Errorf("checksum is not valid")
+	}
+
+	return path, nil
+}
+
+func (u *Updater) start(path string) error {
```
Darkren: what do you think about calling it `restart`? to make things cleaner

pkg/util/updater/updater.go
```diff
@@ -0,0 +1,367 @@
...
+	if first < 0 {
+		return "", ErrMalformedChecksumFile
+	}
+
+	return checksums[first:last], nil
+}
+
+func downloadChecksums(url string) (checksums string, err error) {
+	resp, err := http.Get(url) // nolint:gosec
+	if resp != nil {
```
Darkren: can't we check for the error first, and omit this `resp != nil` check? may `resp` be nil when there's no error?

pkg/util/updater/updater.go
```diff
@@ -0,0 +1,367 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	return string(data), nil
+}
+
+func downloadFile(url, filename string) (path string, err error) {
+	resp, err := http.Get(url) // nolint:gosec
+	if resp != nil {
```
Darkren: same question as above

Darkren (approved):
Other than these small things, it is a really great job!

pkg/util/updater/updater.go
nkryuchkov: I've seen some advice that recommends checking the response body even if `err != nil` because it may happen on a redirection failure condition. E.g.: http://devs.cloudimmunity.com/gotchas-and-common-mistakes-in-go-golang/index.html#close_http_resp_body

However, I've checked `(*Client) Do` in `net/http/client.go` and found this:
```
// On error, any Response can be ignored. A non-nil Response with a
// non-nil error only occurs when CheckRedirect fails, and even then
// the returned Response.Body is already closed.
```

So, actually, we can rewrite it as you said if you think it will be more readable. There may be similar code snippets written by me throughout the project, so in this case, we may also need to change them too.

pkg/util/updater/updater.go
nkryuchkov: I named it `start` because I meant `start updater`. But I think we can rename it if you'd like to.

pkg/util/updater/updater.go
nkryuchkov: Answered above

pkg/util/updater/updater.go
Darkren: I think it's not that crucial, but I didn't get the thought at the first glance

pkg/util/updater/updater.go
Darkren: I think it would be great to do so, just to make things a bit cleaner

Kifen (approved):
Nice work

jdknives (approved):
Works well in the integration env. Really good job man! Would like to give that another test outside of the integration env as soon as test deployment is stable again though.

Darkren (approved):

jdknives:
Also, we need to add support for updating appliations.

jdknives (approved):
Works like a charm. 👑

PR #200, opened 2020-03-04
Remove some panics

### LINKED ISSUE #186: Get rid of panics where possible
We have a lot of `panic()`'s in the codebase. They often fire unexpectedly, usually in tests, but also may do in production, which is unsafe. Moreover, they are a code smell (https://github.com/golang/go/wiki/CodeReviewComments#dont-panic). So, `panic()`'s are to be rewritten into returning or logging errors wherever it's possible.

This should be done in the following repositories:

- [x] https://github.com/SkycoinProject/skywire-mainnet

- [x] https://github.com/SkycoinPro/skywire-services

- [ ] https://github.com/SkycoinProject/dmsg

nkryuchkov:

Partially fixes #186 

 Changes:	
- Remove some panics

How to test this PR:
`make check`

### DIFF
--- a/pkg/snet/network.go
+++ b/pkg/snet/network.go
@@ -169,14 +169,14 @@ func (n *Network) Dial(ctx context.Context, network string, pk cipher.PubKey, po
 			return nil, err
 		}
 
-		return makeConn(conn, network)
+		return makeConn(conn, network), nil
 	case STcpType:
 		conn, err := n.stcpC.Dial(ctx, pk, port)
 		if err != nil {
 			return nil, err
 		}
 
-		return makeConn(conn, network)
+		return makeConn(conn, network), nil
 	default:
 		return nil, ErrUnknownNetwork
 	}
@@ -191,14 +191,14 @@ func (n *Network) Listen(network string, port uint16) (*Listener, error) {
 			return nil, err
 		}
 
-		return makeListener(lis, network)
+		return makeListener(lis, network), nil
 	case STcpType:
 		lis, err := n.stcpC.Listen(port)
 		if err != nil {
 			return nil, err
 		}
 
-		return makeListener(lis, network)
+		return makeListener(lis, network), nil
 	default:
 		return nil, ErrUnknownNetwork
 	}
@@ -212,13 +212,9 @@ type Listener struct {
 	network string
 }
 
-func makeListener(l net.Listener, network string) (*Listener, error) {
-	lPK, lPort, err := disassembleAddr(l.Addr())
-	if err != nil {
-		return nil, err
-	}
-
-	return &Listener{Listener: l, lPK: lPK, lPort: lPort, network: network}, nil
+func makeListener(l net.Listener, network string) *Listener {
+	lPK, lPort := disassembleAddr(l.Addr())
+	return &Listener{Listener: l, lPK: lPK, lPort: lPort, network: network}
 }
 
 // LocalPK returns a local public key of listener.
@@ -237,7 +233,7 @@ func (l Listener) AcceptConn() (*Conn, error) {
 		return nil, err
 	}
 
-	return makeConn(conn, l.network)
+	return makeConn(conn, l.network), nil
 }
 
 // Conn represent a connection between nodes in Skywire.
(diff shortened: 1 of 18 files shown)


pkg/transport/managed_transport.go
```diff
@@ -412,7 +412,7 @@ func (mt *ManagedTransport) clearConn() {
 
 func (mt *ManagedTransport) updateStatus(isUp bool, tries int) (err error) {
 	if tries < 1 {
-		panic(fmt.Errorf("mt.updateStatus: invalid input: got tries=%d (want tries > 0)", tries))
```
evanlinjin: I feel a panic here is appropriate because it signals the developer that an invalid input is provided.
And the new error message is not clear and will be hard to track where the error came from.

pkg/snet/network.go
```diff
@@ -263,18 +279,24 @@ func (c Conn) RemotePort() uint16 { return c.rPort }
 // Network returns network of connection.
 func (c Conn) Network() string { return c.network }
 
-func disassembleAddr(addr net.Addr) (pk cipher.PubKey, port uint16) {
+func disassembleAddr(addr net.Addr) (pk cipher.PubKey, port uint16, retErr error) {
 	strs := strings.Split(addr.String(), ":")
 	if len(strs) != 2 {
-		panic(fmt.Errorf("network.disassembleAddr: %v %s", "invalid addr", addr.String()))
```
evanlinjin, line 269: I feel like panics here are appropriate, because is the `addr` cannot be disassembled to `pk:port`, it signals an invalid `snet.Conn`.

pkg/visor/rpc.go
```diff
@@ -42,16 +42,18 @@ type RPC struct {
...
+func newRPCServer(v *Visor, remoteName string) (*rpc.Server, error) {
 	rpcS := rpc.NewServer()
 	rpcG := &RPC{
 		visor: v,
 		log:   v.Logger.PackageLogger("visor_rpc:" + remoteName),
 	}
+
 	if err := rpcS.RegisterName(RPCPrefix, rpcG); err != nil {
-		panic(fmt.Errorf("failed to create visor RPC server: %v", err))
+		return nil, fmt.Errorf("failed to create visor RPC server: %v", err)
```
evanlinjin, line 53: I feel this should panic, as it signals a problem with the `rpcG` implementation.

evanlinjin (changes requested):
I feel some panics are appropriate. I think panics should be used to identify issues with the use of code base.

pkg/snet/network.go
Darkren: This may happen if one of our subsystems implements `net.Addr` the wrong way, the way we consider wrong. So, I'd stick to panic in this particular case, cause it will raise only on first test runs and will help us identify error fast instead of looking through logs

pkg/transport/managed_transport.go
Darkren: Probably it would be better if we just go for `uint` tries values here. We're calling this func passing `tries` as a raw number, not a variable, so no type conversions will be needed. And we won't need to decide on panic/error here

Darkren: Also, if we go with `uint` tries, we should change tries loop condition check inclusive down the code

pkg/visor/rpc.go
Darkren: Personally, I think error will be more suitable in such cases, cause it gets wrapped, and we're going `log.Fatal` in the outer code anyway. This way error will be easier to understand in logs

pkg/transport/managed_transport.go
evanlinjin: That's a good point. Thank you @Darkren

pkg/visor/rpc.go
evanlinjin: But panic will direct us to exactly which line it panicked.

pkg/snet/network.go
nkryuchkov: @evanlinjin @Darkren Thanks for your opinions. I agree.

pkg/visor/rpc.go
nkryuchkov: Let's leave panic here. I don't think that an error will give us big benefits.

pkg/transport/managed_transport.go
nkryuchkov: I'm going to leave panic here.

evanlinjin (approved):
Good.

PR #214, opened 2020-03-09
Fix update endpoint.

### LINKED ISSUE #213: Cannot update call update endpoint twice.
**Describe the bug**
Cannot call the hypervisor `/api/visors/{visor-pk}/update` endpoint twice for a given `visor` without resulting in `500 - {"error":"updating already started"}` error.

**Environment information:**
Independent, interactive generic environment.

evanlinjin:
Fixes #213

 Changes:	
* Added defer statement to Updater.Update to revert u.updating value back to 0 after updating.
* Make `restart.Context` reset it's internal `exec.Cmd` on failure.
* Changed restart visor endpoint to only accept `POST`.
* Update `restart.Restart` test.

### DIFF
--- a/pkg/restart/restart.go
+++ b/pkg/restart/restart.go
@@ -36,20 +36,17 @@ type Context struct {
 // Data used by CaptureContext must not be modified before,
 // therefore calling CaptureContext immediately after starting executable is recommended.
 func CaptureContext() *Context {
-	return &Context{
-		cmd:         captureCmd(),
-		checkDelay:  DefaultCheckDelay,
-		appendDelay: true,
-	}
-}
-
-func captureCmd() *exec.Cmd {
 	cmd := exec.Command(os.Args[0], os.Args[1:]...) // nolint:gosec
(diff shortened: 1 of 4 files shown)


jdknives (approved):
Cannot reproduce the error anymore.

pkg/restart/restart.go
```diff
@@ -70,24 +72,28 @@ func (c *Context) CmdPath() string {
...
-	defer ticker.Stop()
 
 	select {
-	case err := <-errCh:
+	case err = <-errCh:
 		c.errorLogger()("Failed to start new instance: %v", err)
-		return err
+
+		// Reset c.cmd on failure so it can be reused.
+		c.cmd = captureCmd()
```
nkryuchkov: I would copy the necessary fields from one `exec.Command` to another because data to be captured can be changed while the process is run. We are not going to change the data, but I think copying `exec.Command` is more robust.

nkryuchkov (approved):
Good job!

PR #217, opened 2020-03-09
Implement endpoint for checking if visor update is available

### LINKED ISSUE #210: Add an API endpoint to check if an update is available
**Feature description**
There should be an API endpoint for knowing if there is an update available. This is because the manager UI currently can show a button for updating the visor, but not a visual clue for letting the user know when to use it.

No, it would be for an UI/UX improvement.

A simple API endpoint returning if an update is available would be enough. However, if the endpoint can return the number of the current version and the new version it would be even better.

Maybe the front-end could check the lastest version using a method similar to the one used by the Go code, but that would mean repeated code, so if a change is made in the Go code and the same change is not made inmediatelly on the front-end code, problems may appear.

**Additional context**

**Possible implementation**

nkryuchkov:

Fixes #210 

 Changes:	
- Implement endpoint for checking if visor update is available
- Improve output of `/update` endpoint (**cc**: @Senyoret1)

### DIFF
--- a/pkg/util/updater/updater.go
+++ b/pkg/util/updater/updater.go
@@ -77,18 +77,18 @@ func (u *Updater) Update() (updated bool, err error) {
 	}
 	defer atomic.StoreInt32(&u.updating, 0)
 
-	lastVersion, err := u.UpdateAvailable()
+	latestVersion, err := u.UpdateAvailable()
 	if err != nil {
 		return false, fmt.Errorf("failed to get last Skywire version: %w", err)
 	}
 
-	if lastVersion == nil {
+	if latestVersion == nil {
 		return false, nil
 	}
 
-	u.log.Infof("Update found, version: %q", lastVersion.String())
+	u.log.Infof("Update found, version: %q", latestVersion.String())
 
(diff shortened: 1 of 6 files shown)


pkg/util/updater/updater.go
```diff
@@ -121,7 +116,28 @@ func (u *Updater) Update() (err error) {
...
+	return true, nil
+}
+
+// UpdateAvailable checks if an update is available.
+// If it is, the method returns the last available version.
+// Otherwise, it returns nil.
+func (u *Updater) UpdateAvailable() (*Version, error) {
+	u.log.Infof("Looking for updates")
+
+	lastVersion, err := lastVersion()
```
evanlinjin: Is this meant to be `lastVersion` or `latestVersion`?

evanlinjin (approved):
Good work. Just a small comment.

pkg/util/updater/updater.go
nkryuchkov: @evanlinjin `latestVersion` (the newest)

PR #222, opened 2020-03-10
Change transport type preference

### LINKED ISSUE #157: Change transport type preference
**Feature description**

When calling `skywire-cli visor add-tp $PubKey` without specifying the transport type explicitly, the visor should try to establish an `stcp` transport by default and only establish a `dmsg` transport once that fails. (As soon as we implemented a holepunch transport, this type should be setup before trying dmsg setup.)

Using an `stcp` transport is preferable when that is possible. However, it is not clear for a visor whether establishing an `stcp` transport is possible, given that the other visor could be behind a NAT. So either we would have to somehow exchange information about the remote visor's network properties or we just attempt to establish transports in the prefered order until transport establishment is successful. The latter option seems more straightforward for now.

nkryuchkov:

Fixes #157

 Changes:	
- When calling `skywire-cli visor add-tp $PubKey` without specifying the transport type explicitly, the visor tries to establish an `stcp` transport by default and only establish a `dmsg` transport once that fails.

How to test this PR:
1. Run integration env
2. Run `skywire-cli visor add-tp $PubKey` without explicitly specifying transport type or run `make integration-startup` in https://github.com/SkycoinPro/skywire-services/pull/109

### DIFF
--- a/cmd/skywire-cli/commands/visor/transports.go
+++ b/cmd/skywire-cli/commands/visor/transports.go
@@ -79,9 +79,15 @@ var (
 )
 
 func init() {
-	addTpCmd.Flags().StringVar(&transportType, "type", "", "type of transport to add")
-	addTpCmd.Flags().BoolVar(&public, "public", true, "whether to make the transport public")
-	addTpCmd.Flags().DurationVarP(&timeout, "timeout", "t", 0, "if specified, sets an operation timeout")
+	const (
+		typeFlagUsage    = "type of transport to add; if unspecified, cli will attempt to establish a transport in the following order: stcp, dmsg"
+		publicFlagUsage  = "whether to make the transport public"
+		timeoutFlagUsage = "if specified, sets an operation timeout"
+	)
+
+	addTpCmd.Flags().StringVar(&transportType, "type", "", typeFlagUsage)
+	addTpCmd.Flags().BoolVar(&public, "public", true, publicFlagUsage)
+	addTpCmd.Flags().DurationVarP(&timeout, "timeout", "t", 0, timeoutFlagUsage)
 }
 
 var addTpCmd = &cobra.Command{
--- a/cmd/skywire-cli/commands/visor/transports.go
+++ b/cmd/skywire-cli/commands/visor/transports.go
@@ -107,7 +107,7 @@ var addTpCmd = &cobra.Command{
 			tp, err = rpcClient().AddTransport(pk, transportType, public, timeout)
 			if err != nil {
 				logger.WithError(err).
-					Warnf("Failed to establish stcp transport. Trying to establish dmsg transport", pk)
+					Warnf("Failed to establish stcp transport. Trying to establish dmsg transport")
 
 				transportType = dmsg.Type
 
--- a/cmd/skywire-cli/commands/visor/transports.go
+++ b/cmd/skywire-cli/commands/visor/transports.go
@@ -12,6 +12,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/SkycoinProject/skywire-mainnet/cmd/skywire-cli/internal"
+	"github.com/SkycoinProject/skywire-mainnet/pkg/snet/stcp"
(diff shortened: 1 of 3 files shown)


jdknives (commented):
When running in deployment, I get: 

```
[2020-03-10T19:19:08+01:00] DEBUG [tp_manager]: Initializing TP with ID 3c776105-f783-039f-b6c8-e1d43b7f9f9a
[2020-03-10T19:19:08+01:00] INFO [tp_manager]: saved transport: remote(02a2f5c95e2275ed59ad82bcf99adf839b2adb1357f449076f4713f172230254e3) type(stcp) tpID(3c776105-f783-039f-b6c8-e1d43b7f9f9a)
[2020-03-10T19:19:08+01:00] WARN [tp_manager]: Underlying connection is not yet established. Will retry later. error="pk table: entry of 02a2f5c95e2275ed59ad82bcf99adf839b2adb1357f449076f4713f172230254e3 does not exist"
[2020-03-10T19:19:08+01:00] INFO [visor_rpc:CLI]: Request processed. _elapsed="125.37µs" _method="AddTransport" _received="7:19PM" input=&{RemotePK:02a2f5c95e2275ed59ad82bcf99adf839b2adb1357f449076f4713f172230254e3 TpType:stcp Public:true Timeout:0s} output=&{ID:3c776105-f783-039f-b6c8-e1d43b7f9f9a Local:03ac383847b90baaecddcdd2daeb29f49df896399e08db41fc13c81e0228600447 Remote:02a2f5c95e2275ed59ad82bcf99adf839b2adb1357f449076f4713f172230254e3 Type:stcp Log:<nil> IsSetup:false}
[2020-03-10T19:19:08+01:00] INFO [tp:02a2f5]: Serving. remote_pk=02a2f5c95e2275ed59ad82bcf99adf839b2adb1357f449076f4713f172230254e3 tp_id=3c776105-f783-039f-b6c8-e1d43b7f9f9a tp_index=1
``` 

I cannot send messages as it reports that there is no transport found for route. `Ls-tp` returns the transport of type `stcp`.

nkryuchkov:
@jdknives The transport manager didn't return an error if the STCP table doesn't have a requested PK. Fixed that behavior and cannot reproduce it anymore. Could you please try again?

jdknives (approved):

pkg/transport/manager.go
```diff
@@ -216,18 +217,29 @@ func (tm *Manager) SaveTransport(ctx context.Context, remote cipher.PubKey, tpTy
...
-
 		mTp, err := tm.saveTransport(remote, tpType)
 		if err != nil {
 			return nil, err
 		}
 
 		if err = mTp.Dial(ctx); err != nil {
+			// TODO(nkryuchkov): Check for an error that underlying connection is not established
+			// and try again in this case. Otherwise, return the error.
+			pkTableErr := fmt.Sprintf("pk table: entry of %s does not exist", remote.String())
```
evanlinjin, line 228: In my opinion, we should make our errors implement `net.Error` and make use of the `.Permanent() bool` method here. If `.Permanent() == true`, we stop serving.

evanlinjin (changes requested):

cmd/skywire-cli/commands/visor/transports.go
```diff
@@ -78,7 +79,7 @@ var (
 )
 
 func init() {
-	addTpCmd.Flags().StringVar(&transportType, "type", dmsg.Type, "type of transport to add")
+	addTpCmd.Flags().StringVar(&transportType, "type", "", "type of transport to add")
```
evanlinjin: We should probably expand on the flag description.
```
If unspecified, cli will attempt to establish a transport in the following order: stcp, dmsg.
```
Or something similar.

pkg/transport/manager.go
nkryuchkov: @evanlinjin Good idea, but I think it should be out of the scope of this PR. What do you think of leaving it as is for now and modifying after your proposal is implemented?

evanlinjin (approved):

PR #223, opened 2020-03-10
Send router packets asynchronously

### LINKED ISSUE #178: Send route keep alives asynchronously
**Feature description**

TBD

nkryuchkov:

Fixes #178 

 Changes:	
- Send router packets asynchronously

How to test this PR:
1. `make check`
2. Integration env

### DIFF
--- a/pkg/router/route_group.go
+++ b/pkg/router/route_group.go
@@ -367,9 +367,11 @@ func (rg *RouteGroup) keepAliveLoop(interval time.Duration) {
 				continue
 			}
 
-			if err := rg.sendKeepAlive(); err != nil {
-				rg.logger.Warnf("Failed to send keepalive: %v", err)
-			}
+			go func() {
+				if err := rg.sendKeepAlive(); err != nil {
+					rg.logger.Warnf("Failed to send keepalive: %v", err)
+				}
+			}()
 		}
 	}
 }
@@ -418,7 +420,7 @@ func (rg *RouteGroup) close(code routing.CloseCode) error {
 		rg.closeDone.Add(len(rg.tps))
 	}
 
-	rg.broadcastClosePackets(code)
+	go rg.broadcastClosePackets(code)
 
 	if closeInitiator {
 		// if this visor initiated closing, we need to wait for close packets
--- a/pkg/router/route_group.go
+++ b/pkg/router/route_group.go
@@ -367,37 +367,37 @@ func (rg *RouteGroup) keepAliveLoop(interval time.Duration) {
 				continue
 			}
 
-			go func() {
-				if err := rg.sendKeepAlive(); err != nil {
-					rg.logger.Warnf("Failed to send keepalive: %v", err)
-				}
-			}()
+			rg.sendKeepAlive()
 		}
 	}
 }
 
-func (rg *RouteGroup) sendKeepAlive() error {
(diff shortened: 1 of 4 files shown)


pkg/router/router.go
```diff
@@ -294,24 +294,34 @@ func (r *router) Serve(ctx context.Context) error {
 }
 
 func (r *router) serveTransportManager(ctx context.Context) {
+	done := make(chan struct{}, 1)
```
evanlinjin: What do you think about accompanying this with `once := new(sync.Once)` and call `once.Do(func(){close(done)})`? Just in case we encounter more than two `handleTransportPacket` go routines that reach `err == transport.ErrNotServing`. I'm assuming that in this case, one will hang.

Maybe another option is to have the `<-done:` case trigger a loop that empties `done`.

evanlinjin: I know that `serveTransportManager` will only be called once, but I feel we should still maintain a clean closing logic.

pkg/router/route_group.go
```diff
@@ -418,7 +420,7 @@ func (rg *RouteGroup) close(code routing.CloseCode) error {
 		rg.closeDone.Add(len(rg.tps))
 	}
 
-	rg.broadcastClosePackets(code)
+	go rg.broadcastClosePackets(code)
```
evanlinjin: I feel like we should use `rg.writePacketAsync` within `rg.broadcastClosePackets`, instead of calling `rg.broadcastClosePackets` in a go routine.

pkg/router/route_group.go
```diff
@@ -367,9 +367,11 @@ func (rg *RouteGroup) keepAliveLoop(interval time.Duration) {
 				continue
 			}
 
-			if err := rg.sendKeepAlive(); err != nil {
-				rg.logger.Warnf("Failed to send keepalive: %v", err)
-			}
+			go func() {
+				if err := rg.sendKeepAlive(); err != nil {
```
evanlinjin: I also feel like `sendKeepAlive` should range through `rg.tps` and use `rg.wrotePacketAsync`.

evanlinjin (changes requested):
Good work!

pkg/router/router.go
nkryuchkov: @evanlinjin This sounds like a good idea, thanks.

pkg/router/router.go
```diff
@@ -294,24 +294,39 @@ func (r *router) Serve(ctx context.Context) error {
...
+				}
+				r.logger.WithError(err).Error("Stopped reading packets due to unexpected error.")
 				return
 			}
 
-			r.logger.Warnf("Failed to handle transport frame: %v", err)
+			go func(packet routing.Packet) {
+				if err := r.handleTransportPacket(ctx, packet); err != nil {
+					if err == transport.ErrNotServing {
+						r.logger.WithError(err).Warnf("Stopped serving Transport.")
```
evanlinjin: Sorry for being annoying, But it might be a good idea to put this log line in the `once` as well.

evanlinjin (approved):

pkg/router/router.go
nkryuchkov: @evanlinjin Not annoying at all! I agree.

PR #229, opened 2020-03-18
Update README.md

jdknives:
Updating the documentation to improve readability ease of getting started.


README.md
```diff
@@ -156,301 +132,6 @@ Then you can start a hypervisor with:
 $ hypervisor 
 ```
 
-### Apps
```
evanlinjin, line 159: Why have we gotten rid of this section?

evanlinjin: Also, isn't the `# Transports` and `# App programming API` sections also needed?

jdknives: Added the transport section back in. The API section needs to be fixed afaics.

evanlinjin: @jdknives yes, these need to be updated. Should we do that in another PR/ticket?

PR #231, opened 2020-03-19
Replace unix pipes with TCP sockets

### LINKED ISSUE #230: Replace Unix Pipes with TCP sockets
**Feature description**

We should replace the unix pipes used for communication between `apps` and `visor` with TCP sockets. This is related to issue #138 . @skycoin-main said we should not worry about potential performance losses (if any) from this change. 

We will likely be unable to port Skywire to mobile with unix pipes. Also we could theoretically support windows with switching to tcp sockets. Although this is not a requirement or a deciding factor in this decision.

Darkren:
Yes
Fixes #230 	

 Changes:	
- Now app module uses TCP sockets instead of unix pipes
- Corresponding visor config field was changed
- Tests are adapted


pkg/app/appcommon/config.go
```diff
@@ -2,9 +2,10 @@ package appcommon
...
 // Config defines configuration parameters for `Proc`.
 type Config struct {
-	Name         string `json:"name"`
-	SockFilePath string `json:"sock_file_path"`
-	VisorPK      string `json:"visor_pk"`
-	BinaryDir    string `json:"binary_dir"`
-	WorkDir      string `json:"work_dir"`
+	Name       string `json:"name"`
+	ServerHost string `json:"server_host"`
+	ServerPort uint   `json:"server_port"`
```
evanlinjin: Should we just have `ServerAddr string`? I feel like it will make things simpler.

evanlinjin (changes requested):

pkg/app/appcommon/config.go
Darkren: Might be, agreed, will do

Darkren: Done

cmd/skywire-cli/commands/visor/gen-config.go
```diff
@@ -147,7 +147,7 @@ func defaultConfig() *visor.Config {
 
 	conf.Interfaces.RPCAddress = "localhost:3435"
 
-	conf.AppServerSockFile = "/tmp/visor_" + pk.Hex() + ".sock"
+	conf.AppServerAddr = ":5505"
```
nkryuchkov: Should we bind to `0.0.0.0` instead of `localhost` here? Probably it's also worth extracting a constant with the port as this magic number is widely used in the codebase.

cmd/skywire-visor/commands/root.go
```diff
@@ -196,9 +196,7 @@ func (cfg *runCfg) runVisor() *runCfg {
 	time.Sleep(startDelay)
 
 	if cfg.conf.DmsgPty != nil {
-		err = visor.UnlinkSocketFiles(cfg.conf.AppServerSockFile, cfg.conf.DmsgPty.CLIAddr)
-	} else {
-		err = visor.UnlinkSocketFiles(cfg.conf.AppServerSockFile)
+		err = visor.UnlinkSocketFiles(cfg.conf.DmsgPty.CLIAddr)
```
nkryuchkov: As the `else` statement is removed, the `err` variable is assigned not in all cases. E.g. if the line `185` returns an error and `cfg.conf.DmsgPty == nil` then the line `203` would be executed.

nkryuchkov (changes requested):

nkryuchkov (approved):
Looks great

PR #246, opened 2020-03-24
Use default values if required visor config fields are empty

### LINKED ISSUE #168: Config migration on binaries update
**Feature description**

> We may need to deal with updating scenarios, where new config values are required for visor to run. In that case we will need to update the config file before restart of the visor after an update.

_Originally posted by @jdknives in https://github.com/SkycoinProject/skywire-mainnet/issues/129#issuecomment-588314046_

https://github.com/SkycoinProject/skywire-mainnet/pull/153 implements support for updating `skywire-visor`, `skywire-cli` and app binaries. However, the scenario with new config values was not implemented.

nkryuchkov:

Fixes #168
Depends on https://github.com/SkycoinPro/skywire-services/pull/127

 Changes:	
- Use default values if required visor config fields are empty
- Cleanup visor config and related configs

How to test this PR:
- `make test`
- Run integration env, then check if nothing is broken
- Remove visor keys from integration env configs, run integration env. Scripts for checking communication should be broken. Visor logs should contain messages `INFO []: Starting visor PK=` with generated PKs instead of empty PKs. Restart integration env and check if generated PKs changed.
- Generate configs with `skywire-cli`:
  - `./skywire-cli visor gen-config -o ./test.json -r -m WD`
  - `./skywire-cli visor gen-config -o ./test.json -r -m HOME`
  - `./skywire-cli visor gen-config -o ./test.json -r -m LOCAL`


evanlinjin (approved):
Good work.

pkg/visor/config.go
```diff
@@ -1,63 +1,49 @@
...
-		SetupNodes         []cipher.PubKey `json:"setup_nodes"`
-		RouteFinder        string          `json:"route_finder"`
-		RouteFinderTimeout Duration        `json:"route_finder_timeout,omitempty"`
-	} `json:"routing"`
-
-	Uptime struct {
-		Tracker string `json:"tracker"`
-	} `json:"uptime"`
+	Version       string               `json:"version"`
+	Visor         *KeyPair             `json:"visor"`
```
Darkren: Can we rename it to something like `KeyPair` or `Keys`? I mean the struct field itself. Cause it's like visor config having visor field. Which is just a pair of keys. Looks confusing

Darkren (approved):
Great job! Just a small question in there

pkg/visor/config.go
nkryuchkov: @Darkren That's a good idea, had this in mind and forgot to do. However, I'm going to leave it in JSON as `visor` for now to not break configs. There're other names in the config that should also be changed. I'm going to come back to this later.

PR #252, opened 2020-03-25
Make the manager open the terminal with the same protocol

Senyoret1:
Go code was not changed.

 Changes:	
-	Now the manager opens the terminal using the same protocol in which the manager itself was openned (http or https).

How to test this PR:
Use the manager to open the full terminal.

### DIFF
--- a/static/skywire-manager-src/src/app/components/pages/node/actions/actions.component.ts
+++ b/static/skywire-manager-src/src/app/components/pages/node/actions/actions.component.ts
@@ -240,8 +240,9 @@ export class ActionsComponent implements AfterViewInit, OnDestroy {
     SelectOptionComponent.openDialog(this.dialog, options).afterClosed().subscribe((selectedOption: number) => {
       if (selectedOption === 1) {
         // Open the complete terminal in a new tab.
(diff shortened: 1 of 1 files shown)


evanlinjin (approved):

jdknives:
Both @Darkren and me tried to test this PR, but still the `dmsgpty` is opened with `https` regardless of whether `TLS` field is enabled or disabled in `hypervisor` config. I rebuilt the UI before testing. @Senyoret1

Senyoret1:
@jdknives
I just tested it and it worked. Here is what  did:
- Compiled the manager with `npm run build`.
- Copied the src folder from the location were it was compiled to the location were the hypervisor is loading the manager.
- Deactivated TLS in the hypervisor config.
- Ran the hypervisor (I used skywire-services).
- Openned the managet using http.
- Openend the terminal using the manager.

Following those steps the terminal was openned with http instead of https. Testing with the test server (not compiling the manager but using `npm start`) won't work, as the test server is configured to use https by default, no matter if TLS is active in the hypervisor instance or not.

jdknives:
Alright. Might have been my mistake somewhere. Works now. Great.

PR #256, opened 2020-03-26
Generate PubKey from SecKey if only SecKey is set in Visor config

### LINKED ISSUE #255: Auto-generate PubKey if only SecKey is specified
**Feature description**
Related to discussions around https://github.com/SkycoinProject/skyflash/issues/7

`skywire-visor` should generate the `PubKey` of a given `SecKey` if only the `SecKey` is specified in the config file of a `visor`. 

That will make it easier to setup the flashing tool we are writing for mainnet.

nkryuchkov:

Fixes #255

Depends on https://github.com/SkycoinPro/skywire-services/pull/128

 Changes:	
- Generate PubKey from SecKey if only SecKey is set in Visor config
- Change `static_secret_key` to `secret_key` in Visor JSON config.
- Change `static_public_key` to `public_key` in Visor JSON config.
- If a necessary config field is not set and default config value is generated, then the value is updated in the JSON config.

How to test this PR:
- Run generic integration env and check if Visor PK is generated correctly for the following conditions:
  - If both PK and SK are set in config, PK from config should be shown.
  - If only SK is set in config, the same PK should be shown.
  - If only PK is set in config, a newly generated PK should be shown.
  - If no keys are set in config, a newly generated PK should be shown.


jdknives:
Should the PubKey be added to the config itself when it is generated? @nkryuchkov

jdknives:
Also, when the secret key is empty and public key is empty, I get the following error: 
```
FATAL [skywire]: Failed to decode &{%!s(*os.file=&{{{0 0 0} 7 {0} <nil> 0 1 true true true} /Users/rudikleine/go/src/github.com/SkycoinProject/skywire-mainnet/skywire-config.json <nil> false false false})}: Invalid secret key length
```

Config looks like this: 
```
	"key_pair": {
		"static_public_key": "",
		"static_secret_key": ""
	}
```

nkryuchkov:
@jdknives 
> Should the PubKey be added to the config itself when it is generated?

Yes, because if it is not set, it will be generated from SecKey next time `Keys()` is called. Also, `cipher.GenerateKeyPair()` generates both keys, so I think it's better to fill both at one time.

nkryuchkov:
@jdknives 
> Also, when the secret key is empty and public key is empty, I get the following error:

In this case, keys are considered malformed, because the `key_pair` section is present in the config. However, I think it might be a good idea to generate new keys if they are malformed.

jdknives (approved):

PR #257, opened 2020-03-26
Fix dmsgpty in manager UI

### LINKED ISSUE #248: Cannot use dmsgpty in manager UI
**Describe the bug**
I cannot run the dmsgpty-terminal in the UI. I setup a visor and a hypervisor, accessed the UI and went to Terminal>Full Terminal. 

It returns `SSL_ERROR_RX_RECORD_TOO_LONG`. The issue is that the UI tries to call https://localhost. When I enter the URL without https and use http instead, everything runs smoothly. 

I rememeber distantly, that we discussed this issue before. It needs to be fixed either by making https work or by disabling https.

Darkren:
Yes
Fixes #248 

 Changes:	
- Cookie config now doesn't contain `HTTPOnly`, `Secure` and `SameSite` fields. `SameSite` is not allowed to be changed for simplicity, `HTTPOnly` and `Secure` are calculated from the `EnableTLS` value of hypervisor's config

### DIFF
--- a/pkg/hypervisor/config.go
+++ b/pkg/hypervisor/config.go
@@ -146,18 +146,32 @@ type CookieConfig struct {
 
 	ExpiresDuration time.Duration `json:"expires_duration"` // Used for determining the 'expires' value for cookies.
 
-	Path     string        `json:"path"`   // optional
-	Domain   string        `json:"domain"` // optional
-	Secure   bool          `json:"secure"`
-	HTTPOnly bool          `json:"http_only"`
-	SameSite http.SameSite `json:"same_site"`
+	Path   string `json:"path"`   // optional
+	Domain string `json:"domain"` // optional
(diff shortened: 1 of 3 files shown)


Senyoret1:
@Darkren I made some tests and was able to run the terminal with tls and auth active with the change made on https://github.com/SkycoinProject/skywire-mainnet/pull/258 . I'm not very familiar with the go code, so it would be good if you can take a look at that PR and check what can be done with it and what you have investigated. Also, please take a quick look at https://github.com/SkycoinProject/skywire-mainnet/issues/248#issuecomment-604092148

jdknives (approved):
Works for me. Well done. Together with Senyorets fix, this seems to solve the TLS/https issue.

nkryuchkov (approved):

PR #258, opened 2020-03-26
Make the session cookie work with the terminal

### LINKED ISSUE #248: Cannot use dmsgpty in manager UI

Senyoret1:
Yes

Fixes #248 

 Changes:	
- Now the `Path` param, set in the hypervisor configuration file, is added to the session cookie. This makes the web browser start sending the session cookie to the terminal window while using the default config, as previously web browsers were sending the cookie only to the `/api` paths (default browser behavior if the `Path` param is not set). This makes the terminal work with https connections while auth is activated.

How to test this PR:
Open the full terminal using the manager while auth and tls are active. See note below.

### DIFF
--- a/pkg/hypervisor/user_manager.go
+++ b/pkg/hypervisor/user_manager.go
@@ -297,6 +297,7 @@ func (s *UserManager) newSession(w http.ResponseWriter, session Session) error {
 	http.SetCookie(w, &http.Cookie{
 		Name:     sessionCookieName,
 		Value:    value,
+		Path:     s.c.Path,
 		Domain:   s.c.Domain,
 		Expires:  time.Now().Add(s.c.ExpiresDuration),
 		Secure:   s.c.Secure,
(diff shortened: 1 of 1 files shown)


Senyoret1:
There is a small issue with this: if you use the manager without the change, the hypervisor will send a cookie without the path and the browser will save it; if you then add the changes, the hypervisor will send the new cookie with the path and the browser will send both cookies with every request, which will make the authentication process to fail. To solve this, erase the cookies on the browser or wait until the old cookie expires (I think it last something like 2 days)

jdknives (changes requested):
The intended change works as expected. I can successfully open the dmsgpty terminal with https when TLS is enabled. However, the overview page cannot be loaded because the uptime request fails:
```
2020/03/27 11:00:16 "GET https://localhost:8080/api/visors/02980b3933c56b68193efed1bb2f6438d20c381e0d34a29829ac91040d5e6f4a23/uptime HTTP/2.0" from 127.0.0.1:64574 - 500 59B in 877.671997ms
```
We recently changed the config name of the uptime tracker. Maybe merging latest changes from develop into your branch will solve this issue.

PR #274, opened 2020-03-28
Make the manager work with http while using the test server

Senyoret1:
Go code was not changed.

 Changes:	
- The `npm run start-no-ssl` starts the dev server for checking the manager, but while using it the manager was not able to connect to the hypervisor instance, no mater if the hypervisor was running with or without tls. Now it is possible to use that command to open the manager and connect it to a hypervisor instance which is running without tls. It is still possible to use `npm start` to run the test server and connect the manager to a hypervisor instance which is running with tls.

- The proxy configuration of the test server was changed to make the manager connect to a hypervisor instance running on `127.0.0.1:8080`.

- The documentation was updated.

How to test this PR:
Instruction on the readme file


jdknives (approved):
Looks good from my side.

Issue #237, opened 2020-03-22
Data races in router


Darkren:
**Describe the bug**
Found data race on `make test`:

```
==================
WARNING: DATA RACE
Read at 0x00c0003f4558 by goroutine 275:
  github.com/SkycoinProject/skywire-mainnet/pkg/router.(*RouteGroup).sendKeepAlive()
      /Users/darkrengarius/go/src/github.com/SkycoinPro/skywire-mainnet/pkg/router/route_group.go:379 +0xcd
  github.com/SkycoinProject/skywire-mainnet/pkg/router.(*RouteGroup).keepAliveLoop()
      /Users/darkrengarius/go/src/github.com/SkycoinPro/skywire-mainnet/pkg/router/route_group.go:370 +0x2e6

Previous write at 0x00c0003f4558 by goroutine 291:
  [failed to restore the stack]

Goroutine 275 (running) created at:
  github.com/SkycoinProject/skywire-mainnet/pkg/router.NewRouteGroup()
      /Users/darkrengarius/go/src/github.com/SkycoinPro/skywire-mainnet/pkg/router/route_group.go:126 +0x777
  github.com/SkycoinProject/skywire-mainnet/pkg/router.(*router).saveRouteGroupRules()
      /Users/darkrengarius/go/src/github.com/SkycoinPro/skywire-mainnet/pkg/router/router.go:375 +0x4dd
  github.com/SkycoinProject/skywire-mainnet/pkg/router.(*router).DialRoutes()
      /Users/darkrengarius/go/src/github.com/SkycoinPro/skywire-mainnet/pkg/router/router.go:234 +0xb17
  github.com/SkycoinProject/skywire-mainnet/pkg/router.Test_router_DialRoutes()
      /Users/darkrengarius/go/src/github.com/SkycoinPro/skywire-mainnet/pkg/router/router_test.go:97 +0xf58
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:909 +0x199
==================
==================
WARNING: DATA RACE
Read at 0x00c0003f4570 by goroutine 275:
  github.com/SkycoinProject/skywire-mainnet/pkg/router.(*RouteGroup).sendKeepAlive()
      /Users/darkrengarius/go/src/github.com/SkycoinPro/skywire-mainnet/pkg/router/route_group.go:379 +0x4a3
  github.com/SkycoinProject/skywire-mainnet/pkg/router.(*RouteGroup).keepAliveLoop()
[...]

PR #278, opened 2020-03-30
VPN apps

### LINKED ISSUE #100: Implement VPN application
We want to implement a VPN application for Skywire. That VPN application should consist of a VPN Client and a VPN server. Initially we should build a routed VPN that operates through a TUN adapter. The traffic should be encrypted by default.

Darkren:

Fixes #100 , #237 

 Changes:	
- Implemented VPN server
- Implemented VPN client
- Modified apps to accept custom envs on start
- Fixed datarace in router


cmd/apps/vpn-client/vpn-client.go
```diff
@@ -0,0 +1,363 @@
...
+	"syscall"
+	"time"
+
+	"github.com/SkycoinProject/dmsg/cipher"
+	"github.com/SkycoinProject/skycoin/src/util/logging"
+	"github.com/SkycoinProject/skywire-mainnet/internal/netutil"
+	"github.com/SkycoinProject/skywire-mainnet/internal/vpn"
+	"github.com/SkycoinProject/skywire-mainnet/pkg/app"
+	"github.com/SkycoinProject/skywire-mainnet/pkg/routing"
+
```
nkryuchkov: Could you please group the imports?

cmd/apps/vpn-client/vpn-client.go
```diff
@@ -0,0 +1,363 @@
...
+	log.Debugf("Got server hello: %v", sHello)
+
+	if sHello.Status != vpn.HandshakeStatusOK {
+		return nil, nil, fmt.Errorf("got status %d (%s) from the server", sHello.Status, sHello.Status)
+	}
+
+	return sHello.TUNIP, sHello.TUNGateway, nil
+}
+
+func main() {
```
nkryuchkov, line 60: In my opinion, `main` looks hard to read. I would consider splitting it into multiple functions.

cmd/apps/vpn-server/vpn-server.go
```diff
@@ -0,0 +1,131 @@
...
+	appName = "vpn-server"
+	netType = appnet.TypeSkynet
+	vpnPort = routing.Port(44)
+)
+
+var (
+	log = app.NewLogger(appName)
+)
+
+func main() {
```
nkryuchkov, line 28: I would consider splitting it into multiple functions.

cmd/apps/vpn-server/vpn-server.go
```diff
@@ -0,0 +1,131 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"os/signal"
+	"syscall"
+
+	"github.com/SkycoinProject/skywire-mainnet/internal/vpn"
```
nkryuchkov: Could you please group imports?

cmd/apps/vpn-server/vpn-server.go
```diff
@@ -0,0 +1,131 @@
...
+		}
+	}()
+
+	appCfg, err := app.ClientConfigFromEnv()
+	if err != nil {
+		log.WithError(err).Errorln("Error getting app client config")
+		return
+	}
+
+	appClientt, err := app.NewClient(logging.MustGetLogger(fmt.Sprintf("app_%s", appName)), appCfg)
```
nkryuchkov: `appClientt` => `appClient`

internal/vpn/env.go
```diff
@@ -0,0 +1,123 @@
...
+	// STCPTableLenEnvKey is env arg holding Stcp table length.
+	STCPTableLenEnvKey = "STCP_TABLE_LEN"
+	// STCPKeyEnvPrefix is prefix for each env arg holding STCP entity key.
+	STCPKeyEnvPrefix = "STCP_TABLE_KEY_"
+	// STCPValueEnvPrefix is prefix for each env arg holding STCP entity value.
+	STCPValueEnvPrefix = "STCP_TABLE_"
+)
+
+// AppEnvArgs forms env args to pass to the app process.
+func AppEnvArgs(dmsgDiscovery, tpDiscovery, rf, uptimeTracker string,
```
nkryuchkov: Would you consider passing a struct instead of that arg list?

internal/vpn/net.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+// ReadJSON reads portion of data from the `conn` and unmarshals it into `data`.
+func ReadJSON(conn net.Conn, data interface{}) error {
+	const bufSize = 1024
+
+	var dataBytes []byte
+	buf := make([]byte, bufSize)
+	for {
+		n, err := conn.Read(buf)
```
nkryuchkov, line 35: Could `ioutil.ReadAll` be used here instead?

internal/vpn/os.go
```diff
@@ -0,0 +1,204 @@
...
+
+// SetupTUN sets the allocated TUN interface up, setting its IP, gateway, netmask and MTU.
+func SetupTUN(ifcName, ip, netmask, gateway string, mtu int) error {
+	return run("/sbin/ifconfig", ifcName, ip, gateway, "mtu", strconv.Itoa(mtu), "netmask", netmask, "up")
+}
+
+// NetworkIfcGateway gets gateway of the network interface with name `ifcName`.
+func NetworkIfcGateway(ifcName string) (net.IP, error) {
+	cmd := fmt.Sprintf(gatewayForIfcCMDFmt, ifcName)
+	outBytes, err := exec.Command("/bin/bash", "-c", cmd).Output() //nolint:gosec
```
nkryuchkov: Would `/bin/sh` instead of `/bin/bash` work? Sometimes, `bash` is not pre-installed.

internal/vpn/os.go
```diff
@@ -0,0 +1,204 @@
...
+	"strconv"
+)
+
+// SetupTUN sets the allocated TUN interface up, setting its IP, gateway, netmask and MTU.
+func SetupTUN(ifcName, ip, netmask, gateway string, mtu int) error {
+	return run("/sbin/ifconfig", ifcName, ip, gateway, "mtu", strconv.Itoa(mtu), "netmask", netmask, "up")
+}
+
+// NetworkIfcGateway gets gateway of the network interface with name `ifcName`.
+func NetworkIfcGateway(ifcName string) (net.IP, error) {
```
nkryuchkov: What do you think about renaming to `NetworkInterfaceGateway` to improve readability?

internal/vpn/server.go
```diff
@@ -0,0 +1,183 @@
...
+	if err != nil {
+		sHello.Status = HandshakeStatusInternalError
+		if err := WriteJSON(conn, &sHello); err != nil {
+			s.log.WithError(err).Errorln("Error sending server hello")
+		}
+
+		return nil, nil, fmt.Errorf("error breaking IP into octets: %w", err)
+	}
+
+	sTUNIP := net.IPv4(subnetOctets[0], subnetOctets[1], subnetOctets[2], subnetOctets[3]+2)
```
nkryuchkov, line 242: Could you please add more comments about the logic of `+2`, `+1`, `+4`, `+3` here and below? I think it may increase readability.

pkg/app/appserver/proc.go
```diff
@@ -34,7 +34,8 @@ type Proc struct {
 }
 
 // NewProc constructs `Proc`.
-func NewProc(log *logging.Logger, c appcommon.Config, args []string, stdout, stderr io.Writer) (*Proc, error) {
+func NewProc(log *logging.Logger, c appcommon.Config, args []string, envs map[string]string,
```
nkryuchkov: I think we should put these args into a struct as they got large.

nkryuchkov (commented):
Great job! Could you please remove the committed `vpn-client` binary?

internal/vpn/net.go
Darkren: @nkryuchkov unfortunately it won't work. `ReadAll` keeps reading until `EOF` or some other error occurs. but in our case `EOF` won't happen, since it rises only if remote got closed or empty data received from it. which never happens because of how route group is implemented

nkryuchkov: @Darkren I see. Thank you for the explanation.

pkg/app/appserver/proc.go
Darkren: @nkryuchkov I've seen @evanlinjin changed this signature in one of his PRs too. I think it's simpler to merge it as is and raise a separate issue to simplify this part. which could be tackled right after this PR and Evan's are merged

nkryuchkov (approved):

go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/SkycoinProject/skywire-mainnet
 go 1.13
 
 require (
-	github.com/SkycoinProject/dmsg v0.2.0
+	github.com/SkycoinProject/dmsg v0.1.1-0.20200420091742-8c1a3d828a49
```
evanlinjin, line 6: Is the accidental or is there a problem with `v0.2.0`?

cmd/apps/vpn-client/vpn-client.go
```diff
@@ -0,0 +1,111 @@
...
+	"net"
+	"os"
+	"os/signal"
+	"syscall"
+	"time"
+
+	"github.com/SkycoinProject/dmsg/cipher"
+	"github.com/SkycoinProject/skycoin/src/util/logging"
+
+	"github.com/SkycoinProject/skywire-mainnet/internal/netutil"
```
evanlinjin: I've moved `netutil` to https://github.com/SkycoinProject/dmsg/tree/master/netutil

Let's get rid of the `skywire-mainnet/internal/netutil` package.

cmd/apps/vpn-client/vpn-client.go
```diff
@@ -0,0 +1,111 @@
...
+	"github.com/SkycoinProject/skywire-mainnet/internal/vpn"
+	"github.com/SkycoinProject/skywire-mainnet/pkg/app"
+	"github.com/SkycoinProject/skywire-mainnet/pkg/app/appnet"
+	"github.com/SkycoinProject/skywire-mainnet/pkg/routing"
+)
+
+const (
+	appName = "vpn-client"
+	netType = appnet.TypeSkynet
+	vpnPort = routing.Port(44)
```
evanlinjin: We can use the `skyenv` value here.

internal/vpn/ip_generator.go
```diff
@@ -0,0 +1,88 @@
...
+			continue
+		}
+
+		return ip, nil
+	}
+
+	return nil, errors.New("no free IPs left")
+}
+
+func fetchIPv4Octets(ip net.IP) ([4]uint8, error) {
```
evanlinjin, line 65: Good job implementing the logic here. However, can't we just use the `(net.IP).To4()` function directly and `copy` the result to the output (or return an error if `.To4()` returns `nil`)?

internal/vpn/os_linux.go
```diff
@@ -0,0 +1,68 @@
...
+package vpn
+
+import (
+	"bytes"
+	"fmt"
+	"os/exec"
+)
+
+const (
+	gatewayForIfcCMDFmt         = "/sbin/route -n | /bin/grep %s | /usr/bin/awk '$1 == \"0.0.0.0\" {print $2}'"
```
evanlinjin: Seems the command `route` is deprecated: https://unix.stackexchange.com/questions/83096/what-is-the-difference-between-sbin-ip-route-and-sbin-route

For example, on Manjaro, this command does not exist.

Also, not all distros put binaries in `/sbin`. Would be be able to do: `$(command -v ip)` (assuming we are to use the `ip` command).

evanlinjin (commented):
Good job making this relatively easy to follow (considering what it is).

I'm not going to pretend I know how TUN/STUN is meant to work, but I've listed my concerns below.

go.mod
Darkren: @evanlinjin no, we've got my PR merged into Dmsg that I need here. but that PR is not included into the released version unfortunately, so I had to do this

internal/vpn/ip_generator.go
Darkren: well spot!

internal/vpn/os_linux.go
Darkren: @evanlinjin we still have to maintain such commands for Mac systems at least. directory with the binary is actually a problem, since `PATH` seems to be lost when we try to execute command from the process which is itself a command executed by another process. I'll try to deal with this within this PR. but regarding moving to `ip` command set for Linux systems, I'll put this point into https://github.com/SkycoinProject/skywire-mainnet/issues/347 , I think it's best to be done there

evanlinjin: > @evanlinjin we still have to maintain such commands for Mac systems at least. directory with the binary is actually a problem, since `PATH` seems to be lost when we try to execute command from the process which is itself a command executed by another process. I'll try to deal with this within this PR. but regarding moving to `ip` command set for Linux systems, I'll put this point into #347 , I think it's best to be done there

@Darkren I thought `command` was a shell built-in? You can confirm this with `command -V command`. And sure, handling the `ip` situation elsewhere is fine.

Darkren: @evanlinjin tried running `$(command -v ifconfig)` got PATH error. so, just passed correct PATH to the app cmd, seems to work fine with just the binary name

PR #293, opened 2020-04-02
Embed static files in hypervisor binary.

evanlinjin:
This will make make life easier for end users and Skybian maintainers. 

Changes:
* Embed static files in hypervisor.
* Provide `/api/about` endpoint for hypervisor that displays hypervisor public key and build info.


Makefile
```diff
@@ -145,7 +145,9 @@ lint-ui:  ## Lint the UI code
 
 build-ui:  ## Builds the UI
 	cd $(MANAGER_UI_DIR) && npm run build
-
+	mkdir -p ./bin
+	${OPTS} GOBIN=${PWD}/bin go get github.com/rakyll/statik
+	./bin/statik -src=$(MANAGER_UI_DIR)/dist -dest ./cmd/hypervisor -f
```
nkryuchkov: Is it possible to change it to the same as above (`${PWD}/bin`) or to change above to the same as this line (`GOBIN=./bin`)?

cmd/hypervisor/commands/root.go
```diff
@@ -47,11 +49,15 @@ var rootCmd = &cobra.Command{
 		if _, err := buildinfo.Get().WriteTo(os.Stdout); err != nil {
 			log.Printf("Failed to output build info: %v", err)
 		}
-
```
nkryuchkov, line 50: I think the new line here looks better because it split logical parts of the function: getting build info and preparing config.

pkg/hypervisor/hypervisor.go
```diff
@@ -54,13 +54,14 @@ type VisorConn struct {
...
 	c      Config
+	assets http.FileSystem             // Web UI.
 	visors map[cipher.PubKey]VisorConn // connected remote visors.
 	users  *UserManager
 	mu     *sync.RWMutex
 }
 
 // New creates a new Hypervisor.
-func New(config Config) (*Hypervisor, error) {
+func New(assets http.FileSystem, config Config) (*Hypervisor, error) {
```
nkryuchkov, line 64: What do you think of putting `assets` as the second argument? In my opinion, `config` always should be the first.

nkryuchkov (approved):
It works great, good job!

Makefile
evanlinjin: `GOBIN=./bin` is not allowed. `golang` only accepts absolute paths for `GOBIN` FYI.
I'll do it the other way.

pkg/hypervisor/hypervisor.go
evanlinjin: @nkryuchkov I've always had the habit of putting `config` as the last argument. 😅

Makefile
nkryuchkov: @evanlinjin You're right, I forgot about that. Then, I'd change all `./bin`s to `${PWD}/bin`s then for consistency and readability.

PR #311, opened 2020-04-08
Added app config fields to RPC.Apps call.

### LINKED ISSUE #270: Add an API endpoint to the hypervisor for checking the config of the apps
**Feature description**
It is possible to change the config of the apps with the `/api/visor/{key}/apps/{app}` API endpoint, but the hypervisor does not have an endpoint for checking the current config of an app. This means that the manager UI can show options for changing the config, but the user will always see empty forms, which is not very good for the UX, as the expected behavior is to see the form filled with the current config.

Add an API endpoint for checking the current config of an app. However, some care would be needed with sensitive data, like passwords.

evanlinjin:
Fixes #270

### DIFF
--- a/cmd/skywire-cli/commands/visor/app.go
+++ b/cmd/skywire-cli/commands/visor/app.go
@@ -41,7 +41,7 @@ var lsAppsCmd = &cobra.Command{
 			if state.Status == visor.AppStatusRunning {
 				status = "running"
 			}
-			_, err = fmt.Fprintf(w, "%s\t%s\t%t\t%s\n", state.Name, strconv.Itoa(int(state.Port)), state.AutoStart, status)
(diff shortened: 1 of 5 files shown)


evanlinjin:
@Senyoret1 Now the `GET /api/visors/{visor-pk}/apps` call responds as follows:
```
$ curl -k "https://localhost:8000/api/visors/024ec47420176680816e0406250e7156465e4531f5b26057c9f6297bb0303558c7/apps" | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    89  100    89    0     0   6357      0 --:--:-- --:--:-- --:--:--  6357
[
  {
    "app": "skysocks",
    "auto_start": true,
    "port": 3,
    "args": [
      "-passcode",
      "123456"
    ],
    "status": 1
  }
]
```

jdknives (approved):
Tested in deployment and integration env.

Darkren (approved):

PR #318, opened 2020-04-13
Proxy discovery client.

evanlinjin:


pkg/proxydisc/client.go
```diff
@@ -0,0 +1,199 @@
...
+func (c *HTTPClient) addr(path string) string {
+	return c.conf.DiscAddr + path
+}
+
+// Auth returns the internal httpauth.Client
+func (c *HTTPClient) Auth(ctx context.Context) (*httpauth.Client, error) {
+	if c.auth != nil {
+		return c.auth, nil
+	}
+	fmt.Println(c.conf.DiscAddr)
```
Darkren: Is it intended ot was left after debugging?

Darkren (commented):
Great job!

pkg/proxydisc/client.go
evanlinjin: Left after debugging. Thanks for the heads up.

pkg/proxydisc/error.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/sirupsen/logrus"
+)
+
+// HTTPError represents an HTTP error.
+type HTTPError struct {
+	Status int    `json:"http_status"` // HTTP Status.
```
nkryuchkov: Is there a reason why struct and JSON names differ?

nkryuchkov (approved):
The code looks fine, great job!

pkg/proxydisc/error.go
evanlinjin: No reason. I'll change it.

PR #332, opened 2020-04-18
Sanitize the URLs used by the manager

Senyoret1:

 Changes:	
- Code was added to avoid the manager to try to use URLs with two `/` characters in the path. Example, if `http://first/second//third` is detected, the code calls `http://first/second/third`.

How to test this PR:
The URL used to connect to the `exec` endpoint in the simple terminal now should work well.


jdknives (approved):
Nice. Exec endpoint works again.

PR #340, opened 2020-04-21
Proxy discovery.

evanlinjin:


pkg/app/appdisc/discovery_updater.go
```diff
@@ -0,0 +1,68 @@
...
+	if u.cancel == nil {
+		return
+	}
+
+	u.cancel()
+	u.cancel = nil
+	u.wg.Wait()
+}
+
+func (u *proxyUpdater) ChangeConnCount(delta int) {
```
Darkren: Could you please add comments on this? It's not really clear from the naming what this method is supposed to do. And also, shouldn't this be implemented at this point?

pkg/app/appdisc/factory.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/sirupsen/logrus"
+
+	"github.com/SkycoinProject/skywire-mainnet/pkg/app/appcommon"
+	"github.com/SkycoinProject/skywire-mainnet/pkg/proxydisc"
+	"github.com/SkycoinProject/skywire-mainnet/pkg/skyenv"
+)
+
+// Factory creates appdisc.Updater instances based on the app name.
+type Factory struct {
+	Log            logrus.FieldLogger
```
Darkren, line 18: Shouldn't these fields be unexported for safety?

Darkren (changes requested):
Overall great job! Just few small queries. Also, `mock_proc_manager` shouldn't be removed as it'll be generated with the old name once we do `make generate`. And it also seems more conventional to me

Makefile
```diff
@@ -102,6 +102,7 @@ install-linters: ## Install linters
 	${OPTS} go get -u golang.org/x/tools/cmd/goimports
 
 format: ## Formats the code. Must have goimports installed (use make install-linters).
+	${OPTS} go mod tidy && go mod vendor
```
nkryuchkov: Why is vendoring dependencies also necessary when formatting the project?

pkg/app/appdisc/factory.go
```diff
@@ -0,0 +1,77 @@
...
+		}, true
+
+	default:
+		return &emptyUpdater{}, false
+	}
+}
+
+func containsFlag(args []string, flag string) bool {
+	for _, arg := range args {
+		if strings.HasPrefix(arg, "-") && strings.Contains(arg, flag) {
```
nkryuchkov: I think this may work incorrectly if we add a flag containing the `passcode` substring in the future. What do you think about making the change below?
```suggestion
		if arg == "-" + flag {
```

pkg/app/appserver/proc.go
```diff
@@ -10,6 +10,8 @@ import (
 	"sync"
 	"sync/atomic"
 
+	"github.com/SkycoinProject/skywire-mainnet/pkg/app/appdisc"
```
nkryuchkov: Could you please group this line with #17?

pkg/app/appserver/proc_manager.go
```diff
@@ -8,6 +8,8 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/SkycoinProject/skywire-mainnet/pkg/app/appdisc"
```
nkryuchkov: Could you please group imports?

pkg/app/appserver/server.go
```diff
@@ -34,10 +36,10 @@ func New(log *logging.Logger, addr string) *Server {
 
 // Register registers an app key in RPC server.
 func (s *Server) Register(appKey appcommon.Key) error {
-	logger := logging.MustGetLogger(fmt.Sprintf("app_gateway:%s", appKey))
-	gateway := NewRPCGateway(logger)
+	logger := logging.MustGetLogger(fmt.Sprintf("app_gw:%s", appKey))
```
nkryuchkov: I would leave it as `app_gateway` because `gw` is not clear in the logs.

pkg/proxydisc/client.go
```diff
@@ -176,7 +176,11 @@ func (c *HTTPClient) UpdateLoop(ctx context.Context, updateInterval time.Duratio
 				time.Sleep(time.Second * 10) // TODO(evanlinjin): Exponential backoff.
 				continue
 			}
-			c.log.WithField("entry", entry).Debug("Entry updated.")
+			if entry.Geo != nil {
+				c.entry.Geo = entry.Geo
+			}
+			j, _ := json.Marshal(entry) //nolint:errcheck
```
nkryuchkov: Could you please check the error here?

Makefile
evanlinjin: I'm open for opinions. Please check my reasoning here: https://github.com/SkycoinPro/skywire-service-proxy-discovery/pull/12#discussion_r416265768

pkg/app/appdisc/factory.go
evanlinjin: Good spotting. However, CLI flags may be structured in many ways. I came up with the following solution instead:

```go
func containsFlag(args []string, flag string) bool {
	for _, arg := range args {
		if argEqualsFlag(arg, flag) {
			return true
		}
	}
	return false
}

func argEqualsFlag(arg, flag string) bool {
	arg = strings.TrimSpace(arg)

	// strip prefixed '-'s.
	for {
		if len(arg) < 1 {
			return false
		}
		if arg[0] == '-' {
			arg = arg[1:]
			continue
		}
		break
	}

	// strip anything after (inclusive) of '='.
	arg = strings.Split(arg, "=")[0]

	return arg == flag
}
```

pkg/app/appdisc/factory.go
evanlinjin: My thoughts are that `appdisc.Factory` acts as a *config* or sorts that generates discovery updaters. These fields are only read for generating such discovery updaters.

Hope that is okay?

PR #351, opened 2020-04-28
Add method to get visor's config value via RPC call

### LINKED ISSUE #345: Give app an ability to request config changes from the visor
**Feature description**

Apps should be able to request config changes from the visor. This may done as discussed on Discord:

```
i think the best solution is if we add a function in appserver.RPCGateway along the lines of func (g *RPCGateway) Value(name string, v interface{}) error
```
but also, thinking further, it might not work well with the nested values. probably should be thought of deeper

Darkren:
Yes

Fixes #345 

 Changes:	
- Switch to develop branch of Dmsg
- Add method to get visor's config value via RPC call


nkryuchkov (approved):

evanlinjin:
Because apps are meant to use this, shouldn't this be implemented here: https://github.com/SkycoinProject/skywire-mainnet/blob/master/pkg/app/appserver/rpc_gateway.go ?

evanlinjin:
Or are you thinking to expose the entirety of `visor.RPC` to apps?

Darkren:
@evanlinjin I was thinking to create the RPC client instance within the app, yes. cause appserver rn is only responsible for networking. and it doesn't even have access to the visor instance, so I was willing to keep it this way. but this is truly discussable

evanlinjin:
@Darkren Is this feature urgent right now? Do you mind if I am made responsible for continuing with this to find an elegant solution?

Darkren:
@evanlinjin I don't, just please let me know, when you're done. I'll need to integrate this within VPN apps

PR #356, opened 2020-05-01
VPN improvements

### LINKED ISSUE #343: Build and run on Windows
**Feature description**

It would be nice to have (although not essential) to be able to run and build mainnet on Windows. Currently, the Sirupsen package prevents that as it uses build tags that prevent it from building on Windows.

### LINKED ISSUE #347: VPN improvements
**Feature description**

A tentative list of improvements that need to be added to the VPN apps. 

- [x] add e2e encryption

- [x] add authentication via passcode

- [x] add RPC methods (change VPN server remote Pubkey, change password auth)

- [x] allow to use VPN client as local access point

- [x] add Windows client (depends on https://github.com/SkycoinProject/skywire-mainnet/issues/343)

- [x] move on to the `ip` command set for Linux systems - described in here https://github.com/SkycoinProject/skywire-mainnet/pull/278#discussion_r416737151

- [x] add `README.md`

- [x] there's a piece of code which searches for the default network interface. currently is invoked only on Mac systems. it works in most cases but not in all of them. code is unreliable and should be replaced

Darkren:
Yes

Fixes #347 Fixes #343 

 Changes:	
- Added passcode to authenticate VPN client within the server
- Added RPC calls to changes passcode for both client and server
- Added RPC call to change remote PK for client
- Modified hypervisor's `putApp` API method to perform the RPC calls mentioned above
- Added an ability to run this guy on Windows

How to test this PR:
- Passcode
1. Add `"-passcode", "some_passcode"` to the `args` parameter of both server and client configs
2. Try to connect. Correct password will result in successful traffic exchange, incorrect password will result in `Forbidden` status in client logs

- RPC calls
Try calling the hypervisor API to change passcode and PK. Passcode may be changed for both server and client. Functionality takes place in the `putApp` method, the same as used to change PK and passcode for skysocks apps


jdknives (commented):
Changes work well so far. Should the VPN app stop upon not successfully registering with a remote server?

jdknives (approved):
Works smoothly for me. I could not test the Windows setup unfortunately.

internal/vpn/client.go
```diff
@@ -112,26 +113,48 @@ func (c *Client) Serve() error {
...
+			return fmt.Errorf("failed to enable encryption: %w", err)
+		}
+
+		c.log.Infoln("Encryption enabled")
+	} else {
+		c.log.Infoln("Encryption disabled")
+	}
+
 	connToTunDoneCh := make(chan struct{})
 	tunToConnCh := make(chan struct{})
```
evanlinjin, line 145: Instead of using 2 channels here, I would use one channel and a `sync.Once` here, but I guess that is personal taste.

internal/vpn/errors.go
```diff
@@ -0,0 +1,7 @@
+package vpn
+
+import "errors"
+
+var (
+	errCouldFindDefaultNetworkGateway = errors.New("couldn't find default network gateway")
```
evanlinjin: *could not* should be used instead of *couldn't*.

internal/vpn/noise_credentials.go
```diff
@@ -0,0 +1,23 @@
...
+
+// NoiseCredentials encapsulates sec and pub keys for noise usage.
+type NoiseCredentials struct {
+	PK cipher.PubKey
+	SK cipher.SecKey
+}
+
+// IsValid returns true only if PK and SK are valid.
+func (c NoiseCredentials) IsValid() bool {
+	return !c.PK.Null() && !c.SK.Null()
```
evanlinjin, line 41: I feel like we should only use the `SK` and generate the `PK` from the `SK`. This way, one has no chance of providing a key pair that are not associated with one another.

internal/vpn/os_client_darwin.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"bytes"
+	"fmt"
+	"net"
+	"os/exec"
+)
+
+const (
+	defaultNetworkGatewayCMD = "netstat -rn | sed -n '/Internet/,/Internet6/p' | grep default | awk '{print $2}'"
```
evanlinjin, line 13: I would use **`** instead of **"** quotes to encapsulate the string. Might bring a better reading experience. Again, this is personal taste.

evanlinjin (commented):
Good work and clean code. Some of my comments are due to personal taste, if you don't agree with them, feel free to ignore it! 😅

evanlinjin (approved):

internal/vpn/os_client_darwin.go
evanlinjin: This comment is really for other `os_client_*.go` files with the same variable that resorts to using backslashes.

README.md
```diff
@@ -120,6 +121,44 @@ $ hypervisor
...
+
+Apps may be built the same way:
+
+`go build -o ./apps/vpn-client.exe ./cmd/apps/vpn-client`
+
+Apps should be stated in the config without `.exe` extension, as usual.
+
+Some log lines may seem strange due to the Windows encoding in terminal. To make life easier you may use:
+```
+REM change CHCP to UTF-8
```
nkryuchkov: Is this comment necessary?

internal/vpn/net.go
```diff
@@ -3,7 +3,13 @@ package vpn
 import (
 	"encoding/json"
 	"fmt"
+	"io"
 	"net"
+
+	"github.com/SkycoinProject/dmsg/cipher"
+
```
nkryuchkov: Could you please format imports?

pkg/visor/init.go
```diff
@@ -5,14 +5,14 @@ import (
...
 	"fmt"
 	"net"
-	"os"
-	"path/filepath"
 	"sync"
 	"time"
 
+	"github.com/SkycoinProject/dmsg/netutil"
+	"github.com/sirupsen/logrus"
+
```
nkryuchkov: Not sorted.

nkryuchkov (approved):
Great job!

README.md
jdknives: I think it is alright to mention that as otherwise the output is hard to read or use to debug.

nkryuchkov: @jdknives I think this could be added to line 138 as it's a part of the description. No need to execute a comment.

internal/vpn/noise_credentials.go
Darkren: Made this a bit different. this func is still the same, and SK may be nil if encryption is disabled. but in the `main` func of client and server I'm deriving PK out of SK if only SK is provided

internal/vpn/os_client_darwin.go
Darkren: it would've bring more readability if there were some some escaped quotes there, so we could remove escaping. but in this case it changes nothing, literally. but I applied this in `os_client_linux`, there were escaped quotes there. so thank you for this anyway

internal/vpn/client.go
Darkren: Personally I don't like these 2 channels much, but `sync.Once` due to its syntax may just produce more code. I'd stick to 2 channels in this case

PR #357, opened 2020-05-01
Appserver improvements.

### LINKED ISSUE #338: Improvements to appserver.
**Feature description**

Currently `appserver.RPCGateway` is registered in the `rpc.Server` for every app connection. This, although a quick solution at the time, results in the follow problems:
1. If apps disconnect/reconnect this will lead to memory leaks as we cannot *unregister* from an `rpc.Server`.
2. It becomes difficult to track number of multiplexed connections to the app and whether the app is connected with the same structure.

Additionally, the responsibilities of the `ProcManager` structure can be expanded, removing the need for `visor.Visor` to include app management logic.

- Only register one instance of `appserver.RPCGateway`
- For each RPC method, include field for inputting the *App Secret*.

evanlinjin:
Fixes #338

Changes:	
- Combined the `appserver.Server` and `appserver.ProcManager` concepts.
- Have one `rpc.Server` per app proc.
- Simplified `app.Client` initialization.
  - Visor to App initialization is done via a single env defined for the app [`PROC_CONFIG`](https://github.com/SkycoinProject/skywire-mainnet/pull/357/files#diff-b2002858cbe252d3f34287789a14d1e2R22) which contains a JSON object.
  - App to Visor initialization is done by App connecting the Visor via app server address. The first 16 bytes that the App sends to Visor is the [`appcommon.ProcKey`](https://github.com/SkycoinProject/skywire-mainnet/pull/357/files#diff-b2002858cbe252d3f34287789a14d1e2R31).
  - Changed `appcommon.ProcKey` format so that it is a deterministic length.
- Improved app logging and removed any possibilities of log database race conditions with `bbolt`.
- Various logging improvements.


cmd/apps/helloworld/helloworld.go
```diff
@@ -4,11 +4,11 @@ simple client server app for skywire visor testing
 package main
 
 import (
-	"log"
 	"os"
 
+	"github.com/sirupsen/logrus"
+
```
nkryuchkov: Could you please remove the empty line?

cmd/apps/skychat/chat.go
```diff
@@ -11,11 +11,13 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"os"
 	"sync"
 	"time"
 
+	"github.com/sirupsen/logrus"
+
```
nkryuchkov: Could you please remove the empty line?

cmd/apps/skychat/chat.go
```diff
@@ -34,40 +35,28 @@ var addr = flag.String("addr", ":8001", "address to bind")
 var r = netutil.NewRetrier(50*time.Millisecond, 5, 2)
 
 var (
-	chatApp   *app.Client
-	clientCh  chan string
-	chatConns map[cipher.PubKey]net.Conn
-	connsMu   sync.Mutex
-	log       *logging.MasterLogger
+	log      = logrus.New()
```
nkryuchkov, line 37: Why did you decide to use `logrus` directly instead of `logging` wrapper in your changes?

cmd/apps/skysocks-client/skysocks-client.go
```diff
@@ -5,13 +5,13 @@ package main
 
 import (
 	"flag"
-	"fmt"
 	"io"
 	"net"
 	"time"
 
+	"github.com/sirupsen/logrus"
+
```
nkryuchkov: Could you please remove the empty line?

cmd/apps/vpn-client/vpn-client.go
```diff
@@ -3,16 +3,16 @@ package main
...
 	"flag"
-	"fmt"
 	"net"
 	"os"
 	"os/signal"
 	"syscall"
 	"time"
 
+	"github.com/sirupsen/logrus"
+
```
nkryuchkov: Could you please remove the empty line?

internal/skysocks/client.go
```diff
@@ -7,14 +7,16 @@ import (
 	"sync"
 	"time"
 
+	"github.com/sirupsen/logrus"
+
```
nkryuchkov: Could you please remove the empty line?

internal/skysocks/server.go
```diff
@@ -5,7 +5,8 @@ import (
 	"net"
 	"sync/atomic"
 
-	"github.com/SkycoinProject/skycoin/src/util/logging"
+	"github.com/sirupsen/logrus"
+
```
nkryuchkov: Could you please remove the empty line?

internal/vpn/client.go
```diff
@@ -9,7 +9,8 @@ import (
 	"strconv"
 	"sync"
 
-	"github.com/SkycoinProject/skycoin/src/util/logging"
+	"github.com/sirupsen/logrus"
+
```
nkryuchkov: Could you please remove the empty line?

internal/vpn/server.go
```diff
@@ -7,15 +7,16 @@ import (
 	"net"
 	"sync"
 
-	"github.com/SkycoinProject/skycoin/src/util/logging"
+	"github.com/sirupsen/logrus"
+
```
nkryuchkov: Could you please remove the empty line?

pkg/app/appcommon/log_store.go
```diff
@@ -1,15 +1,41 @@
...
 	"bytes"
 	"fmt"
 	"io"
+	"os"
 	"strings"
+	"sync"
 	"time"
 
+	"github.com/SkycoinProject/skycoin/src/util/logging"
+
```
nkryuchkov: Could you please remove the empty line?

pkg/app/appserver/proc.go
```diff
@@ -111,7 +180,10 @@ func (p *Proc) Stop() error {
 
 	// the lock will be acquired as soon as the cmd finishes its work
```
nkryuchkov, line 187: Do I understand correctly that is works like a `WaitGroup`?

pkg/app/client_test.go
```diff
@@ -2,103 +2,19 @@ package app
 
 import (
 	"errors"
-	"os"
 	"testing"
 
+	"github.com/SkycoinProject/skywire-mainnet/pkg/app/appcommon"
```
nkryuchkov: Could you please move it into the last block?

pkg/hypervisor/hypervisor.go
```diff
@@ -13,6 +13,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/SkycoinProject/skywire-mainnet/pkg/app/appcommon"
```
nkryuchkov: Could you please move this line into the last block?

nkryuchkov (approved):
Great job!

pkg/app/appserver/proc.go
evanlinjin: From the looks of it, yes. Although I guess that this way it's safer for reuse (though I'm not sure if we plan to do that).

cmd/apps/skychat/chat.go
evanlinjin: @nkryuchkov As we are displaying logs of apps within visor, I didn't want to end up with something like this:
```
[2020-05-11T03:46:57+12:00] INFO [proc_manager]: [2020-05-11T03:46:57+12:00] DEBUG []: This is an example log.
```
Also, the standard `logrus.TextFormatter` format is easier to parse than in `logging`. This will be helpful as we need to display logs in `hypervisor`.

nkryuchkov: @evanlinjin Thank you for the explanation.

pkg/app/appcommon/proc_config.go
```diff
@@ -0,0 +1,119 @@
...
+// Envs returns the env variables that are passed to the associated proc.
+func (c *ProcConfig) Envs() []string {
+	const format = "%s=%s"
+	return append(c.ProcEnvs, fmt.Sprintf(format, EnvProcConfig, string(c.encodeJSON())))
+}
+
+func (c *ProcConfig) encodeJSON() []byte {
+	b, err := json.Marshal(c)
+	if err != nil {
+		panic(err)
```
Darkren, line 112: What do you think about getting rid of panic and just returning an error here?

pkg/app/appdisc/const.go
```diff
@@ -0,0 +1,7 @@
+package appdisc
+
+// ChangeValue keys.
+const (
+	ConnCountValue     = "conn_count"
```
Darkren: Could you please add comments to these? Not clear what are these consts for

Darkren (commented):
Just a few minor comments, overall - great job! btw, I think we're mixing `app` and `proc` terms here. probably we should consider moving on to just a single one of it as it may get confusing at some point

pkg/app/appcommon/proc_config.go
evanlinjin: I feel like `panic` is suited here because `json.Marshal` will only return error if the `ProcConfig` definition is problematic (which should never happen).

pkg/app/appdisc/const.go
evanlinjin: Will do, thank you.

evanlinjin:
> Just a few minor comments, overall - great job! btw, I think we're mixing `app` and `proc` terms here. probably we should consider moving on to just a single one of it as it may get confusing at some point

As mentioned on [Discord](https://discordapp.com/channels/466450852098605059/552769767904116737/709660342870868050):
> Proc is an instance of an app
> My app2 specs stated to have the ability to run multiple instances of a single app
> But seems we were rushed so I didn't push for that feature
> The current layout edges us closes to that vision (hopefully).

evanlinjin:
https://discordapp.com/channels/466450852098605059/552769767904116737/709679913309044759

PR #371, opened 2020-05-21
Appevent module implementation. 

### LINKED ISSUE #370: Expose an RPC interface from a skywire app that the visor can call.
**Feature description**

There are certain use cases where a visor may wish to communicate with a skywire app in which the communication is initiated by the visor (and not the skywire app): i.e. https://github.com/SkycoinProject/dmsg/issues/93

In the above example, the visor will wish to initiate some sort of communication with the VPN app within the `SessionPreDialCallback` and `SessionDisconnect` callbacks.

* https://github.com/SkycoinProject/dmsg/issues/93

**Possible implementation**
We need to have the `app.Client` host an RPC interface that is only advertised to the visor.
To do this, we can reuse the proc key to be the name in which the RPC object is registered as.

A possible gateway implementation is as follows:
```golang
type RPC struct{ /*TODO*/ }

type Event struct {
    Name string // event name
    Data  []byte
}

type EventResp struct {
    Ok    bool
    Data []byte
}

func (rpc *RPC) OnEvent(e *Event, resp *EventResp) error { /*TODO*/ }
```

In which the app client has a go routine that handles the incoming events.

evanlinjin:
Fixes #370

The appevent module allows for events to be broadcasted from visor (via appserver) to apps.
Events have an event type what determines the contents of the event data.
Apps subscribe to chosen event types, which are broadcasted via event channels (one event channel per event type).

* Introduced a new optional line of communication between app/visor for event broadcast, in which app hosts the listener and the visor (via appserver) does the dial. The app advertises this optional listening address (if it exists) to the visor via the main app <-> visor connection.
* Apps now use the appcommon.Hello JSON object as their initial dial message to appserver (this contains the proc_key, and event broadcast connection related fields).
* app.NewClient() now has an optional appevent.Subscriber as input. If defined, the broadcast event connection will be created and the appevent.Subscriber object will be used for receiving events.
* Updated dmsg version to make use of callbacks (for dialing and closing sessions).

Other changes:
* Updated golangci-lint version for travis.
* Updated helloworld app to listen/write in a loop and to log all events comming from appserver.

## Testing instructions
1. Run visor with `helloworld` app.
2. Restart the dmsg server which the visor is connected to.
3. Inspect visor/app logs.


cmd/apps/helloworld/helloworld.go
```diff
@@ -4,84 +4,167 @@ simple client server app for skywire visor testing
...
+		}
+		log.WithField("n", n).Info("Wrote response message.")
 	}
+}
 
-	buf := make([]byte, 4)
-	if _, err = conn.Read(buf); err != nil {
-		log.Fatalf("Failed to read remote data: %v\n", err)
+func runClient(appC *app.Client) {
+
```
Darkren: Could you please remove this empty line? and one in the for-loop below

pkg/app/appcommon/hello.go
```diff
@@ -0,0 +1,74 @@
...
+	size := binary.BigEndian.Uint16(sizeRaw)
+
+	helloRaw := make([]byte, size)
+	if _, err := io.ReadFull(r, helloRaw); err != nil {
+		return Hello{}, fmt.Errorf("failed to read hello data: %w", err)
+	}
+
+	var hello Hello
+	if err := json.Unmarshal(helloRaw, &hello); err != nil {
+		return Hello{}, fmt.Errorf("failed to unmarshal hellp data: %w", err)
```
Darkren: just a small mistype in `hello`

pkg/app/appevent/broadcaster.go
```diff
@@ -0,0 +1,114 @@
...
+
+	// Notify all clients of event (if client is subscribed to the event type).
+	for client, errCh := range mc.clients {
+		go notifyClient(ctx, e, client, errCh)
+	}
+
+	// Delete inactive clients and reset error channels.
+	for client, errCh := range mc.clients {
+		err := <-errCh
+		close(errCh)
```
Darkren: I'm sorry, why do we need to close and recreate a channel here?

pkg/app/appevent/broadcaster_test.go
```diff
@@ -0,0 +1,135 @@
...
+	// Arrange:
+	// - There is a n(C) number of RPCClients within the Broadcaster.
+	// - All the aforementioned RPCClients are subscribed to all possible event types.
+	// Act:
+	// - Broadcast n(E) number of events using Broadcaster.Broadcast.
+	// Assert:
+	// - Each of the n(C) RPCClients should receive n(E) event objects.
+	// - Received event objects should be in the order of sent.
+	t.Run("broadcast_events", func(t *testing.T) {
+
```
Darkren: Could you please remove this empty line and the one in the test case below?

pkg/app/appevent/subscriber.go
```diff
@@ -0,0 +1,123 @@
...
+		m:        make(map[string]chan *Event),
+		closed:   false,
+	}
+}
+
+// TCPDial subscribes to the TCPDial event channel.
+// This should only be called once.
+func (s *Subscriber) TCPDial(action func(data TCPDialData)) {
+	s.mx.Lock()
+	ch := make(chan *Event, s.chanSize)
```
Darkren: Can't we just check if the event already presents in map? And if it is, simply return. This way we may allow to call this func more than once safely

Darkren: This one is true for all subscribing funcs

pkg/snet/network.go
```diff
@@ -81,16 +82,28 @@ type Network struct {
...
 
 	if conf.Dmsg != nil {
-		c := &dmsg.Config{
+		dmsgConf := &dmsg.Config{
 			MinSessions: conf.Dmsg.SessionsCount,
+			Callbacks: &dmsg.ClientCallbacks{
+				OnSessionDial: func(network, addr string) error {
+					data := appevent.TCPDialData{RemoteNet: network, RemoteAddr: addr}
+					event := appevent.NewEvent(appevent.TCPDial, data)
+					_ = eb.Broadcast(context.Background(), event) //nolint:errcheck
```
Darkren, line 96: Why don't we return an error here? Wondering, cause the func signature allows that

Darkren (changes requested):
Mostly stylish comments, may be considered a matter of taste. Overall good job!

pkg/app/appevent/broadcaster.go
evanlinjin: Good spotting.

pkg/app/appevent/subscriber.go
evanlinjin: @Darkren But what is the usecase for calling subscribing funcs multiple times? Should we end up triggering all the `action` input funcs of all `.TCPDial` call?

pkg/snet/network.go
evanlinjin: @Darkren If an error is returns, the session dial is canceled. This may not be intended behavior. I'll write a comment.

PR #374, opened 2020-05-25
Make proxy disc generic

Darkren:
Yes
Fixes #7 of discovery

 Changes:	
- All endpoints changed according to the changes in the #17 of discovery
- VPN discovery added to the factory


pkg/app/appdisc/factory.go
```diff
@@ -36,32 +36,38 @@ func (f *Factory) setDefaults() {
...
+	// Do not update in proxy discovery if passcode-protected.
+	if containsFlag(conf.ProcArgs, "passcode") {
+		return &emptyUpdater{}, false
+	}
 
+	switch conf.AppName {
+	case skyenv.SkysocksName:
 		return &proxyUpdater{
-			client: proxydisc.NewClient(log, proxydisc.Config{
+			client: servicedisc.NewClient(log, servicedisc.Config{
```
nkryuchkov: What do you think of extracting the config to a variable here? I think this may increase readability. Can it be unified with the same config below?

nkryuchkov (approved):
Great job!

pkg/app/appdisc/factory.go
evanlinjin: I would instead, create a closure named `newServiceDiscClient` or something. Because the `default:` target does not require a `servicedisc.Client`.

pkg/servicedisc/client.go
```diff
@@ -22,31 +22,38 @@ type Config struct {
...
-	entry   Proxy
+	entry   Service
 	entryMx sync.Mutex // only used if UpdateLoop && UpdateStats functions are used.
 	auth    *httpauth.Client
 	client  http.Client
 }
 
 // NewClient creates a new HTTPClient.
-func NewClient(log logrus.FieldLogger, conf Config) *HTTPClient {
+func NewClient(log logrus.FieldLogger, conf Config, sType string) *HTTPClient {
```
evanlinjin: Instead of creating a new input, can we create a new field within the `servicedisc.Config` structure? Thanks!

evanlinjin (changes requested):
Good job! Just some requests for some slight tweaks.

PR #376, opened 2020-05-26
Fix transport manager hangs and various improvements.

### LINKED ISSUE #375: Transport manager hangs
**Describe the bug**
Transport manager hangs when calling `skywire-cli visor add-tp $PK --type stcp`:

`[2020-05-25T22:34:17+03:00] FATAL [skywire-cli]: rpc client timeout`

**Environment information:**

- OS: MacOS

**Steps to Reproduce**

Setup two visors. 
Call `skywire-cli visor add-tp $PK --type stcp`

**Additional context**
The issue seems to be related to the transport manager screwing up.

evanlinjin:
Fixes #375

 Changes:	
* Restructured (transport.Manager).SaveTransport to improve readability and ensure we always close the managed transport before calling mTp.wg.Wait(). In fact, we are now using mTp.Close() directly (which does both).
* Visor's RPC client timeout is now optional. Setting a value of 0 duration will result in no timeout.
* Disabled visor RPC client timeout for visor CLI.

Additional Changes:
* Wait for dmsg network during visor startup.

How to test this PR:
* Start a `skywire-visor`.
* Add transports of different types via `skywire-cli`.
* Nothing should hang indefinitely.


pkg/transport/manager.go
```diff
@@ -216,45 +216,43 @@ func (tm *Manager) SaveTransport(ctx context.Context, remote cipher.PubKey, tpTy
...
-				return nil, err
+	if err = mTp.Dial(ctx); err != nil {
+		// This occurs when an old tp is returned by 'tm.saveTransport', meaning a tp of the same transport ID was just
+		// deleted (and has not yet fully closed). Hence, we should close and delete the old tp and try again.
+		if err == ErrNotServing {
+			if closeErr := mTp.Close(); closeErr != nil {
+				tm.Logger.WithError(err).Warn("Closing mTp returns non-nil error.")
 			}
+			delete(tm.tps, mTp.Entry.ID)
+			goto TrySaveTp
```
Darkren: Could you please move to the for-loop instead of `goto`? All my nature just stands against it, even if it's mostly the same as the loop in such case

Darkren (changes requested):
Could you please do just a minor change? Otherwise, really great job!

cmd/skywire-cli/commands/visor/root.go
```diff
@@ -27,17 +25,11 @@ var RootCmd = &cobra.Command{
 }
 
 func rpcClient() visor.RPCClient {
+	const rpcDialTimeout = time.Second * 5
```
nkryuchkov, line 28: Would it be enough for everything?

nkryuchkov (approved):
Except for the @Darkren 's note, the code looks good to me. But not sure about the timeout. Great job!

cmd/skywire-cli/commands/visor/root.go
evanlinjin: This is only for dialing. A.k.a establishing a TCP connection with the visor (in this case).

nkryuchkov: @evanlinjin Thank you for the explanation. It's fine then.

PR #381, opened 2020-05-31
Improve reliability and logging of setup node.

### LINKED ISSUE #379: Troubleshoot setup node.
**Describe the bug**
Sometimes the setup node hangs in production indefinitely. We need to troubleshoot this behavior.

evanlinjin:
Fixes #379

 Changes:	
* Split setup node into multiple functions for easier testability.
* Check inputs for each setup request.
* Use context.Context in all parts of the setup process to ensure we cancel on hard timeout.
* Introduced routerclient.Map to clean up various parts of the logic, as well as allow the setup process to be speedier.
* Got rid of the routing.Path data type as it is confusing. I has resorted to use []routing.Hop directly.
* Unit tests.

How to test this PR:
* `make check`
- [x] I need to update `skywire-services` interactive env to work with these changes.


.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: go
 go:
   # - "1.13.x" At minimum the code should run make check on the latest two go versions in the default linux environment provided by Travis.
-  - "1.13.x"
+  - "1.14.x"
```
nkryuchkov, line 4: Could you please update the comment?

pkg/setup/node_old_test.go
```diff
@@ -0,0 +1,248 @@
+// +build !no_ci
+
+package setup
+
+// TODO(evanlinjin): Either fix or rewrite these tests.
```
nkryuchkov: I think it would be better if these tests are fixed/rewritten before merging.

pkg/setup/node_test.go
```diff
@@ -1,276 +1,147 @@
...
-		Desc:    forwardRoute.Desc,
-		Forward: forwardRules[route.Desc.DstPK()],
-		Reverse: consumeRules[route.Desc.DstPK()],
+	for i, tc := range testCases {
+		t.Run(strconv.Itoa(i), func(t *testing.T) {
+			// arrange
+			rtIDR := newMockReserver(t, nil)
+
+			// act
+			fwd, rev, inter, err1 := GenerateRules(rtIDR, []routing.Route{tc.fwd, tc.rev})
```
nkryuchkov: Can `err1` be renamed to `err`?

nkryuchkov (changes requested):
Great improvements!

pkg/setup/id_reserver_test.go
```diff
@@ -80,10 +80,10 @@ func TestIdReserver_ReserveIDs(t *testing.T) {
 	timeout := time.Second
 
 	type testCase struct {
-		testName string                                  // test name
-		routers  map[cipher.PubKey]*mockGatewayForDialer // arrange: map of mock router gateways
-		paths    [][]routing.Hop                         // arrange: idReserver input
-		expErr   error                                   // assert: expected error
+		testName string                        // test name
+		routers  map[cipher.PubKey]interface{} // arrange: map of mock router gateways
```
nkryuchkov, line 83: Why would `*mockGatewayForDialer` not be better here?

pkg/setup/node_test.go
```diff
@@ -17,6 +20,215 @@ import (
...
+// for tests, we make transport IDs deterministic
+// hence, we can derive the tpID from any pk pair
+func determineTpID(pk1, pk2 cipher.PubKey) (tpID uuid.UUID) {
+	v1, v2 := pk1.Big(), pk2.Big()
+	var hash cipher.SHA256
+	if v1.Cmp(v2) > 0 {
+		hash = cipher.SumSHA256(append(pk1[:], pk2[:]...))
+	} else {
+		hash = cipher.SumSHA256(append(pk2[:], pk1[:]...))
+	}
```
nkryuchkov, line 181: Could you please add empty lines before and after `if` block? I think it may improve readability a bit.

pkg/setup/node_test.go
```diff
@@ -17,6 +20,215 @@ import (
...
+	return len(m)
+}
+
+func biRouteFromKeys(fwdPKs, revPKs []cipher.PubKey, srcPort, dstPort routing.Port) routing.BidirectionalRoute {
+	fwdHops := make([]routing.Hop, len(fwdPKs)-1)
+	for i, srcPK := range fwdPKs[:len(fwdPKs)-1] {
+		dstPK := fwdPKs[i+1]
+		fwdHops[i] = routing.Hop{TpID: determineTpID(srcPK, dstPK), From: srcPK, To: dstPK}
+	}
```
nkryuchkov, line 153: Could you please add a new line after `for` statement?

pkg/routing/route.go
```diff
@@ -76,6 +76,21 @@ func (br *BidirectionalRoute) Check() error {
...
 }
 
+// String implements fmt.Stringer
+func (br *BidirectionalRoute) String() string {
+	m := map[string]interface{}{
+		"descriptor": br.Desc.String(),
+		"keep_alive": br.KeepAlive.String(),
+		"fwd_hops":   br.Forward,
+		"rev_hops":   br.Reverse,
+	}
```
nkryuchkov, line 92: Would you mind adding an empty line after this line and before `return` statement?

nkryuchkov (changes requested):
Great job! Just a few minor requests.

pkg/setup/id_reserver_test.go
evanlinjin: @nkryuchkov because we may have different implementations for the RPC Gateway for future tests.

In fact, as I was writing these tests, there was a time point when I had multiple implementations, but I deemed them not necessary.

I think we should leave it as it, just in case for possible future changes.

nkryuchkov: @evanlinjin What do you think about creating a common interface for RPC Gateway implementations and use it in this `map`?

nkryuchkov (approved):

PR #392, opened 2020-06-05
Adding trusted visors from config on startup

### LINKED ISSUE #368: Use Trusted Visor functionality
**Feature description**

We currently have the trusted visor field in Skywire config. Visors add transports to PubKeys listed there automatically. This should be used to establish transports to a list of publicly deployed visors. Visors will automatically setup (5-10) transports. That makes it easier to establish app routes via visors for users without having to establish transports manually. The trusted visors act as a relay.

If a user wants to establish a route (i.e. VPN client to VPN server), he needs to first establish a transport to the remote. With this feature, users could setup routes to most remotes without manually setting up transports. 

We have a list of trusted visors that visors connect to by default. We need to change the transport type used for trusted visors to `stcp_address_resolution` instead of `dmsg`. 

**Additional context**
We need to deploy a few visors publically.

nkryuchkov:

Fixes #368

 Changes:	
- Add trusted visors from config on startup

How to test this PR:
- Set branch of `skywire-services` to https://github.com/SkycoinPro/skywire-services/pull/184
- Run generic integration env
- Wait 5 seconds
- Run `./bin/skywire-cli --rpc $RPC_A visor ls-tp`
- Run `./bin/skywire-cli --rpc $RPC_C visor ls-tp`
- Run `./bin/skywire-cli --rpc $RPC_B visor ls-tp`
- There should be 1 transport on `$RPC_A` and `$RPC_C`, 2 transports on `$RPC_B`; all transports should be up


pkg/transport/manager.go
```diff
@@ -136,6 +143,26 @@ func (tm *Manager) serve(ctx context.Context) {
 	}
 }
 
+// AddTrustedVisors adds transports to trusted visors from config.
+func (tm *Manager) AddTrustedVisors(ctx context.Context) {
```
evanlinjin: I feel like this does not add enough functionality to be a member here. Instead, I would keep this in the `/pkg/visor/init.go` file and just call `tm.SaveTransport` in a loop.

pkg/visor/visor.go
```diff
@@ -161,6 +162,12 @@ func NewVisor(conf *visorconfig.V1, restartCtx *restart.Context) (v *Visor, ok b
 		return v, ok
 	}
 
+	ctx := context.Background()
```
evanlinjin: As aforementioned, I would keep this in the `init.go` file.

evanlinjin (changes requested):
Let me know your thoughts on my thoughts.

pkg/transport/manager.go
nkryuchkov: @evanlinjin I was going to put it into the initialization step. However, I decided that trusted visors should be added only after all initialization steps finish with success. If I keep this in `NewVisor`, the function would be too large. 

Let me know if changed your mind. If not, I will move it into the transport initialization step.

evanlinjin: > If I keep this in NewVisor, the function would be too large.

@nkryuchkov A separate function will be fine. I would, however, put this function in the `visor` package.

> However, I decided that trusted visors should be added only after all initialization steps finish with success.

@nkryuchkov How about we create another `init` function? Maybe `initConnectTrustedVisors` or something, and append it to the end of `initStack()`.

nkryuchkov: @evanlinjin That sounds like a good idea.

Darkren (approved):
Great job!

PR #397, opened 2020-06-08
Implement a UDP transport

### LINKED ISSUE #391: Implement reliable UDP transport
**Feature description**

We should implement a reliable UDP transport that can later be used for UDP holepunching. A library that seems suitable is [https://github.com/xtaci/kcp-go](https://github.com/xtaci/kcp-go)

nkryuchkov:

Fixes #391

 Changes:	
- Implement a UDP transport

How to test this PR:
- Checkout `skywire-services` against `feature/sudp`
- Run integration build
- Run the generic integration environment
- Run `make integration-startup-sudp`
- Run `curl --data {'"recipient":"'$PK_A'", "message":"Hello"}' -X POST $CHAT_C`
- Make sure that message appeared in `VisorA` logs


pkg/snet/network.go
```diff
@@ -243,13 +267,23 @@ func (n *Network) Init() error {
...
+	}
+
 	return nil
 }
 
 // Close closes underlying connections.
 func (n *Network) Close() error {
 	wg := new(sync.WaitGroup)
-	wg.Add(4)
+	wg.Add(5)
```
Darkren: Feels like we should try to define some kind of interface here. not neccessary to be done within this PR, but this part may become a pandora's box some day

Darkren: I mean for the networks. so we could have like a slice of these and then just call `wg.Add(len(networks))`

pkg/snet/sudp/handshake.go
```diff
@@ -0,0 +1,242 @@
...
+
+// middleware to add deadline and HandshakeError to handshakes
+func handshakeMiddleware(origin Handshake) Handshake {
+	return func(conn net.Conn, deadline time.Time) (lAddr, rAddr dmsg.Addr, err error) {
+		if err = conn.SetDeadline(deadline); err != nil {
+			return
+		}
+
+		if lAddr, rAddr, err = origin(conn, deadline); err != nil {
+			err = HandshakeError(err.Error())
```
Darkren, line 52: shouldn't we return here?

pkg/snet/sudp/handshake.go
```diff
@@ -0,0 +1,242 @@
...
+	sig := f2.Sig
+	f2.Sig = cipher.Sig{}
+
+	var b bytes.Buffer
+	if err := json.NewEncoder(&b).Encode(f2); err != nil {
+		return err
+	}
+
+	hash := cipher.SumSHA256(b.Bytes())
+	rPK, err := cipher2.PubKeyFromSig(cipher2.Sig(sig), cipher2.SHA256(hash))
```
Darkren, line 194: shouldn't we handle this error?

Darkren (commented):
Just a couple of minor questions. Otherwise - great job!

pkg/snet/network.go
nkryuchkov: @Darkren That's a good idea. I thought about it but forgot to implement it.

pkg/snet/sudp/handshake.go
nkryuchkov: @Darkren I guess we should. I copied it from `stcp` with no changes. @evanlinjin What do you think about it?

pkg/snet/sudp/handshake.go
nkryuchkov: Same here, just copied. I'm not sure. @evanlinjin What do you think?

PR #400, opened 2020-06-10
Implementation of UDP hole punch transport

### LINKED ISSUE #394: Implement UDP holepunch transport
**Feature description**

We need to implement a UDP holepunch transport. This ticket depends on #391 
UDP holepunch transports should be more versatile than TCP holepunching.

nkryuchkov:

Fixes #394

 Changes:	
- Implementation of UDP hole punch transport
- Partial refactoring of direct transports
- For transports using `address-resolver`, if visors use the same IP, they try to establish a connection via local network first.

How to test this PR:
- Start generic integration env
- Run `make integration-startup-sudph`


pkg/snet/arclient/client.go
```diff
@@ -1,4 +1,5 @@
 // Package arclient implements address resolver client
+// TODO(nkryuchkov): move to internal
```
Darkren, line 2: do you plan to move it in a separate PR?

pkg/snet/arclient/client.go
```diff
@@ -325,11 +410,121 @@ func (c *httpClient) WS(ctx context.Context, dialCh <-chan cipher.PubKey) (<-cha
...
+
+	c.sudphAddrCh = addrCh
+
+	return nil
+}
+
+func (c *client) Close() error {
+	defer func() {
+		c.stcphConn = nil
+		// TODO(nkryuchkov): uncomment
```
Darkren, line 496: Was it forgotten? As well as closing of sudph conn below

pkg/snet/directtransport/client.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	"context"
+
+	"github.com/SkycoinProject/dmsg/cipher"
+	"github.com/SkycoinProject/skycoin/src/util/logging"
+)
+
+// Client is a direct transport client.
+type Client interface { // TODO(nkryuchkov): use
```
Darkren: Should these todos be removed?

pkg/snet/network.go
```diff
@@ -605,6 +787,22 @@ func (n *Network) Listen(network string, port uint16) (*Listener, error) {
 			return nil, err
 		}
 
+		return makeListener(lis, network), nil
+	case sudpr.Type:
+		sudprC := n.clients.SudprC()
```
Darkren, line 727: Can't client be `nil` here, like with stcph above?

pkg/snet/sudph/client.go
```diff
@@ -0,0 +1,428 @@
...
+
+		c.mx.Lock()
+		defer c.mx.Unlock()
+
+		if err := c.addressResolver.Close(); err != nil {
+			c.log.WithError(err).Warnf("Failed to close address resolver client")
+		}
+
+		for _, lis := range c.lMap {
+			_ = lis.Close() // nolint:errcheck
```
Darkren, line 409: Why won't we log errors?

pkg/snet/sudpr/client.go
```diff
@@ -0,0 +1,266 @@
...
+	log *logging.Logger
+
+	lPK             cipher.PubKey
+	lSK             cipher.SecKey
+	p               *porter.Porter
+	addressResolver arclient.APIClient
+	localAddr       string
+
+	lUDP net.Listener
+	lMap map[uint16]*directtransport.Listener // key: lPort
```
Darkren, line 36: Might be a matter of taste, but what do you think about calling this `listeners`? might be longer but not that much. it just feels to me like calling map a map is too explicit lol

Darkren (commented):
Few minor questions, some may be of a taste. But anyway, a huge job is done, great!

pkg/snet/arclient/client.go
nkryuchkov: @Darkren Yes, I'm going to do that in a PR for https://github.com/SkycoinProject/skywire-mainnet/issues/409

pkg/snet/arclient/client.go
nkryuchkov: @Darkren No, I left it for https://github.com/SkycoinProject/skywire-mainnet/issues/409, because I'm going to combine all direct transports in one and I also need to add redialing logic.

pkg/snet/directtransport/client.go
nkryuchkov: @Darkren I started to use this interface in a branch that fixes #409 and removed the `TODO` there.

pkg/snet/network.go
nkryuchkov: @Darkren Well spotted, thanks.

pkg/snet/sudph/client.go
nkryuchkov: I copied this code from `stcp` and did no changes to it because I was going to combine all transports in one. I agree that we should have an error check here, but would you mind if this would be done in a PR for #409? This code won't be duplicated 6 times there.

pkg/snet/sudpr/client.go
nkryuchkov: @Darkren Good idea. To not rename this in 6 clients, do you mind if I do that in #409?

pkg/snet/sudph/client.go
Darkren: Of course, do as comfortable

pkg/snet/sudpr/client.go
Darkren: Sure

Darkren (approved):

PR #402, opened 2020-06-11
Allow visor to start in case of address resolver or tp disc failure

### LINKED ISSUE #401: Visor fails to start if connection to address resolver cannot be established.
**Describe the bug**

Whether visor starts all not should not depend on whether address resolver is up or down.

Ideally, visor should still start, but it logs the connection error. Afterwards, it attempts to connect to the address resolver via exponential backoff.

If the aforementioned is difficult to implement, maybe another temporary solution would be to block until the connection to the address resolver can be established.

Darkren:
Yes
Fixes #401 

 Changes:	
- Visor won't start before it connects to TP disc. it keeps dialing TP disc till sucess
- Added callback to notify outer-network code when new network gets established
- Fixed transport manager to handle background-initiated networks
- Fixed already-known transports initiation 

Tested partially, except for the transport manager initialization and already-known transports. We need to have `stcpr/stcph` transports working to test this. Also code may look clumsy in some places, but this should get sorted out as soon as we deal with the refactoring of network part which is proposed by @nkryuchkov


pkg/snet/network.go
```diff
@@ -163,32 +204,81 @@ func New(conf Config, eb *appevent.Broadcaster) (*Network, error) {
...
 
-	if conf.NetworkConfigs.STCPR != nil {
-		ar, err := arclient.NewHTTP(conf.NetworkConfigs.STCPR.AddressResolver, conf.PubKey, conf.SecKey)
-		if err != nil {
-			log.WithError(err).Warnf("Failed to connect to address-resolver, STCPR will be disabled")
-		} else {
-			addressResolver = ar
+	if conf.NetworkConfigs.STCPR != nil || conf.NetworkConfigs.STCPH != nil {
+		var (
+			addressResolver   arclient.APIClient
```
evanlinjin, line 179: When I first made the `snet` library, I intended it to be just a wrapper for all the network types. I think it would be cleaner if we didn't have any reconnecting logic here at all. Would we be able to implement the reconnect logic within `arclient.httpClient`?

pkg/snet/network.go
```diff
@@ -114,26 +131,50 @@ type NetworkConfigs struct {
...
 }
 
 // Network represents a network between nodes in Skywire.
 type Network struct {
-	conf     Config
-	networks []string // networks to be used with transports
-	clients  NetworkClients
+	conf       Config
+	networksMu sync.RWMutex
+	networks   map[string]struct{} // networks to be used with transports
```
evanlinjin: Would we be able to rename these variable names to `netsMu` and `nets`. I think they are getting too long and it's affecting the readability of the code.

pkg/snet/network.go
```diff
@@ -114,26 +131,50 @@ type NetworkConfigs struct {
 
 // NetworkClients represents all network clients.
 type NetworkClients struct {
```
evanlinjin: Would we be able to move this to another file? It's getting a little cluttered.

pkg/transport/manager.go
```diff
@@ -74,49 +75,69 @@ func (tm *Manager) Serve(ctx context.Context) {
...
 			}
-		}()
+		}
+	}()
+}
+
+func (tm *Manager) serve(ctx context.Context) {
+	// some networks may not be ready yet, so we're setting a callback first
+	tm.n.OnNewNetworkType(func(netType string) {
+		tm.serveNetwork(ctx, netType)
```
evanlinjin, line 134: If we make arclient be responsible for redialing, we don't need this callback logic here.

Instead, we should make `tm.serveNetwork` return an error when the `n.Listen` call fails, and wrap this with some kind of redialing logic.

pkg/transport/manager.go
```diff
@@ -153,9 +176,42 @@ func (tm *Manager) initTransports(ctx context.Context) {
...
+					}
+
+					return nil
+				})
+				if err != nil {
+					tm.Logger.WithError(err).Warnf("INIT: failed to init tp: type(%s) remote(%s) tpID(%s)", tpType, remote, tpID)
+				}
+			}(entry)
+		} else {
+			tm.Logger.Debugf("Successfully initialized TP %v", *entry.Entry)
```
evanlinjin, line 180: If this entire section of code needed? I don't see a problem with the `.Dial` calls returning an error within `managedTransport`.

evanlinjin (changes requested):
Probably needs further discussion.

pkg/transport/manager.go
Darkren: @evanlinjin what if we need to call `Accept`, not `Dial`? we need to know about the transport first

Darkren: at least, I think that this transport initialization is made for that. tbh can't be sure

pkg/transport/manager.go
Darkren: @evanlinjin well, 2 things here. I totally forgot about the error being returned there, that's fair. but I'd stick to the callback just for one reason. if we put redialing logic here, we'll be having AR rediailing and redialing-per-failed-network-type here. which may make start of a visor painful for small machines, for RPi, for example

pkg/transport/manager.go
evanlinjin: Yes, calls to `.Listen` would fail, but I think that is okay.

Darkren: @evanlinjin I'm sorry I'm a bit lost. I'm starting to wonder why we have this TP initialization logic at all lol. but wouldn't it be crucial to have this retrying logic for intermediary visors? these ones may not have apps running, so they don't really known whether they should call `Dial` or `Accept`, they may only rely on the discovery info. again, correct me if I'm wrong please. I might be getting more and more confused here

evanlinjin: `saveTransport` adds a `managedTransport` to the `transport.Manager`. A `managedTransport` reconnects itself on failure, so reconnection logic is not needed.

Also, I do not think we need all networks to be working to save transports associated with them. Transports have `UP` and `DOWN` statuses. Let's say `arclient` is not working, then all transports that require `arclient` to function would be `DOWN`.

Darkren: you have your point, probably this would be enough. I'll deal with this part. also please take a look at the query just a bit above @evanlinjin

evanlinjin: I'm thinking about the above query right now... it's tough haha

pkg/transport/manager.go
evanlinjin: @Darkren I think it is either we completely use this callback, or do the loop through `n.TransportNetworks()`. Having both is kind of confusing in my opinion.

Darkren: @evanlinjin that's fair. I don't really like callbacks that much too. but choosing between callback and beating up CPU I prefered the first one. well, we have a lot of different network types rn. @nkryuchkov was going to generalize code there a bit. probably an interface for network would be good with method like `Ready` which exposes channel. this way we would be able to setup just goroutines for networks like `stcpr/stcph`, wait for a signal on the channel and then start servingn network. that would be much cleaner I think. but this is not really possible rn before generalization is done

evanlinjin: Yes, it will be good to have something like `Ready() <-chan struct{}` as a standard across all transport types. For the meantime, could you make a note?

I think making this standard is priority, otherwise the codebase will just get messier.

Darkren: @evanlinjin sure, will add `TODO` there

pkg/snet/network.go
```diff
@@ -9,6 +9,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/SkycoinProject/dmsg/netutil"
+
```
nkryuchkov: Imports are not sorted

pkg/snet/network.go
```diff
@@ -399,14 +552,24 @@ func (n *Network) Dial(ctx context.Context, network string, pk cipher.PubKey, po
...
 		if err != nil {
 			return nil, fmt.Errorf("stcpr client: %w", err)
 		}
 
 		return makeConn(conn, network), nil
 	case stcph.Type:
-		conn, err := n.clients.StcphC.Dial(ctx, pk, port)
+		stcphC := n.clients.StcphC()
+		if stcphC == nil {
+			return nil, errors.New("stcph client is not ready")
```
nkryuchkov, line 540: What do you think of making this error and the error below package-level?

pkg/transport/manager.go
```diff
@@ -9,6 +9,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/SkycoinProject/dmsg/netutil"
+
```
nkryuchkov: Not sorted.

pkg/snet/network.go
nkryuchkov: @evanlinjin @Darkren Would you mind if I work on the reconnection logic of `arclient.httpClient`? Its code is familiar to me and I have some ideas for reconnection logic. Moreover, I'm going to add a UDP connection in https://github.com/SkycoinProject/skywire-mainnet/pull/400. In my opinion, it would make more sense to implement reconnection logic after UDP implementation is finished.

Darkren: @nkryuchkov I'll leave a `TODO` comment then

nkryuchkov (approved):

jdknives (approved):
Works fine. Can be merged if @evanlinjin is happy to fix the rest in another issue tackled by nikita.

evanlinjin (approved):
Good job! Looks like it is ready to merge.

PR #412, opened 2020-06-18
Setup node metrics.

### LINKED ISSUE #405: Metrics for setup-node.
We need [Prometheus](https://prometheus.io/) metrics reporting for `setup-node`.

### Metrics
- [x] Gauge for ongoing request count.
- [x] Histogram for request duration.
- [x] Counters for failed and successful requests.

evanlinjin:
Fixes #405 

Changes:	
- Re-implemented how metrics are handled.
- Updated .gitignore

How to test this PR:
- Start `setup-node` with flag `-m :2121`
- Navigate to `http://localhost:2121/metrics`


jdknives:
@evanlinjin binaries are still there.

cmd/setup-node/commands/root.go
```diff
@@ -60,52 +70,62 @@ var rootCmd = &cobra.Command{
...
+	go func() { log.Fatal(http.ListenAndServe(metricsAddr, mux)) }()
+
+	return m
 }
 
 // Execute executes root CLI command.
 func Execute() {
 	if err := rootCmd.Execute(); err != nil {
-		log.Fatal(err)
+		panic(err)
```
Darkren, line 129: Why do you think this change is necessary?

pkg/setup/setupmetrics/metrics.go
```diff
@@ -0,0 +1,65 @@
...
+		m.reqDurations,
+	}
+}
+
+func (m *metrics) RecordRequest(_ routing.BidirectionalRoute) func(rules *routing.EdgeRules, err *error) {
+	start := time.Now()
+	m.activeRequests.Inc()
+
+	return func(rules *routing.EdgeRules, err *error) {
+		var successStr string
```
Darkren: we may simplify this part a bit with something like this:
```
successStr := "true"
if *err != nil {
    successStr = "false"
}
```

btw, is it neccessary to have interface pointer in the func signature, wouldn't just `error` be enough here?

Darkren (changes requested):
Few minor questions, but otherwise - great job! Also, could you please remove transport logs?

cmd/setup-node/commands/root.go
evanlinjin: `log.Fatal` logs the error and then panics. However, when a `cobra.Command` fails, it already logs the error. So we only need to panic here.

pkg/setup/setupmetrics/metrics.go
evanlinjin: That code looks cleaner. I'll get it done.

The pointer of an `error` is needed. We need to get the pointer of the error, as the error value will change.

PR #417, opened 2020-06-23
packaging improvements

0pcom:
no
### Fixes
* autoconfiguration of visor and hypervisor upon package installation

### Changes:	
* generates ssl key and cert and adds them to the package at the location they are found at in skybian following the procedure detailed [here](https://github.com/SkycoinProject/skywire-mainnet/wiki/Skywire-Mainnet-Installation-From-Source#enable-tls)
* adds `postinst` script which is automatically executed upon package installation to accomplish configuration of hypervisor and visor.
* adds package-based method of setting / distributing the hypervisor public key
* adds local apt repository configuration for distributing the hypervisorconfig package
* hosts the apt repo on the local network with `readonlycache` a simple golang webserver

### Notes:
* the hypervisor configuration does not take place if the `hypervisorconfig` package is installed
* the `hypervsorconfig` package is created in the `postinst` script when the hypervisor is configured, but it is not installed on the hypervisor. It is added to the hypervisor's apt repo in /var/cache/apt/repo and made available on the LAN to visors
* the configuration of the apt repo hosted by the hypervisor in the visor's /etc/apt/sources.list **is not included in these scripts.**

This should be done either manually by the user in a ssh session or via a functionality which can be added to skyimager. Example:
```
add-apt-repository 'deb [trusted=yes] http://192.168.0.2 sid main'
```
Substituiting in the above command the debian codename which is listed in `/etc/os-release`

### Issues:
* the visor fails to start when the hypervisor key is set in /etc/skywire-visor.json
```
failed to parse config. error="json: cannot unmarshal object in go struct field V1.hypervisors of type cypher.Pubkey"
```
This does not happen with the master branch, only with develop.

### Conflicts:
The extent of the configuration which is accomplished by the postinst script __may cause difficulty__ when it is installed in a chroot in skybian. 

### How to test this PR:
```
make package-amd64
dpkg -i *.deb
```
go to 127.0.0.1:8000

hypervisor is running, visor __would be running__ if not for the aforementioned errors.


0pcom:
Moses Narrow is me

PR #434, opened 2020-07-08
Automate trusted visor functionality

### LINKED ISSUE #431: Automate trusted visor functionality
**Feature description**

Automate trusted visor functionality

We can register the trusted visors in the service discovery and every user that sets up a visor on public IP can register there.

**Additional context**
We need to have checks of some sort to prevent someone from accidentally registering their visor in service discovery if they are not on public IP. This may be done by sending an HTTP request from service discovery to visor to check availability.

nkryuchkov:

Fixes #431

 Changes:	
- Register trusted visor in service discovery

How to test this PR:
- Append `"public_trusted_visor": true` to visor's config that is going to be tested
- Run integration env
- If the visor is available from the internet, `Sent request to register visor as public trusted` followed by no errors would appear in logs; otherwise, the error `Unable to register visor as public trusted as it's unreachable from WAN` would appear.
- If the visor is available from the internet, it should appear here: http://service.discovery.skywire.cc/api/services?type=visor

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,6 +20,7 @@ require (
 	github.com/mmcloughlin/avo v0.0.0-20200523190732-4439b6b2c061 // indirect
 	github.com/pkg/profile v1.5.0
 	github.com/prometheus/client_golang v1.7.1
+	github.com/prometheus/common v0.10.0 // indirect
 	github.com/rakyll/statik v0.1.7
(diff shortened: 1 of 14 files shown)


Darkren (approved):
Great job!

jdknives (approved):

Issue #430, opened 2020-07-05
Remove the unused code from the manager


Senyoret1:
**Feature description**
The Skywire manager has a lot of unused code that was left from the previous version and was not deleted just in case something was needed, as a lot of changes were still being made. However, as the manager is now more mature, the code should be deleted. If something is needed, the code is still in the testnet repo.

**Is your feature request related to a problem? Please describe.**
The unused code makes the UI code much more difficult to underestand and maintain.

**Describe the solution you'd like**
All unused components should be removed, all related translation strings deleted and the styles removed. After that, the style files can be restructured in a simpler way.

PR #436, opened 2020-07-11
Improvements for the UI/UX of the manager

### LINKED ISSUE #429: Add UI/UX improvements to the manager
**Feature description**

Describe the feature
This is a list of improvements that can be done to improve the UI and UX of the Skywire manager:

- [x] If the user has too much data of a single type (visors, transports, routes or apps), the manager UI should show a search option, for being able to find the desired element easily.
- [ ] The manager should show more data in the routes list, as it is currently only showing the rule, which is not a friendly data for the user.
- [x] Show a paginator in the visor list if there are too many visors, just like it happens with the transport list.
- [x] The tab changes should be saved in the navigation history, to be able to return to the previous tab with the back button, as the current behavior is a bit confusing.
- [x] A method for reusing the node data should be added to the node details page, to avoid having to reload all the data each time the active tab is changed. This would improve the navigation.
- [x] Add a button for removing all offline nodes, as the manager keeps them in the list waiting for them to be online again.
- [x] Create a procedure for updating the network traffic graphs in constant intervals, as the current method refresh the graphs each time the page data is refresed, which is very inconsistent.
- [x] Add help icons with information about the data shown by the manager in different parts. This would allow new users to quickly underestand concepts like "transports" and "routes", which are specific for Skywire and may be confusing for new users.

Senyoret1:

Fixes #429

 Changes:	
- The procedure for updating the node list and node data was changed, so that now the data is managed by a centraliced service. This avoids the app always having to update the data after each navigation, which makes navigating some parts of the UI much faster.
- Now the code does not check the state of the login before each navigation, but just in relevant parts, which makes the navigation much faster in all parts of the app.
- Now the node list (with and without dmsg info) has pagination.
- An option for removing all offline visors was added to the visor list.
- Tab changes are now saved in the navigation history.
- The delay for retrying to get the data after an error was centralized in a configuration variable.
- Some other small improvements and fixes.

How to test this PR:
The first 2 points can be checked by navigating between the routing and apps tabs in the visor details page. With the old version it had to reload the data in each tab, but now the data should continue being available just after navigating.

For checking the pagination in the node list, the hypervisor must return a list of more than 40 connected visors.

The option for removing all offline visors can be found by pressing the new button at the top-right part of the visor list.

For testing the tab changes, simply navigate to a new tab and use the browser's back button for returning to the previous tab.


Senyoret1:
The last commits include the following changes:
- The day in which the manager updates the traffic data is nowmuch more predictable and less chaotic.
- Now the manager has generic code for filtering the lists and the UI allows the user to set custom serarch terms to filter all the lists.
- Some improvements were added to the way in which the data refreshed.
- Several other small fixes and improvements.

Senyoret1:
The last commits include the following changes:
- The node labeling code of the manager UI was coverted into a generic ID labeling code.
- Now it is possible to add labels to local nodes, remote nodes, dmsg servers and transport IDs. When done, the UI shows the labels along the IDs, to make it easier for the user to identify the elements.
- The manager UI allows to filter using the new labels.

Senyoret1:
The previous commits where cherry picked and force push was used to make the PR work well with the changes made to the develop branch.

Senyoret1:
- A problem with the color of disabled buttons was solved.
- Several changes were made to how the manager displays the labels and uses them for filtering, to give a more predictable UX.

@jdknives You can check and merge this PR anytime if you want. There are still some changes pending, but the PR works as it is. I will continue working using the commits of this PR as a basis, but If needed I can create another PR.

Senyoret1:
The last commits include the following changes:
- All old unused components were removed from the repository.
- The translatable strings and styles that ended unused after the deletion were also removed.
- All unused unit test files (with initialization code but no real tests) were removed.
- Some folders with code related to the apps were moved, to have them in locations consistent with the structure of the rest of the project.

 This fixes #430

Senyoret1:
The last commits include the following changes:
- Now various functionalities of the lists shown by the manager are in centralized helper classes, to avoid code repetifion, improve the separation of concerns and make the code much more maintanable.
- The manager now has options for managing the labels.
- The manager was updated to Angular 9.
- Several other small improvements and fixes.

jdknives (approved):
Tested all the new features and they worked well. We can merge this to not let it get too big a PR.

PR #441, opened 2020-07-15
Change uptime update interval

Darkren:
Yes

 Changes:	
- Uptime update interval is set to 10 seconds


jdknives:
Wrong target branch. @Darkren

PR #443, opened 2020-07-16
WebSocket API for updater

### LINKED ISSUE #413: Better API for updating the visors
**Feature description**
The hypervisor should have an API endpoint just for starting to update a visor and maybe another for checking the progress.

Currently the hipervisor API has an endpoint for updating a visor, which only returns a response after finishing the process. The problem with the current endpoint is that the update procedure may take too long and that may cause the hypervisor API to return an error because the timeout. This causes problems for the UI, as the current bahavior makes the UI to keep the user waiting a long time just to end showing an error.

It would be better just to have an endpoint for starting the process, returning only if the update procedure was started. That would allow the UI to tell the user in a short time that procedure has been started and that it could take a while. Additionally, an endpoint for checking the progress would allow to show the user how the process is going. However, as the Hypervisor API does not have events, if this last endpoint is going to be implemented it should be possible to check it frecuently.

Not sure about this, but maybe the update can be downloaded by the hypervisor and not the visor itself.

nkryuchkov:

Fixes #413 

Changes:	
- Add WebSocket API for updating visor
- Add WebSocket API for updating hypervisor
- Add draft support of WebSocket API to the frontend for testing and debugging purposes, the code likely needs to be modified or rewritten (/cc @Senyoret1) 
- Send update progress to frontend in WebSocket API
- If called multiple times, updater WebSocket endpoints return current progress instead of starting a new updating process
- Fix `.goreleaser.yml`

How to test this PR:
1. Change `VERSION` variable in `Makefile` to `v0.1.0`
2. Change `channel: 'stable'` to `channel: 'testing'` at `./static/skywire-manager-src/src/app/services/node.service.ts#213`
3. Run `make build-ui`
4. Run `make integration-build`
5. Run `make integration-run-generic`
6. Open hypervisor web UI
7. Update a visor
8. The visor should become updated with no errors
9. To test hypervisor updating, change `visors/${nodeKey}/update/ws` to `update/ws` at `./static/skywire-manager-src/src/app/services/node.service.ts#216` and follow steps 3-8, instead of visor, hypervisor will be updated
10. To test returning current progress on multiple WebSocket endpoint calls, start an update, refresh the browser page and run the update again. Instead of starting a new updating, the progress of the old one should be shown.


Senyoret1:
I'm not sure why, but I was unable to test the changes. When using wss to access the `visors/${nodeKey}/update/ws` endpoint  I get `Error in connection establishment: net::ERR_CERT_AUTHORITY_INVALID`. When using ws I get `Error during WebSocket handshake: Unexpected response code: 404`.

Darkren:
@nkryuchkov 

Tested it. Didn't have the issue described by @Senyoret1 . but noticed these:
1. Updated single visor. after update in the right panel there was `Visor version: 0.2.4-test-4bacd5e0` which is correct. but if I click on `Update` once again I get this message: `Currently, there is no update for the visor. The currently installed version is:
- v0.1.0` - version is incorrect here
2. Updated all 3 visors with `update all`. 
2.1 after update I got the message `Operation completed. All visors are being updated.` I believe it should be like `all visors are updated` or `all visors are up to date`
2.2 trying this second time after the update I got message `Could not update any of the visors. Please, try again later.`. but really all the visors got updated successfully based on logs
2.2 when I click `Update all` after the update I get 
```
Get "https://api.github.com/repos/SkycoinProject/skywire-mainnet/releases/latest": x509: failed to load system roots and no roots provided; signal: killed. 
```
same thing may be observed in visor logs

Senyoret1:
As it worked well for Darkren, I tested again using the recommended testing procedure (as I was using the Angular test enviroment, to be able to check more in detail howthe UI is working) and found this:

Opening the manager in the same pc with the https://localhost:8000 URL it works well. However, the procedure works almost the same as it did before, so some changes would be needed. In my case, I used a virtual machine limited to an 512kbps internet connection and this was the result:

m

The process took almost 15 mins and the web socket connection only responded after the update was completed, so there was no way to show any progress to the user. The idea would be to get progress data, to be able to show the user how the update is going. Also, as there are multiple ways in which the connection may be lost, it should be possible to call the endpoint again while the update procedure is still being made and get the current state.

Appart from that, in my dev enviroment I normally run Skywire in a virtual machine and the Skywire manager in the Angular test server in the host machine. After several tests and configuration changes, I was not able to make the connection as the backend always returned a 403 response. The thing is that I'm using the same settings I use to avoid the same problem with the origin headers while using https connections (changing the origin headers to have the ip of the virtual machine running Skywire). Maybe the code used for the web sockets checks the origin in a different way? Making it work is somewhat important for front-end development.

nkryuchkov:
@Darkren @Senyoret1 Thank you for testing. I'm working on reporting updating status via WebSocket now. What is pushed now is a move from HTTP to WebSocket to avoid HTTP timeout errors and adopt the updater API for WebSocket. I'm going to include the work on reporting updating status in this PR. The errors @Darkren mentioned should be gone now. Calling the endpoint again to get the current status is a great idea, I'm going to implement it. I'll also have a look at connection issues.

Senyoret1:
@nkryuchkov Just in case, if you want to test what I did, for starting the Angular server you just have to run `npm start` in  `./static/skywire-manager-src`. That will allow you to access the manager from `https://localhost:4200`.

To make the Angular server replace the origin headers, you would have to change the line in api.service.ts from `const ws = webSocket('wss://localhost:8000/' + this.apiPrefix + url);` to `const ws = webSocket('wss://localhost:4200/ws/'  + url);` and add this to `./static/skywire-manager-src/proxy.config.json`:
```
"/ws": {
    "target": "http://localhost:8000",
    "secure": false,
    "changeOrigin": true,
    "ws": true,
    "wss": true,
    "pathRewrite": {
      "^/ws" : "/api"
    }
  }
```

nkryuchkov:
@Senyoret1 Thanks, I will try that.

Senyoret1:
@nkryuchkov Oh, sorry, I forgot to mention, after changing `./static/skywire-manager-src/proxy.config.json` the test server must be stopped and restarted or the changes will no be applied.

nkryuchkov:
@Senyoret1 Could you please have a look at the latest changes? I have implemented sending progress of an update of visor and hypervisor on the backend and showing the progress of a single visor update on the frontend. However, I wasn't able to create a quick implementation of updating all visors on the frontend due to a lack of experience with RxJS. I think I need your help with that.

jdknives (changes requested):
Great work. The changes to the updater work well in the local integration env, but when I run the PR in production with the prod or test environment, I get an empty error message and the update does not go through. 

Bildschirmfoto 2020-07-22 um 13 05 45

Senyoret1:
Great, the changes appear to be working well and I can take care of the UI issues. The only thing that would be needed is the endpoint we talked about for knowing if the visor is already being updated, but without starting the update procedure, like the `/visors/{pk}/update/ws` endpoint does.

Here is example case of why it is needed: using this pr at its current state, start updating a visor using the UI and refresh the browser just after you start seing the download progress. After the page is reloaded, press the update button again and you will see that the modal window will show again that there is an update available and ask for confirmation, as if no update procedure were being made; only after you acept the operation you will be able to see the progress again.

With an endpoint for checking if the visor is being updated, it would be possible to show the progress as soon as the user opens the modal window, among other UX improvements that can be done later.

Appart from that, is the IU showing you the number of the new version when asking for confirmation to install the update? Because it is no longer showing it to me, but I think it started to happen after an error.

nkryuchkov:
Just pushed the changes with the implemented endpoint for checking status. Route is `/visors/{pk}/update/ws/running`, response format is either `{"running":false}` or `{"running":true}`.

The wrong version in the confirmation window is okay because the endpoint for checking an update checks the stable channel by default. As `2.` in `How to test this PR` changes the channel for the updating endpoint only, versions mismatch. To set channel for the checking endpoint, the route `/visors/{pk}/update/available` should to be changed to `/visors/{pk}/update/available/{channel}`.

Senyoret1 (approved):
Great, the API is ready for improving the updater UX. Thank you :+1:

PR #444, opened 2020-07-17
Route encryption

### LINKED ISSUE #432: Encryption Module
**Feature description**

The VPN encryption currently relies on a keypair specified within the config fields for the VPN. It would improve usability a lot if we would use the local visors keypair for encryption. Maybe it makes sense to move encryption to its own module within router for instance to e2e encrypt all route traffic instead of doing encryption on an application level. 

It is hard to setup encryption for the VPN at the moment.

Darkren:
Not yet
Fixes #432 

 Changes:	
- Data packets are E2E encrypted between communicating visors


pkg/router/router.go
```diff
@@ -137,7 +139,8 @@ type router struct {
 	trustedVisors map[cipher.PubKey]struct{}
 	tm            *transport.Manager
 	rt            routing.Table
-	rgs           map[routing.RouteDescriptor]*RouteGroup // route groups to push incoming reads from transports.
+	nrgs          map[routing.RouteDescriptor]*noiseRouteGroup // noise-wrapped route groups to push incoming reads from transports.
+	rgsRaw        map[routing.RouteDescriptor]*RouteGroup      // not-yet-noise-wrapped route groups. when one of these gets wrapped, it gets removed from here
```
nkryuchkov: Could you please change both comments' case to sentence case?

pkg/router/router.go
```diff
@@ -339,30 +363,67 @@ func (r *router) serveSetup() {
 	}
 }
 
-func (r *router) saveRouteGroupRules(rules routing.EdgeRules) *RouteGroup {
+func (r *router) saveRouteGroupRules(rules routing.EdgeRules, nsConf noise.Config) (*noiseRouteGroup, error) {
 	r.logger.Infof("Saving route group rules with desc: %s", &rules.Desc)
+
+	// when route group is wrapped with noise, it's put into `nrgs`. but before that,
```
nkryuchkov: Could you please rewrite it to the sentence case?

pkg/router/router.go
```diff
@@ -339,30 +363,67 @@ func (r *router) serveSetup() {
...
 	r.mx.Lock()
-	defer r.mx.Unlock()
 
-	rg, ok := r.rgs[rules.Desc]
-	if ok && rg != nil {
-		r.logger.Infof("Route group with desc %s already exists, closing the old one and replacing...", &rules.Desc)
+	// first ensure that this rg is not being wrapped with noise right now
+	if _, ok := r.rgsRaw[rules.Desc]; ok {
+		r.mx.Unlock()
+		r.logger.Warnf("Desc %s already reserved, skipping...")
```
nkryuchkov: Missed arg for `%s`

pkg/router/router_test.go
```diff
@@ -720,8 +810,8 @@ func (e *TestEnv) GenRouterConfig(i int) *Config {
 		PubKey:           e.TpMngrConfs[i].PubKey,
 		SecKey:           e.TpMngrConfs[i].SecKey,
 		TransportManager: e.TpMngrs[i],
-		RouteFinder:      rfclient.NewMock(),
-		SetupNodes:       nil, // TODO
+		//RouteFinder:      rfclient.NewMock(),
```
nkryuchkov: Have you forgotten to uncomment it or may this line be completely removed?

nkryuchkov (changes requested):
Great job! Just minor requests

jdknives:
@Darkren do you want to do that in a separate PR or should we remove encryption logic from VPN apps?

pkg/router/router.go
```diff
@@ -398,19 +459,37 @@ func (r *router) handleDataPacket(ctx context.Context, packet routing.Packet) er
 	}
 
 	desc := rule.RouteDescriptor()
-	rg, ok := r.routeGroup(desc)
+	nrg, ok := r.noiseRouteGroup(desc)
 
 	r.logger.Infof("Handling packet with descriptor %s", &desc)
 
+	if ok {
```
evanlinjin, line 461: I know this is not your doing, but the logic defined in lines 448 and 455 where we check for the `rule.Type()` does not make any sense to me.

I think we should either use `if` statements, or the `switch/case` statements. Not both.

Also it seems that the `RuleForward` and `RuleIntermediary` should be checked first.

pkg/router/router.go
```diff
@@ -137,7 +139,8 @@ type router struct {
 	trustedVisors map[cipher.PubKey]struct{}
 	tm            *transport.Manager
 	rt            routing.Table
-	rgs           map[routing.RouteDescriptor]*RouteGroup // route groups to push incoming reads from transports.
+	nrgs          map[routing.RouteDescriptor]*noiseRouteGroup // noise-wrapped route groups to push incoming reads from transports.
```
evanlinjin: Can we rename this to `rgsNs` to be consistent? Thanks!

pkg/router/router_test.go
```diff
@@ -42,119 +44,183 @@ func TestMain(m *testing.M) {
...
+	desc := routing.NewRouteDescriptor(keys[0].PK, keys[1].PK, 1, 1)
 
 	forwardHops := []routing.Hop{
-		{From: r0.conf.PubKey, To: r1.conf.PubKey, TpID: uuid.New()},
+		{From: keys[0].PK, To: keys[1].PK, TpID: transport.MakeTransportID(keys[0].PK, keys[1].PK, dmsg.Type)},
 	}
 
 	reverseHops := []routing.Hop{
-		{From: r1.conf.PubKey, To: r0.conf.PubKey, TpID: uuid.New()},
+		{From: keys[1].PK, To: keys[1].PK, TpID: transport.MakeTransportID(keys[1].PK, keys[0].PK, dmsg.Type)},
```
evanlinjin: Do you mean:
```golang
{From: keys[1].PK, To: keys[0].PK, TpID: transport.MakeTransportID(keys[1].PK, keys[0].PK, dmsg.Type)},
```
?

pkg/router/router_test.go
```diff
@@ -42,119 +44,183 @@ func TestMain(m *testing.M) {
 	os.Exit(m.Run())
 }
 
-func Test_router_DialRoutes(t *testing.T) {
+func Test_router_NoiseRouteGroups(t *testing.T) {
```
evanlinjin, line 50: This is a rather lengthy test and a little hard to read and understand.

It would be good if we can have some summary comments describing the test as a whole and what exactly is being tested. This way one does not have to read the entire test to understand what is going on.

Thanks in advance!

evanlinjin: It would be even better if we can break the test up into more-understandable functions.

evanlinjin (changes requested):
Good work! Tested it with local env.

Please also ensure `make check` passes.

pkg/router/router_test.go
Darkren: @evanlinjin all parts of this test are connected, it's hard to move anything into a separate func. I'll leave comments there through the code

Darkren:
> @Darkren do you want to do that in a separate PR or should we remove encryption logic from VPN apps?

@jdknives I'll do it here. Sorry, just forgot about this

nkryuchkov (approved):

PR #453, opened 2020-07-30
Improvements for the update procedure

Senyoret1:

 Changes:	
A new modal window for updating the visors was added. Is has several advantages:
- Now the updating code is centraliced, which removes repeated code, improves the separation of concerns and makes the code more maintanable.
- The new component is able to parse the progress responses returned by the backend and shows a much better UI:
uw
- The new component takes into account a lot of posible cases which were not previously being correctly managed, like the nodes being updated when opening the modal window, even if only some of them are being updated.
Other changes are:
- The Spanish translation was updated.
- A new proxy rule for connection to the ws and wss endpoint with the Angular dev server was added.

How to test this PR:
Use the update options in the visor list and the visor details pages.

Known problem:

If you use the option for updating all visors, in the visor list, the UI will call the `visors/${nodeKey}/update/ws` API endpoint for all of them and will show progress for all of them. However, all visors will show the same progress, but the command window of just one visor will show progress.

This means that when calling the `visors/${nodeKey}/update/ws` API endpoint several times for updating several visors, only one visor will be updated and all web sockets will show the progress of that single visor, not mater for which visor the web socket was created.

This can be tested using `make integration-run-generic`.

cc @nkryuchkov


Senyoret1:
The problem described in the initial msg was solved in https://github.com/Senyoret1/skywire-mainnet/pull/2 and merged into the branch used for this PR. @jdknives It includes Go code, so maybe you would like to take a look.

There is also a new commit by @nkryuchkov including changes to the vendored dependencies, as the Travis tests were failing.

static/skywire-manager-src/proxy.config.json
```diff
@@ -11,5 +11,23 @@
...
+      "^/wss-api" : "/api"
+    }
+  },
+  "/ws-api": {
+    "target": "ws://127.0.0.1:8000",
+    "secure": false,
+    "ws": true,
+    "headers": {"host":"127.0.0.1:8000", "origin":"ws://127.0.0.1:8000", "referer":"ws://127.0.0.1:8000"},
+    "pathRewrite": {
+      "^/wss-api" : "/api"
```
nkryuchkov: @Senyoret1 I don't know how this exactly works, but it looks like `wss-api` may be misspelled. Should it be `wss-api` or `ws-api`?

nkryuchkov (approved):
It works fine. Great job!

static/skywire-manager-src/proxy.config.json
Senyoret1: You are right, fixed in the last commit 👍

Senyoret1:
- The app protocol version was removed from the right bar, as requested a few days ago.
- A small problem with the proxy configuration was fixed.

jdknives:
@Senyoret1 good job. UI looks much cleaner. Please fix the conflicts and we can merge this.

Senyoret1:
I force pushed changes to remove https://github.com/skycoin/skywire/pull/453/commits/024fd3986d41200df05b0ed7322d529fbf43225c from the PR, for solving most of the conflicts as the changes were already added in https://github.com/skycoin/skywire/pull/454 .

PR #470, opened 2020-08-13
Fix VPN client remote startup

### LINKED ISSUE #467: Starting VPN client leads to connection shutdown
**Describe the bug**
When starting the VPN client through the hypervisor on a remote Skybian board, the VPN client logs that there are not remote PubKeys set. Upon closing the application through the hypervisor, the request times out and the connection between visor and hypervisor is subsequently shut down. The board becomes unusable until it is restarted. 

**Environment information:**

- OS: e.g. Armbian

Bildschirmfoto 2020-08-11 um 20 07 36

Possible hints: 

I would think this is related to the client not being started up properly and running into some error.

Darkren:
Yes
Fixes #467 

 Changes:	
- VPN client now gracefully handles slice of direct IPs containing equal ones
- `Wait` is now called correctly for app process. This way app process correctly finishes in case of failure and hypervisor correctly gets its status

How to test this PR:
1. Start hypervisor
2. Start visor
3. From web UI run the app which is said to fail (i.e. VPN client without remote PK set)
4. You will see that the process started in the UI message, which is true, but you also will see an error in app logs. App will stay down in the UI. It may also happen that app will appear is running. But it will get down during next status update

### DIFF
--- a/cmd/apps/vpn-client/vpn-client.go
+++ b/cmd/apps/vpn-client/vpn-client.go
@@ -5,6 +5,7 @@ import (
 	"flag"
 	"net"
 	"os"
+	"os/exec"
 	"os/signal"
 	"syscall"
 	"time"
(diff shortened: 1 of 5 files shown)


jdknives (approved):
Fixes the issue. This can be merged after a codereview.

nkryuchkov (approved):

PR #486, opened 2020-08-17
Add an option for removing all offline transports

Senyoret1:

 Changes:	
- The transport list now has an option for removing all offline transports. This helps to manage large amounts of unused transports when using the vpn apps.

- Now the option for removing all offline visors, in the visor list, takes into account the filters and shows an alert informing that. There are also some small UX improvements. The option for removings all offline transports also have those changes.

How to test this PR:
Using the Skywire manager, go to the transport list of a visor. The option is shown by clicking the button with the 3 horizontal dots at the top-right corner of the transport list.


jdknives (approved):
Good work!

PR #490, opened 2020-08-24
Implement Route Multiplexer

### LINKED ISSUE #440: Implement Route Multiplexer
**Feature description**

Currently applications can make use of only one route to a remote. Applications such as the VPN would benefit quite heavily from a route multiplexing module that allows to open and use multiple routes in case of 

- congestion on a route

- a route drops

This way, a VPN could create multiple routes upon startup and switch from one to the next one in case a route breaks down smoothly without an interruption. 

**Possible implementation**

The route multiplexer should request 3 routes and always have two routes as backup. Upon failure of one route, the multiplexer should switch to the next route.

nkryuchkov:

Fixes #440

Changes:	
- Implement Route Multiplexer with backward compatibility 

How to test this PR:
- Run the generic integration env, it should set up the following transports: `A -> B`, `C -> B`
- Set up a `C -> A` transport (`./bin/skywire-cli --rpc $RPC_A visor add-tp $PK_C --type stcpr`)
- Send a message from chat C to chat A (`curl --data {'"recipient":"'$PK_A'", "message":"Hello"}' -X POST $CHAT_C`)
- The message should appear in `VisorA` logs; logs of sending should appear in `VisorC` logs, as well as a message what transport was used (it should be `C -> A` and a total amount of transports)
- Delete the `C -> A` transport (`./bin/skywire-cli --rpc $RPC_C visor rm-tp cad4c72c-90fa-0117-b225-41eca240e785`)
- Send the message again
- Message should appear in `VisorA`'s logs and be sent from `VisorC` using the 2nd transport via `VisorB`


pkg/router/route_group.go
```diff
@@ -164,21 +165,47 @@ func (rg *RouteGroup) Write(p []byte) (n int, err error) {
 	}
 
 	rg.mu.Lock()
-	tp, err := rg.tp()
-	if err != nil {
-		rg.mu.Unlock()
-		return 0, err
+	tps := make([]*transport.ManagedTransport, len(rg.tps))
```
Darkren: I know that we've been doing it this way for a long time. but what do you think if we get rid of mutex usage and copying in here? afaik we initialize `tps` and `fwd` at the point of route group creation. and after that these slices are actually not changed. also `rvs` is a subject for discussion, it seems like it's not used at all

pkg/router/route_group.go
```diff
@@ -164,21 +165,47 @@ func (rg *RouteGroup) Write(p []byte) (n int, err error) {
...
+	for i := range tps {
+		rule := rules[i]
+		tp := tps[i]
+
+		if tp == nil {
+			rg.logger.Infof("Bad transport at index %v, using the next one", i)
+			continue
+		}
+
+		rg.logger.Infof("Attempting to use TP %v (%v -> %v) (%v of %v) for writing",
```
Darkren: Should we use `debug` level in here?

Darkren (changes requested):
Looks great. Just one small comment and a suggestion to discuss

pkg/router/route_group.go
nkryuchkov: @Darkren I agree

pkg/router/route_group.go
nkryuchkov: @Darkren Good idea

PR #491, opened 2020-08-25
VPN improvements

### LINKED ISSUE #473: Improvements to the VPN
**Feature description**

In order to increase the convenience with which the VPN can be used, we need to expose more information from the VPN client to the hypervisor. 

Among the most obvious ones we should focus on are:

1. Connection status --> is the connection to the VPN server working

2. Latency/throughput of the connection

3. Amount of bandwidth sent.

Once these are metrics can be collected on the client side, we need to expose them to the hypervisor.

Darkren:
Yes
Fixes #473 

 Changes:	
- Packet of type `NetworkProbe` added. Packet handling goes this way. Once an interval of time (3 seconds now) route group sends network probe to the remote. Packet includes timestamp of when the packet was sent and a throughput of the remote. Throughput is an amount of bytes per second successfully delivered to the remote. So the source route group tells remote the throughput of the remote. Upon receiving this packet remote calculates latency based on the timestamp in a packet;
- Amount of bandwidth sent/receiced is calculated on each sent/received packet;
- `NoiseRouteGroup` and `SkywireConn` are exposed to allow outer code get RG's metrics;
- `IsAlive` method is added to route group. RG is considered alive as soon as underlying route is functional;
- Method to fetch connections metrics for a specific app is added to hypervisor.

How to test this PR:
1. Run any app. VPN for example
2. Request data from hypervisor, like this:
```
http://{HYPERVISOR_URL}/api/visors/{VISOR_PK}/apps/{APP_NAME}/connections
```


pkg/router/network_stats.go
```diff
@@ -0,0 +1,67 @@
+package router
+
+import (
+	"sync"
+	"sync/atomic"
+	"time"
+)
+
+type networkStats struct {
+	bandwidthSent     uint32
```
nkryuchkov: Would 32 bit be enough for bandwidth?

nkryuchkov (approved):
Good job! Works great.

jdknives (approved):
Very smooth. Please respond to the request by Nikita and change target branch to v0.4.0 and then this can go.

PR #507, opened 2020-09-01
Adapt route encryption for old visors

### LINKED ISSUE #501: There are problems connecting and old visor with a new one
**Describe the bug**
Trying to connect an old version of a Skywire visor with a new one has some problems.

**Environment information:**

- OS: Linux (Ubuntu 18.04.1)

**Steps to Reproduce**

1. Start a node updated to the latest changes of the develop branch, and another one updated until the 3c87c1a5 commit.
2. Create a transport connecting both nodes.
3. Use the UI of the Skychat app and try to send a msg from the first visor to the second one.

**Actual behavior**
The msg will not be sent. The UI of the second visor will show something like this:
err
Also, the console will show an error like `error="error performing noise handshake: deadline exceeded"`.

**Expected behavior**
The error should not appear and the message should be sent.

**Additional context**
Previously I had a similar problem connection an old version of the mobile visor with an updated visor running in a desktop computer. The error in that case was `error performing noise handshake: noise prefix value 31522B exceeds maximum 4094B, closing...`. The mobile visor used was the one in the e310948 commit from https://github.com/skycoin/skywire/pull/447

Darkren:
Yes
Fixes #501 

 Changes:	
- Handshake packet type was added;
- Visor now exchange handshake packets while creating route groups. Initiator sends it first, remote visor responds. There are 3 possible cases here:
1) both visors support route encryption. in this case handshake packets are exchanged, visors start communicating with the traffic encrypted;
2) initiator supports encryption, while remote visor is not. in this case initiator sends handshake and waits for response. either it gets no response during timeout interval (2 seconds now) or it gets data packet. in both cases remote visor is known to not support the encryption, communication goes without it;
3) initiator doesn't support encryption, while remote visor does. in this case remote visor waits for the handshake packet to come in. again, either 2 seconds timeout passes or data packet arrives. in both cases initiator is known to not support the encryption, communication goes without it.

How to test this PR:
1. Setup 2 visors with some apps (VPN server and client, for example).
2. Try to communicate. Should work for all visors supporting encryption and not.


pkg/router/route_group.go
```diff
@@ -402,6 +407,31 @@ func (rg *RouteGroup) sendKeepAlive() error {
...
+func (rg *RouteGroup) sendHandshake(encrypt bool) error {
+	rg.mu.Lock()
+
+	if len(rg.tps) == 0 || len(rg.fwd) == 0 {
+		rg.mu.Unlock()
+		// if no transports, no rules, then no keepalive
+		return nil
+	}
+
+	rule := rg.fwd[0]
```
nkryuchkov: Could you please handle all rules from `rg.fwd` and transports from `rg.tps` instead of just the 1st ones?

transport_logs/33612f81-905c-0fd5-b2e4-ee4602e7327c.log
```diff
@@ -1 +0,0 @@
-{"recv":258,"sent":186}
```
nkryuchkov, line 1: Could you please add this directory to `.gitignore`?

nkryuchkov (changes requested):

pkg/router/route_group.go
nkryuchkov: Done in the following commits.

transport_logs/33612f81-905c-0fd5-b2e4-ee4602e7327c.log
nkryuchkov: Done in the following commits.

nkryuchkov (approved):

PR #510, opened 2020-09-07
Attach started by updater binaries to TTY of their parent

### LINKED ISSUE #336: Visor does not receive terminal commands after update
**Describe the bug**
After updating via the `/update` endpoint, the visor runs a new binary with the same stdin/out/err. However, the new binary cannot receive terminal commands like `CTRL-C`.

**Steps to Reproduce**

1. Update version in `Makefile` to `v0.1.0`
2. Run integration env
3. Call the `/update` endpoint (`curl -k -X POST http://localhost:8000/api/visors/03374f658c7107071ae35fc1d030e926fb99e461dd13662988d1306cb823ab05d4/update`)
4. After the update, try to kill a new process with `CTRL-C`. It will remain run (check `ps aux | grep visor`).

**Actual behavior**
Visor remains run after `CTRL-C`.

**Expected behavior**
Visor stops after `CTRL-C`.

**Additional context**
The signal is not received because a new process is not attached to the same controlling terminal (CTTY). This can be checked by `ps aux` after the update. Before the update, the visor has `S+` attributes, after that, it has `S`, which means it's not attached to a CTTY. 

**Possible implementation**
We need to set the same CTTY for a new process, which can be set via `SysProcAttr` field of `exec.Cmd`. However, from what I've found, it's impossible to set the same CTTY from the parent process. So we need to find some workarounds. 

The following links might help to implement the desired behavior.
https://blog.nelhage.com/2011/02/changing-ctty/
https://blog.habets.se/2009/03/Moving-a-process-to-another-terminal.html
https://linux.die.net/man/4/tty_ioctl (`TIOCSCTTY`)
https://github.com/golang/go/issues/20454
https://github.com/google/goexpect/blob/master/expect.go#L929
https://github.com/golang/go/issues/29458
https://stackoverflow.com/questions/593724/redirect-stderr-stdout-of-a-process-after-its-been-started-using-command-lin
https://unix.stackexchange.com/questions/31824/how-do-i-attach-a-terminal-to-a-detached-process
https://www.gnu.org/software/libc/manual/html_node/Job-Control.html

nkryuchkov:

Fixes #336 

 Changes:	
- Attach started by updater binaries to TTY of their parent

How to test this PR:
- Run the integration env
- Run the following:
    - macOS: `curl -k -d '{"checksums_url":"https://github.com/nkryuchkov/skywire/releases/download/v0.0.1-alpha.test.tty/checksums.txt", "archive_url":"https://github.com/nkryuchkov/skywire/releases/download/v0.0.1-alpha.test.tty/skywire-v0.0.1-alpha.test.tty-darwin-amd64.tar.gz", "version": "v0.0.1-alpha.test.tty"}' -X POST https://localhost:8000/api/visors/024ec47420176680816e0406250e7156465e4531f5b26057c9f6297bb0303558c7/update
`
    - Linux/amd64: `curl -k -d '{"checksums_url":"https://github.com/nkryuchkov/skywire/releases/download/v0.0.1-alpha.test.tty/checksums.txt", "archive_url":"https://github.com/nkryuchkov/skywire/releases/download/v0.0.1-alpha.test.tty/skywire-v0.0.1-alpha.test.tty-linux-amd64.tar.gz", "version": "v0.0.1-alpha.test.tty"}' -X POST https://localhost:8000/api/visors/024ec47420176680816e0406250e7156465e4531f5b26057c9f6297bb0303558c7/update
`
- The updating of `VisorA` should finish with no errors and an updated binary should be started
- Press Ctrl-C, the updated binary should be stopped


Darkren (approved):

jdknives (approved):
Great work. Runs smoothly now.

PR #520, opened 2020-09-14
Add password options to the UI of the proxy and vpn client apps

### LINKED ISSUE #519: Missing frontend support for setting VPN client passcode arg
**Describe the bug**
In the hypervisor UI it is currently possible for a user to set a passcode on a `vpn-server` but not possible to specify it for the `vpn-client`. That means it is impossible for users to set the right arguments for them to connect to a passcode protected `vpn-server`. A field for filling in the passcode would be needed. I am fairly sure the endpoint supports setting a passcode already and a remote pubkey already.

Senyoret1:

Fixes #519

 Changes:	
- Now the user can add a password when configuring the `vpn-client` and `skysocks-client` apps.
- When selecting a history entry created with a password, a modal window is shown for the user to enter the password again, as it is not saved.

How to test this PR:
Using the manager open the app configuration modal window for the `vpn-client` and `skysocks-client` apps. You will be able to enter a password manually.

NOTE: the changes in this PR don’t work totally well because of 2 things:
- When changing the password of the `skysocks-client` app, the server returns `App skysocks-client is not allowed to change password` as error. For the PR work well, the backend has to be changed to avoid that error. The password field could be removed when changing the configuration of the `skysocks-client` app, to avoid the problem, but that does not appear to be a good solution, as it is possible to set a password for the `skysocks` app, so the situation would be similar to https://github.com/skycoin/skywire/issues/519

- The hypervisor API does not allow to remove the `-passcode` setting. Because of that, if the user does not enter a password the manager sends an empty string to the backend as the password, so the app configuration will have an empty string as password. If that solution is ok, then there is no problem with this point.

### DIFF
--- a/static/skywire-manager-src/src/app/components/pages/node/apps/node-apps/skysocks-client-settings/skysocks-client-settings.component.ts
+++ b/static/skywire-manager-src/src/app/components/pages/node/apps/node-apps/skysocks-client-settings/skysocks-client-settings.component.ts
@@ -21,6 +21,7 @@ import {
   FilterWindowData
 } from './skysocks-client-filter/skysocks-client-filter.component';
 import { countriesList } from 'src/app/utils/countries-list';
+import { SkysocksClientPasswordComponent } from './skysocks-client-password/skysocks-client-password.component';
 
 /**
  * Data of the entries from the history.
@@ -43,6 +44,10 @@ export interface HistoryEntry {
    * Custom note added by the user.
    */
   note?: string;
+  /**
+   * If the user entered a password.
+   */
+  hasPassword?: boolean;
 }
 
 /**
@@ -168,6 +173,7 @@ export class SkysocksClientSettingsComponent implements OnInit, OnDestroy {
         Validators.maxLength(66),
         Validators.pattern('^[0-9a-fA-F]+$')])
       ],
+      'password': ['', Validators.maxLength(100)]
     });
 
     setTimeout(() => (this.firstInput.nativeElement as HTMLElement).focus());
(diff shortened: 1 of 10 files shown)


nkryuchkov:
@Senyoret1 

There's no password setting for `skysocks-client` app, it just redirects credentials to `skysocks`, therefore it's present in the list of apps that are allowed to change the password.

I've implemented removing args if their passed value is empty: https://github.com/skycoin/skywire/pull/522

static/skywire-manager-src/src/app/components/pages/node/apps/node-apps/skysocks-client-settings/skysocks-client-settings.component.ts
```diff
@@ -406,51 +412,88 @@ export class SkysocksClientSettingsComponent implements OnInit, OnDestroy {
     });
   }
 
+  /**
+   * Makes the app use the data from a history entry.
+   * @param entry Entry to be used.
+   */
+  useFronHistory(entry: HistoryEntry) {
```
nkryuchkov: Did you mean `useFromHistory`?

static/skywire-manager-src/src/assets/i18n/en.json
```diff
@@ -384,6 +386,13 @@
         "note": "Note"
       },
 
+      "password-dialog": {
+        "title": "Enter Password",
+        "password": "Password",
+        "info": "You are being asked for a password because a a password was set when the selected entry was created, but the it was not saved for security reasons. You can leave the password empty if needed.",
```
nkryuchkov: It seems there's a redundant `a` in `because a a password was set when the selected entry was created`

static/skywire-manager-src/src/assets/i18n/es_base.json
```diff
@@ -382,6 +384,13 @@
         "note": "Note"
       },
 
+      "password-dialog": {
+        "title": "Enter Password",
+        "password": "Password",
+        "info": "You are being asked for a password because a a password was set when the selected entry was created, but the it was not saved for security reasons. You can leave the password empty if needed.",
```
nkryuchkov: Same here

nkryuchkov (changes requested):
Works great. Just minor requests.

Senyoret1:
@nkryuchkov Thanks for the code review. I removed the password field for the `skysocks-client` app and made the changes you mentioned, so everything should be ready after merging #522

Issue #383, opened 2020-06-01
Properly use callbacks in VPN


jdknives:
**Feature description**

We recently implemented callbacks for dmsg connections and disconnections. @Darkren mentioned on discord, that we need to properly implement them. 

We need to properly  setup callbacks from VPN client. 

For STcp:
- On each connection we add direct route to the remote node via router
- On each disconnection we remove direct route to the remote node via router

For Dmsg:
- On each TCP dial to the Dmsg server we add a direct route to that server via router
- On each TCP close we remove a direct route to the remote Dmsg server via router

PR #523, opened 2020-09-16
Make VPN work properly over stcpr

### LINKED ISSUE #517: VPN Error during route group close
**Describe the bug**
When running the VPN client, a number of errors pop up after a certain amount of time that do not break the VPN clients connection but should be investigated. The logs for the remote and the local visor are attached. 

**Environment information:**

- OS: e.g. Mac OS X

**Steps to Reproduce**

1. Setup VPN client and server
2. Setup transport between two visors
3. Run VPN client
4. Use the VPN and wait until the error pops up (took usually around a minute for me)

Darkren:
Yes
Fixes #517, partially implements #383 

 Changes:	
- Callbacks for stcpr/stcp connection establishment implemented;
- These callbacks are properly used by the VPN client app;
- Remote IPs for all of the known STCPR transports are now passed to the VPN client on start;

How to test this PR:
Setup VPN connection over STCPR transport

### DIFF
--- a/cmd/apps/vpn-client/vpn-client.go
+++ b/cmd/apps/vpn-client/vpn-client.go
@@ -9,14 +9,13 @@ import (
 	"syscall"
 	"time"
 
-	"github.com/skycoin/skywire/pkg/app/appevent"
-
 	"github.com/sirupsen/logrus"
 	"github.com/skycoin/dmsg/cipher"
 	"github.com/skycoin/dmsg/netutil"
 
 	"github.com/skycoin/skywire/internal/vpn"
 	"github.com/skycoin/skywire/pkg/app"
+	"github.com/skycoin/skywire/pkg/app/appevent"
 	"github.com/skycoin/skywire/pkg/app/appnet"
 	"github.com/skycoin/skywire/pkg/routing"
 	"github.com/skycoin/skywire/pkg/skyenv"
--- a/cmd/apps/vpn-client/vpn-client.go
+++ b/cmd/apps/vpn-client/vpn-client.go
@@ -9,6 +9,8 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/skycoin/skywire/pkg/app/appevent"
+
 	"github.com/sirupsen/logrus"
 	"github.com/skycoin/dmsg/cipher"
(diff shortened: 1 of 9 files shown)


jdknives (approved):
Works well. @nkryuchkov after you review we can merge.

cmd/apps/vpn-client/vpn-client.go
```diff
@@ -70,7 +71,29 @@ func main() {
...
 
-	appClient := app.NewClient(nil)
+	var directIPsCh = make(chan net.IP, 100)
+
+	eventSub := appevent.NewSubscriber()
+
+	eventSub.OnTCPDial(func(data appevent.TCPDialData) {
+		log.WithField("event_type", data.Type()).
+			WithField("event_data", data).
+			Info("GOT EVENT")
```
nkryuchkov: What do you think about changing the case of this message to the sentence case?

pkg/snet/network.go
```diff
@@ -146,6 +146,12 @@ func New(conf Config, eb *appevent.Broadcaster) (*Network, error) {
 			SK:        conf.SecKey,
 			Table:     pktable.NewTable(conf.NetworkConfigs.STCP.PKTable),
 			LocalAddr: conf.NetworkConfigs.STCP.LocalAddr,
+			BeforeDialCallback: func(network, addr string) error {
+				data := appevent.TCPDialData{RemoteNet: network, RemoteAddr: addr}
+				event := appevent.NewEvent(appevent.TCPDial, data)
+				_ = eb.Broadcast(context.Background(), event) //nolint:errcheck
```
nkryuchkov, line 152: Could you please explain why an error is ignored here instead of returning?

nkryuchkov (approved):

pkg/snet/network.go
Darkren: @nkryuchkov dial will be cancelled in case of error, cause callback is fired before dialing

cmd/apps/vpn-client/vpn-client.go
Darkren: @nkryuchkov that's debug log, just missed that, will remove

PR #534, opened 2020-09-25
Make VPN client runnable without root

### LINKED ISSUE #382: Avoid running VPN as root
**Feature description**

Currently we need to run `VPN-client` and `VPN-server` as root in order to successfully create the TUN interfaces. It would be preferable to create a new user with access to only those directories and parts of the system needed to create the TUN interface. 

Skywire VPN needs to run as root currently.

Darkren:
Yes
Fixes #382 

 Changes:	
- VPN client now is runnable by non-root users. But, the binary has to be owned by root and have permissions like 4755, so any user could run it
- Linux capabilities are used for Linux version.

How to test this PR:
`Mac OS`:
1. `make bin`
2. `make host-apps`
3. `sudo chown root ./apps/vpn-client`
4. `sudo chmod 4755 ./apps/vpn-client`
5. `./skywire-visor ./skywire-config.json`

`Linux`:
1. `make bin`
2. `make host-apps`
3. `sudo setcap 'cap_net_admin+p' ./apps/vpn-client`
4. `./skywire-visor ./skywire-config.json`


internal/vpn/os_linux.go
```diff
@@ -42,3 +43,35 @@ func AddRoute(ip, gateway string) error {
 func DeleteRoute(ip, gateway string) error {
 	return run("ip", "r", "del", ip, "via", gateway)
 }
+
+var (
```
nkryuchkov: Could we get rid of this block and return errors from `syscall` directly? I think this doesn't improve readability.

internal/vpn/os_linux.go
```diff
@@ -42,3 +43,35 @@ func AddRoute(ip, gateway string) error {
...
+		return errEINVAL
+	case syscall.ENOENT:
+		return errENOENT
+	}
+	return e
+}
+
+// Setuid sets uid of current OS user.
+func Setuid(uid int) (err error) {
+	// due to iss. 1435 `syscall.Setuid` is disabled for Linux system, so we'
```
nkryuchkov: Could you please add the link (https://github.com/golang/go/issues/1435) to comment? That could be a bit clearer. Also, `so we'` looks unfinished.

Darkren: Not used anymore

internal/vpn/os_linux.go
Darkren: Not used anymore

nkryuchkov (approved):

jdknives (approved):

PR #552, opened 2020-10-18
Improvements for configuring the VPN from the manager

Senyoret1:

 Changes:	
- Now the modal window for configuring the `vpn-client` and `skysocks-client` apps allows to copy the public keys from the backend list. Also, the modal window is not closed after changing the backend.

- A small bug which made the history buttons not work was solved.

- The style of the buttons in the history list was changed a bit.

How to test this PR:
Open the modal window for configuring the `vpn-client` or `skysocks-client` app and use it.


jdknives (approved):
Good job.

PR #577, opened 2020-10-24
Make the manager work with the integrated hypervisor

### LINKED ISSUE #562: 'Copy hypervisor pubkey to clipboard' function in hypervisor UI
**Feature description**

It would be helpful when setting up new visors to easily access the pubkey of the hypervisor via
a copy to clipboard function inside the hypervisor UI

No, it would just be simpler instead of opening the config file

Like described above. I think the function could be implemented in the dropdown menu in the right upper corner

Senyoret1:

Fixes #562 (to some extend).

 Changes:	
- The option for updating the Hypervisor was removed from the manager, as it is now Hypervisor is integrated in the visor.
- The manager now shows which visor is currently being used as Hypervisor.

How to test this PR:
Open the Skywire manager. The visor list now has a column at the left which contains a star if the visor is the current hypervisor. Also, the option for updating the Hypervisor is no longer present in the options menu of the visor list.


static/skywire-manager-src/src/assets/i18n/en.json
```diff
@@ -131,14 +131,15 @@
...
-    "update-hypervisor": "Update hypervisor",
     "update-all": "Update all visors",
+    "hypervisor": "Hypervisor",
     "state": "State",
     "state-tooltip": "Current state",
     "label": "Label",
     "key": "Key",
     "dmsg-server": "DMSG server",
     "ping": "Ping",
+    "hypervisor-info": "The visor contains the currently used Hypervisor.",
```
jdknives: `This visor is the current Hypervisor.` sounds better. Other than that good job.

Senyoret1:
@jdknives The last commit updates the tooltip text.

jdknives (approved):

Issue #572, opened 2020-10-22
-p flag for skywire-cli visor & skywire-hypervisor gen-config to generate correct default config for skybian (or package-based installations)


0pcom:
**Feature description**

-p flag for skywire-cli visor & skywire-hypervisor gen-config to generate correct default config for skybian (or package-based installations)

example for the hypervisor:

```
[user@linux  ~]$ skywire-hypervisor gen-config -rpo /etc/skywire-hypervisor.json
[2020-10-20T18:48:15-05:00] INFO [pathutil]: Wrote 713 bytes to /etc/skywire-hypervisor.json
{
	"public_key": "03601fb44cbf660f491cdc2cf56af362503ee34e74008647c135330118e00a0cd0",
	"secret_key": "442cec5049b5dbee3fd90199d5f1c46b8338567ee5d371456e784fb78d67ff1d",
	"db_path": "/var/skywire-hypervisor/users.db",
	"enable_auth": true,
	"cookies": {
		"hash_key": "b8da4b60434c0998310caf44bdcafb52817deb9b43bba8bd2c0a9decb0b491f997daad0b4c24d6c7625ddd1872a1a903dec2ec410cca6828879f3b9a7ea50483",
		"block_key": "46afd5872c3870721d4031455823ece7e95e00569e0464c28b70a845c61b368d",
		"expires_duration": 43200000000000,
		"path": "/",
		"domain": ""
	},
	"dmsg_discovery": "http://dmsg.discovery.skywire.skycoin.com",
	"dmsg_port": 46,
	"http_addr": ":8000",
	"enable_tls": true,
	"tls_cert_file": "/etc/skywire-hypervisor/cert.pem",
	"tls_key_file": "/etc/skywire-hypervisor/key.pem"
}

```

skywire-cl desired example usage:

```
skywire-cli visor gen-config -rpo /etc/skywire-visor.json
[2020-10-20T18:35:10-05:00] INFO [visor:config]: Flushing config to file. config_version="v1.0.0" filepath="/etc/skywire-visor.json"
[2020-10-20T18:35:10-05:00] INFO [visor:config]: Flushing config to file. config_version="v1.0.0" filepath="/etc/skywire-visor.json"
[2020-10-20T18:35:10-05:00] INFO [skywire-cli]: Updated file '/etc/skywire-visor.json' to: {
	"version": "v1.0.0",
[...]

Issue #564, opened 2020-10-19
0.3.0 will require manual config file reformatting; visors will not start after update


0pcom:
With the binary release of 0.3.0 on amd64; I upgraded from 0.2.3 via the skywire package I maintain in the Arch User Repos

```
[user@linux  ~]$ cat /etc/skywire-visor.json
{
	"version": "v1.0.0",
	"sk": "8fddfb6c56623f7c9969a663c1843c1a6c836c3d5234f9cc37a4c1bd6e420e38",
	"pk": "03d430ef9c93cbe3d8aa9bce1a79305356e00accbc300241b1223c60cbaf823337",
	"dmsg": {
		"discovery": "http://dmsg.discovery.skywire.skycoin.com",
		"sessions_count": 1
	},
	"dmsgpty": {
		"port": 22,
		"authorization_file": "./dmsgpty/whitelist.json",
		"cli_network": "unix",
		"cli_address": "/tmp/dmsgpty.sock"
	},
	"stcp": {
		"pk_table": null,
		"local_address": ":7777"
	},
	"transport": {
		"discovery": "http://transport.discovery.skywire.skycoin.com",
		"address_resolver": "http://address.resolver.skywire.skycoin.com",
		"log_store": {
			"type": "file",
			"location": "./transport_logs"
		},
		"trusted_visors": null
	},
	"routing": {
		"setup_nodes": [
			"0324579f003e6b4048bae2def4365e634d8e0e3054a20fc7af49daf2a179658557"
		],
		"route_finder": "http://routefinder.skywire.skycoin.com",
		"route_finder_timeout": "10s"
	},
	"uptime_tracker": {
		"addr": "http://uptime-tracker.skywire.skycoin.com"
	},
	"launcher": {
		"discovery": {
			"update_interval": "30s",
			"proxy_discovery_addr": "http://service.discovery.skycoin.com"
		},
		"apps": [
			{
				"name": "skychat",
				"args": [
					"-addr",
					":8001"
				],
				"auto_start": true,
				"port": 1
			},
			{
				"name": "skysocks",
				"auto_start": true,
				"port": 3
			},
			{
				"name": "skysocks-client",
				"auto_start": false,
				"port": 13
			},
			{
				"name": "vpn-server",
				"auto_start": false,
				"port": 44
			},
			{
				"name": "vpn-client",
				"auto_start": false,
				"port": 43
			}
		],
		"server_addr": "localhost:5505",
		"bin_path": "./apps",
[...]

PR #579, opened 2020-10-26
package defaults with gen-config -p

0pcom:
yes

Fixes #
https://github.com/skycoin/skywire/issues/572

 Changes:	
-	add -p flag to config generation commands which generates correct default values for skybian and package-based installations.

How to test this PR:

recompile the binaries
```
./hypervisor gen-config -p
./skywire-cli visor gen-config -p
```

Notes:

I did not make extensive efforts to integrate it with the existing logic of the other flags, nor did I test combinations of these flags. It solves my immediate issue and can be improved upon in the future.

I request that this PR be revewed / approved as soon as possible as I wish to update to the VPN-containing version but as stated in https://github.com/skycoin/skywire/issues/564 when I tested this on my machine, the visor failed to start after the update because of issues with the config. I have not tested or attempted to reproduce https://github.com/skycoin/skywire/issues/564 on arm64, and no other user has complained of this to my knowledge, so perhaps it is some anomaly to do with the architecture or specifically that binary. I don't know.

However, I can say that the content of this pull request is critical for reliably automating configuration at the point of installation, eliminating the need for manual revision of the config file.


jdknives (commented):
@0pcom thanks for this PR. Generally this is a good idea and we will go with it. We will use the `-p` flag to generate configs for package installations. Two points:

1. We will only release official packages with the next release. Therefore, this should target `develop` instead of `master`. The config will change on that release (you can check the changes by testing develop) a bit. We got rid of the `hypervisor` binary and instead a visor can now expose the UI if config was generated with `--hypervisor` flag. That necessitates some changes to this PR. 

2. We have discussed with Synth to put package based installations to `$HOME/skywire` long term as it does not require root access. https://github.com/skycoin/skywire/issues/581 
In anticipation of that change, you can probably already change your PR here.

cmd/hypervisor/commands/gen-config.go
```diff
@@ -17,21 +17,24 @@ import (
 var (
 	output        string
 	replace       bool
+	package1      bool
```
nkryuchkov, line 20: Could you please rename it to something like `packageConfig`?

cmd/hypervisor/commands/gen-config.go
```diff
@@ -68,6 +71,9 @@ var genConfigCmd = &cobra.Command{
 				log.Fatalln("Error retaining old keys", err)
 			}
 		}
+		if package1 {
```
nkryuchkov, line 74: What do you think of moving this condition before the `switch` statement and having `switch` inside an `else`?

cmd/skywire-cli/commands/visor/gen-config.go
```diff
@@ -19,16 +19,18 @@ func init() {
...
 
 var (
-	sk      cipher.SecKey
-	output  string
-	replace bool
-	testEnv bool
+	sk       cipher.SecKey
+	output   string
+	replace  bool
+	package1 bool
```
nkryuchkov, line 25: Could you please rename it to something like `packageConfig`?

0pcom:
> @0pcom thanks for this PR. Generally this is a good idea and we will go with it. We will use the `-p` flag to generate configs for package installations. Two points:
> 
> 1. We will only release official packages with the next release. Therefore, this should target `develop` instead of `master`. The config will change on that release (you can check the changes by testing develop) a bit. We got rid of the `hypervisor` binary and instead a visor can now expose the UI if config was generated with `--hypervisor` flag. That necessitates some changes to this PR.
> 2. We have discussed with Synth to put package based installations to `$HOME/skywire` long term as it does not require root access. #581
>    In anticipation of that change, you can probably already change your PR here.

I suppose I should retarget this PR at develop but I am unfamiliar with how to do that at this stage without filing a new PR. Additionally it would appear that this would require some revision to work with the new setup that you have described. I will take your advice on where to go from here.

On the installation path, it would be OK to use the $HOME/skywire path except that by default the service is running as root. These paths should be created at /root/skywire by default when the config generation is preformed as root.

Is it possible at this stage to run skywire as an unprivlaged user? The last time I tried to I recall there was some issue.

@nkryuchkov i will be able to preform the requested changes tonight. Per the switch / else; I have no opposition but I will have to figure out how to implement that.

jdknives:
@0pcom I just redirected the PR at `develop`. The merge conflicts that popped up need to be fixed. If you have not worked before with go, we can tackle the content of the PR ourselves.

0pcom:
re-pening from a different branch

PR #593, opened 2020-11-04
Feature/allow passing hypervisor pk

### LINKED ISSUE #590: Allow passing hypervisor PubKey
**Feature description**

Allow to pass in hypervisor PK during `skywire-cli visor gen-config`. This should set the PubKey of hypervisor in visors `hypervisors` field.

Adding a hypervisor currently requires manual intervention in the configuration file. With a flag to set it during config generation, we could make efforts to automate the hypervisor setting.

i-hate-nicknames:
Yes

Fixes #590 

 Changes:	
- adds `hypervisor-pk` to `skywire-cli visor gen-config` which takes a pk of some hypervisor and adds it to the hypervisors list
in the generated config

How to test this PR:
`make test`

### DIFF
--- a/cmd/skywire-cli/commands/visor/gen-config.go
+++ b/cmd/skywire-cli/commands/visor/gen-config.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/sirupsen/logrus"
 	"github.com/skycoin/dmsg/cipher"
+	coinCipher "github.com/skycoin/skycoin/src/cipher"
 	"github.com/skycoin/skycoin/src/util/logging"
 	"github.com/spf13/cobra"
 
@@ -76,6 +77,14 @@ var genConfigCmd = &cobra.Command{
 		conf, err := genConf(mLog, output, &sk, hypervisor)
 		if err != nil {
 			logger.WithError(err).Fatal("Failed to create config.")
+		}
+
+		if hypervisorPK != "" {
+			key, err := coinCipher.PubKeyFromHex(hypervisorPK)
+			if err != nil {
+				logger.WithError(err).Fatal("Failed to parse hypervisor private key.")
+			}
+			conf.Hypervisors = append(conf.Hypervisors, cipher.PubKey(key))
 		}
 
 		// Save config to file.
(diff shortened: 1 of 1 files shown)


0pcom:
I like this idea, and it is desparately needed. I have not had time to test this PR.

Question: does this work for specifying multiple hypervisor keys at once or just one at a time?

i-hate-nicknames:
Just one at a time. I can extend it to a comma-separated list though, should be fairly straightforward.

I'm a little bit worried about similarity between existing `hypervisor` flag and the new hypervisor-pk (possibly -pks for multiple). It looks like these two describe the same entity, while the former is about this node being a hypervisor and the latter about other hypervisors. Maybe it would better to use something like `hypervised-by` or `allowed-hypervisors` or `trusted-hypervisors`? Or we can just use `hypervisors`, to mirror the keys in the config that will be generated. @0pcom @jdknives

jdknives:
@0pcom just out of curiosity, do you know of anyone yet who uses multiple hypervisors for the same visors in practice? 

@i-hate-nicknames maybe `--is-hypervisor` and `-hypervisor-pk(s)` would be clearer?

i-hate-nicknames:
I think `--is-hypervisor` and `--hypervisor-pks` are perfect.

Is there anything tied to the current flag naming though, like build scripts? Technically renaming an existing flag is a breaking change.

jdknives:
@i-hate-nicknames the current `hypervisor` flag only was introduced after the latest release and therefore is only on `develop` so far. So if you cannot find it in our services repo, you can safely rename it.

0pcom:
@jdknives I would do this if not for the configuration difficulty.

Why not have all visor instances in your cluster hooked up to multiple hypervisor instances, accessible from the places where you need to be able to access them. I can see doing this on a computer at work and one at home. Then I can get into my cluster from work more easily or vice versa.

the fact is, having multiple hypervisors are supported by the source code, so it has been at the forefront of my thinking how to support this as a package.

We had earlier talked about having the hypervisor key in a package, which has been my practice, but creating such a package with multiple keys (or multiple ones containing single keys with slight variations in name) is something I still need to figure out; just in terms of having an easier way to collect all the keys or else how to go about naming packages so there aren't name conflicts.

@jdknives will it cause trouble with the production deployment if, for instance, every visor in a skyminer used itself and 7 other visors as hypervisor? thats 64 connections..

jdknives:
@0pcom only have seen this right now. 

It might cause issues in which case we would have to expand the prod deployment. But that would not require a big effort as far as I can tell. I just do not see a point of configuring the cluster like that.

PR #596, opened 2020-11-05
Remove the logout option from the manager if not needed

Senyoret1:

 Changes:	
- The logout button is not shown in the manager if the auth features of the backend are not active.

How to test this PR:
Navigate to the server list in the Skywire manager. The menu shown after clicking the upper right button must not show the logout option if the hypervisor does not have the auth option active.


jdknives (approved):
Good job.

PR #621, opened 2020-11-27
Add update-config command to visor-cli.

### LINKED ISSUE #610: appending hypervisor keys with `skywire-cli visor gen-config`
**Feature description**

The most efficient way to interact with the config .json files is through the action of `skywire-cli visor gen-config` flags and arguments. With that in mind, there are a few predefined actions we can anticipate the user might want to do in terms of hypervisor configuration.

One thing I would like to be able to do is to __either__ append or overwrite the hypervisor key field.

Also, I would ideally like to be able to set a visor as it's own hypervisor as well as specifying a remote visor.

for my immediate purposes, being able to **append a key or keys** and **being able to reset the field** or better documentation as to when to expect this behavior

Making configuration and reconfiguraion easier

for example
```
skywire-cli visor gen-config -r --hypervisor-keys-append 0x0000..
```
the above should add the key to the list of hypervisor keys
the name of the flag is a suggestion and not critical to me

```
skywire-cli visor gen-config -r --hypervisor-keys ""
```
It would be nice if resetting the hypervisor keys field was done like that

```
skywire-cli visor gen-config -r -o /opt/skywire/skywire-config.json
```
So when you do a vanilla regeneration like this it shouldn't reset your hypervisor keys. The `vanilla` mode of configuration generation should update the forma of the config file and if necessary change the default values to reflect whatever changes were made in the source.

The worst thing that can happen is that you just have to regenerate the configuration if something gets messed up, which is not so terrible. So it's not critical that what is there will always be retained between versions, etc. but it should be easy to put it back more or less how it was.

It seems like it wouldn't be too difficult to implement this and it seems like what is stated above is more or less the correct way to go about it. The only immediate alternative is to manually edit the config file.

taras-skycoin:
yes
Fixes #610 

 Changes:	
- Add update-config command to visor-cli.

How to test this PR:
1. Run one of next possible commands:
```
go run cmd/skywire-cli/skywire-cli.go visor update-config --reset
go run cmd/skywire-cli/skywire-cli.go visor update-config -e production
go run cmd/skywire-cli/skywire-cli.go visor update-config -e testing
go run cmd/skywire-cli/skywire-cli.go visor update-config --hypervisor-pks=023d1dc11119a7e2d2207f6c9e94ee88a5801702616ca1ff72bc3b2af785b7c1b8,023d1dc11119a7e2d2207f6c9e94ee88a5801702616ca1ff72bc3b2af785b7c1b8
```

### DIFF
--- a/cmd/skywire-cli/commands/visor/update-config.go
+++ b/cmd/skywire-cli/commands/visor/update-config.go
@@ -32,8 +32,8 @@ func init() {
 	updateConfigCmd.Flags().StringVarP(&addOutput, "output", "o", "skywire-config.json", "path of output config file.")
 	updateConfigCmd.Flags().StringVarP(&addInput, "input", "i", "skywire-config.json", "path of input config file.")
 	updateConfigCmd.Flags().StringVarP(&environment, "environment", "e", "production", "desired environment (values production or testing)")
-	updateConfigCmd.Flags().StringVar(&addHypervisorPKs, "hypervisor-pks", "", "public keys of hypervisors that should be added to this visor")
-	updateConfigCmd.Flags().BoolVar(&resetHypervisor, "reset", false, "resets hypervisor`s configuration")
+	updateConfigCmd.Flags().StringVar(&addHypervisorPKs, "add-hypervisor-pks", "", "public keys of hypervisors that should be added to this visor")
+	updateConfigCmd.Flags().BoolVar(&resetHypervisor, "reset-hypervisor-pks", false, "resets hypervisor`s configuration")
 }
 
 var updateConfigCmd = &cobra.Command{
--- a/cmd/skywire-cli/commands/visor/update-config.go
+++ b/cmd/skywire-cli/commands/visor/update-config.go
@@ -13,8 +13,6 @@ import (
 	"github.com/skycoin/skycoin/src/util/logging"
 	"github.com/spf13/cobra"
 
-	"github.com/skycoin/skywire/pkg/skyenv"
-
 	"github.com/skycoin/skywire/pkg/visor/visorconfig"
 )
 
@@ -35,7 +33,7 @@ func init() {
 	updateConfigCmd.Flags().StringVarP(&addInput, "input", "i", "skywire-config.json", "path of input config file.")
 	updateConfigCmd.Flags().StringVarP(&environment, "environment", "e", "production", "desired environment (values production or testing)")
 	updateConfigCmd.Flags().StringVar(&addHypervisorPKs, "hypervisor-pks", "", "public keys of hypervisors that should be added to this visor")
-	updateConfigCmd.Flags().BoolVarP(&resetHypervisor, "reset", "r", false, "resets hypervisor`s configuration")
+	updateConfigCmd.Flags().BoolVar(&resetHypervisor, "reset", false, "resets hypervisor`s configuration")
 }
 
 var updateConfigCmd = &cobra.Command{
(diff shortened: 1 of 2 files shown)


cmd/skywire-cli/commands/visor/update-config.go
```diff
@@ -0,0 +1,115 @@
...
+			mLog.WithError(err).
+				WithField("filepath", addInput).
+				Fatal("Failed to read config file.")
+		}
+
+		raw, err := ioutil.ReadAll(f)
+		if err != nil {
+			mLog.WithError(err).Fatal("Failed to read config.")
+		}
+		conf, ok := visorconfig.Parse(mLog, addInput, raw)
```
Darkren, line 63: Could you please add an empty line before this one?

cmd/skywire-cli/commands/visor/update-config.go
```diff
@@ -0,0 +1,115 @@
...
+		if ok != nil {
+			mLog.WithError(err).Fatal("Failed to parse config.")
+
```
Darkren: Could you please remove this empty line?

cmd/skywire-cli/commands/visor/update-config.go
```diff
@@ -0,0 +1,115 @@
...
+		if addHypervisorPKs != "" {
+			keys := strings.Split(addHypervisorPKs, ",")
+			for _, key := range keys {
+				keyParsed, err := coinCipher.PubKeyFromHex(strings.TrimSpace(key))
+				if err != nil {
+					logger.WithError(err).Fatalf("Failed to parse hypervisor private key: %s.", key)
+				}
+				conf.Hypervisors = append(conf.Hypervisors, cipher.PubKey(keyParsed))
+			}
+
```
Darkren: And this empty line

cmd/skywire-cli/commands/visor/update-config.go
```diff
@@ -0,0 +1,115 @@
...
+		}
+		if environment == "production" {
```
Darkren, line 79: Could you please add an empty line before this one?

cmd/skywire-cli/commands/visor/update-config.go
```diff
@@ -0,0 +1,115 @@
...
+		if environment == "testing" {
+			conf.Dmsg.Discovery = skyenv.TestDmsgDiscAddr
+			conf.Transport.Discovery = skyenv.TestTpDiscAddr
+			conf.Transport.AddressResolver = skyenv.TestAddressResolverAddr
+			conf.Routing.RouteFinder = skyenv.TestRouteFinderAddr
+			conf.UptimeTracker.Addr = skyenv.TestUptimeTrackerAddr
+			conf.Launcher.Discovery.ServiceDisc = skyenv.TestServiceDiscAddr
+
+		}
+		if resetHypervisor {
```
Darkren, line 87: Could you please add an empty line before this one?

cmd/skywire-cli/commands/visor/update-config.go
```diff
@@ -0,0 +1,115 @@
...
+			conf.Hypervisors = []cipher.PubKey{}
+		}
+		// Save config to file.
```
Darkren, line 91: Could you please add an empty line before this one?

Darkren (changes requested):
Great job! I'm mostly concerned about formatting. Except one thing. Lines 82-88 and 92-97, here we store constant values to config fields. And we also do something similar while generating config. If we need to introduce a new field, we may miss it here or there. What do you think about moving these parts to separate funcs and calling these here and in the generation routines?

i-hate-nicknames:
Great job!

I would make one file to be a target, instead of input/output, in my opinion update-config implies that there is something to which we perform the update action and the result stays there. On the other hand, having a default value for both makes it relatively easy to use as it is. 

Also, to make it more consistent with get-config command I would abstain from using -r flag, since it has different meaning in gen-config.  Instead, I would use something like `--append-hypervisor-pks` to add new PKs to existing hypervisors field, and use `--hypervisor-pks` to simply replace current list.

What do you think?

taras-skycoin:
@i-hate-nicknames  @Darkren Thank you, guys. I have added suggested changes. Please review it again when you have time.

jdknives (approved):

PR #626, opened 2020-11-29
Mark visors with problematic services in the visor list

Senyoret1:

 Changes:	
- Now, if one or more services of a visor is not working, the dot shown at the left of the visor in the visor list is shown yellow and blinking, to make it easier to know when a visor has a problem.
- The code for checking the state of the visor services was changed, to make it centralized and use the same code in the visor details page and the visor list. This was made to make the code more maintainable and to avoid potential errors in the future, if changes are made only in one page.

How to test this PR:
Make one of the visors have trouble connecting with a service (like the uptime tracker or the transport discovery). The affected visor must have a yellow dot at the left in the visor list, instead of the green dot used for indicating that the visor is working well. Also, the “health info” section of the visor details page should continue working well.

### DIFF
--- a/static/skywire-manager-src/src/app/app.datatypes.ts
+++ b/static/skywire-manager-src/src/app/app.datatypes.ts
@@ -53,6 +53,8 @@ export interface HealthInfo {
   transport_discovery?: number;
   route_finder?: number;
   setup_node?: number;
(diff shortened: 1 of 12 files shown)


jdknives:
Nice effect with the pulsating circle. Well done!

PR #630, opened 2020-12-02
Shared app state

### LINKED ISSUE #606: Share app state with visor
**Feature description**

Some of the Skywire applications have quite complex logic and state that needs to be displayed to the user. The VPN client for instance needs to perform handshakes and setup local routes etc to run successfully. Simply displaying to the user whether the route used by VPN is up or down is not enough to inform users about current state of the application. We need to think about how to implement either a ad-hoc solution for VPN client specifically or something more generic and versatile.

The VPN client is probably the most relevant case for this feature as we need to report accurately whether the VPN is working and running correctly at any given moment.

Darkren:
Yes
Fixes #606 

 Changes:	
- Additional methods were added to app connection RPC client/server to set detailed app status which is a custom app-specific string provided by the app itself;
- Detailed app status is now returned along with all other info via `/visors/{pk}/apps`;
- Specific statuses were implemented for the VPN client: `Connecting`, `Running`, `ShuttingDown` and `Reconnecting`.

How to test this PR:
Run VPN client and query endpoint specified above to check status change


taras-skycoin (approved):
Great job! 
Status field works.

Issue #614, opened 2020-11-22
The need for Versioned Dependencies or a Static Compilation method (musl)


0pcom:
**Describe the bug**
In recent efforts to create .deb skywire packages on archlinux which reflect the recent changes in the develop branch, I ran into the following error when trying to run the compiled binaries on armbian

```
skywire-cli: /lib/aarch64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by skywire-cli)
```
I did not realise that glibc was a dependancy.

**Environment information:**

- OS: Armbian and ArchlinuxARM
- Platform:
Linux 5.8.16-sunxi64 aarch64 (armbian)
Linux 3.10.65+ aarch64 (archARM)

**Steps to Reproduce**

Compile the skycoin binaries with a newer version of glibc tan exists on the target system where the cmpiled binaries are installed. 

**Actual behavior**
The error :
```
skywire-cli: /lib/aarch64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by skywire-cli)
```

**Expected behavior**
skywire runs successfully

**Additional context**
This breaks my automated build, and so far alternatives have proven untenable.

Is there a list of the runtime dependancies that I might examine?

0pcom:
**possible implementation**

https://dominik.honnef.co/posts/2015/06/statically_compiled_go_programs__always__even_with_cgo__using_musl/

https://www.google.com/amp/s/ops.tips/amp/notes/glibc-golang-learnings/

0pcom:
This is fixed now if the PR gets approved.

jdknives:
Fixed with #634

PR #634, opened 2020-12-08
static compilation w/musl

0pcom:
Static compilation with musl per haltingstate's reuest

I couldn't figure out how to escape the quotation marks in the makefile so i just eliminated that env and put it inline in the go build command.

Also, for deterministic builds you really want to be in the directory where you're building the .go file into a binary so that the path isn't written into the binary.

no changes to go source were made

Fixes #
https://github.com/skycoin/skywire/issues/614

 Changes:	
- static compilation w/musl
- remove -ldflags which interfere with deterministic builds
- update documentation about musl
- update documentation about hypervisor & visor configuration
- remove dPKGBUILD.sh and associated makefile entries as I am not maintaining this method for building packages.

How to test this PR:
`make build`

Additional context:
My current debian build uses dependencies only available on archlinux but it works much better than that script. As I am actively developing .deb packages in https://github.com/the-skycoin-project/deb I have no intention of maintaining the more rudimentary method which avoids the arch deps.


jdknives (changes requested):
This means we cannot build on Mac anymore. I tried setting up `musl` on Mac and it is not as straightforward as I hoped. Also it does not come with most distributions by default so users could not build straight away on those systems if they wanted. Will check with @SkycoinSynth about whether he still wants us to go ahead with this nonetheless.

PR #635, opened 2020-12-09
Fix app startup issues after changing the args via UI

### LINKED ISSUE #627: Problems starting and stopping the VPN apps
**Describe the bug**
Making some tests I found 2 different errors when starting and stopping the VPN apps:
- If you configure the VPN client app to connect to a visor which does not have the VPN server app and start it, the VPN client app will only be started for a small moment and then the API will start returning `status: 0`, indicating that the app was stopped. The problem is that when trying to start the VPN client app again the API returns `app already started` and error 500, even if the app is reconfigured to use the pk of a valid server. The problem continues until the visor is restarted.
- If you have both the VPN client and the VPN server apps in the same visor and start the VPN client app for connecting to the VPN server app in the same visor, it will not be possible to stop the VPN client app, unless the visor is restarted. Also, after the error, the manager is not able to make any operation related to the affected visor, because the API starts returning `connection is shut down`.

**Environment information:**

- OS: Linux (Ubuntu 18.04.1)

**Steps to Reproduce**
Steps to reproduce the first problem:
- Start a visor, including the VPN client app.
- Use the Skywire manager to change the server pk configured in the VPN client app to 024ec47420176680816e0406250e7156465e4531f5b26057c9f6297bb0303558c7.
- Press the start button in the Skywire manager, to start the VPN client app.
- Refresh the info in the Skywire manager until the VPN client app appears stopped.
- Press the start button in the Skywire manager, to start the VPN client app again.
- The manager should show `app already started`. If you use the API to check the state of the app, is will return `status: 0`, which should mean that the app is stopped.
- If you refresh the data continually, the app may appear started at some moments, but the behavior is erratic.

Steps to reproduce the second problem:
- Start a visor with the VPN client and VPN server apps.
- Use the Skywire manager to change the server pk configured in the VPN client app to the pk of the visor itself.
- Make sure the VPN server app is running.
- Press the start button in the Skywire manager, to start the VPN client app.
- Wait until the app appears running.
- Press the stop button of the VPN client app and confirm the operation in the modal window.
- After a moment, the modal window should show the `Context deadline exceeded.` error message.
- After the error, the manager will not be able to continue accessing the visor, as the API starts returning `connection is shut down`. The problem is solved restarting the visor.

**Actual behavior**
Both errors force the user to restart the visor to be able to continue using it normally.

Darkren:
Yes
Fixes #627 

 Changes:	
- Args to the app binary are now taken from the config, not from the previous app start
- App proc is now correctly removed from the proc manager on app exit, no matter what was the exit reason

How to test this PR:
To test that hypervisor restarts app with the changed args:
1. Start any app via UI, VPN client for example
2. Change PK via UI
3. VPN client should be restarted and connected to the new remote server

To test the original issue:
1. Modify `./internal/vpn/client.go`: put this line `return errors.New("fail")` to the line around 183rd, right after the `dialServeConn` and checking its error. This will make the app exit right after connection failure
2. Build binaries
3. Modify VPN client config to connect to the remote visor which doesn't have VPN server 
4. Run VPN client from the UI. After a while it will be displayed as being off again, wait for it
5. Change the remote PK to the visor which has VPN enabled. It should change PK okay
6. Run VPN client from the UI, you should be using VPN now okay

### DIFF
--- a/internal/vpn/client.go
+++ b/internal/vpn/client.go
@@ -181,8 +181,6 @@ func (c *Client) Serve() error {
 			fmt.Printf("dialServeConn: %v\n", err)
 		}
 
-		return errors.New("fail")
-
 		if c.isClosed() {
 			return nil
 		}
--- a/internal/vpn/client.go
+++ b/internal/vpn/client.go
@@ -181,6 +181,8 @@ func (c *Client) Serve() error {
 			fmt.Printf("dialServeConn: %v\n", err)
 		}
 
+		return errors.New("fail")
+
 		if c.isClosed() {
 			return nil
 		}
--- a/pkg/app/appserver/proc.go
+++ b/pkg/app/appserver/proc.go
@@ -58,6 +58,8 @@ func NewProc(mLog *logging.MasterLogger, conf appcommon.ProcConfig, disc appdisc
 	}
 	moduleName := fmt.Sprintf("proc:%s:%s", conf.AppName, conf.ProcKey)
 
+	fmt.Printf("CONF PROC ARGS: %v\n", conf.ProcArgs)
+
 	cmd := exec.Command(conf.BinaryLoc, conf.ProcArgs...) // nolint:gosec
 	cmd.Dir = conf.ProcWorkDir
 	cmd.Env = append(os.Environ(), conf.Envs()...)
--- a/pkg/app/appserver/proc.go
+++ b/pkg/app/appserver/proc.go
@@ -162,6 +162,13 @@ func (p *Proc) Start() error {
 			close(waitErrCh)
 		}()
 
+		defer func() {
+			// here will definitely be an error notifying that the process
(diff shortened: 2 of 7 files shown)


taras-skycoin (approved):
Good job! 

Testing is a bit tricky since you need to switch between two working VPN servers. Please consider it if you are going to test it. You can use these PKs: 
```
036961efd1bdd926e018755ad9cb52f3847191ca776efe412522ffc2781ad6227c
029081ab2f1a9087f7e59b3b58ca2959530335f861725d31aac1b5a3aff150f832
```

jdknives:
@taras-skycoin you tested it? If so we can merge this.

taras-skycoin:
Yes, I tested it and left a suggestion for another person who is going to test it too.

jdknives:
@taras-skycoin Will merge it then as I tested it already.

PR #638, opened 2020-12-11
Fix/dup syscall portability

i-hate-nicknames:
Yes
Fixes:
- Release build for arm64

Changes:	

How to test this PR:
make test

dup2 syscall is absent on arm64 architecture, while dup3 is absent on darwin OS. This implements common function that has build-specific implementations


jdknives (approved):
Works fine on MacOS. Will test later on ARM.

Issue #636, opened 2020-12-09
Route redialing continues after app is closed


jdknives:
**Describe the bug**
After requesting a route and closing the app that created a route, the visor keeps on retrying to redial for the old route. This prevents to dial for a new route for a while for me. 

**Environment information:**

- OS:Mac OS X

- happens on `fix/vpn-start-stop`

**Steps to Reproduce**

1.Setup VPN client with remote PK
2. Start app
3. Stop app
4. Check logs
5. Start app with new PK set

**Actual behavior**
Visor keeps redialing for quite some time

**Expected behavior**
The visor stops when app requesting routes is closed or just does not try as often.

**Possible implementation**
Simplest fix would be to modify route redialing logic to only retry 4-5 times.

jdknives:
This issue is actually not an issue with the redialing logic but the shutdown logic. The VPN is still running even though the shutdown signal was received by visor. Attaching logs, which show that the HTTP request only completes after 52s when I manually kill the VPN process. 

[log-vpn-client.txt](https://github.com/skycoin/skywire/files/5695123/log-vpn-client.txt)

PR #642, opened 2020-12-15
Fix VPN client duplicates

### LINKED ISSUE #641: App startup/shutdown issues
**Describe the bug**
I just encountered a scenario, where I was able to startup two instances of the VPN client. Both instances were continuing to run after the visor shut down. 

```
root               978   0,0  0,1  5016468   9200 s000  S+    2:10pm   0:00.04 /Users/work/go/src/github.com/skycoin/skywire/apps/vpn-client -srv 03d65df7e74a480ab645ade9ae45ec6280c9a86fef2a9e955d99361c9a678b61ee -killswitch true
root               821   0,0  0,1  5015584   7752 s000  S+    2:08pm   0:00.09 /Users/work/go/src/github.com/skycoin/skywire/apps/vpn-client -passcode  -srv 020cee8278305d90b56b6e86d5e5d718929d147c76de4bdb5b83545273320edd47
```

The error popped up after setting up the VPN client with a PK that it could not connect to. After changing app args (for instance toggling killswitch arg), the popup window of the VPN client setting would not show completed operation. Upon closing that window manually, one can start the app again. 

**Environment information:**

- OS: Mac OS X

**Steps to Reproduce**

1. Startup hypervisor
2. Run VPN client with remote it cannot connect to but is a valid PK
3. Change an arg of the VPN client (should likely not complete)
4. Exit the pop up window
4. See the icon is red for VPN client and restart

**Actual behavior**
Starts two instances of the VPN client

**Expected behavior**
Should cleanly kill VPN client instance before restarting. 

**Possible implementation**
Likely related to https://github.com/skycoin/skywire/pull/637 or https://github.com/skycoin/skywire/pull/635

The latter PR changes stopping logic slightly, so this may be related to this bug.

Darkren:
Yes
Fixes #641 #636 

 Changes:	
- Now VPN client exits gracefully during dialing. Previously dialing couldn't be interrupted. So, when we changed anything or tried to stop the app, the request just hung waiting for the process to stop. And this waiting eventually gets interrupted with an error. In this case we just decouple process from the visor due to the internal architecture, therefore allowing app duplicates. However, this shouldn't be a problem for a well-written (in terms of interruption) apps

How to test this PR:
Repeat steps from the original ticket. Step 3 should complete and app should be restarted. Also VPN client should be closed on visor shut down without any duplicates. Also, previously issue could be reproduced not by changing the app arg, but just stopping it and refreshing page afterwards (when stop hung). This should work too, app should be stopped ok during dialing

### DIFF
--- a/internal/vpn/client.go
+++ b/internal/vpn/client.go
@@ -667,10 +667,23 @@ func (c *Client) dialServer(appCl *app.Client, pk cipher.PubKey) (net.Conn, erro
 			PubKey: pk,
 			Port:   vpnPort,
 		})
+
+		if c.isClosed() {
+			// in this case client got closed, we return no error,
+			// so that retrier could stop gracefully
+			return nil
+		}
+
 		return err
 	})
 	if err != nil {
 		return nil, err
+	}
+
+	if c.isClosed() {
+		// we need to signal outer code that connection object is inalid
+		// in this case
+		return nil, errors.New("client got closed")
 	}
 
 	return conn, nil


jdknives (approved):
Works. Nice job.

i-hate-nicknames:
Great job, works for me too.

PR #645, opened 2020-12-18
Show more info about the routes in the manager

### LINKED ISSUE #429: Add UI/UX improvements to the manager

Senyoret1:

Fixes #429

 Changes:	
- The rule string was removed from the routes list. Now it shows the route type, source and destination.
rt
- The route details popup now shows the labels associated with the PKs and IDs, if any.

How to test this PR:
Check the route list of a visor using the Skywire manager.

### DIFF
--- a/static/skywire-manager-src/src/app/components/pages/node/routing/route-list/route-details/route-details.component.html
+++ b/static/skywire-manager-src/src/app/components/pages/node/routing/route-list/route-details/route-details.component.html
@@ -54,6 +54,11 @@
               routeRule.forwardFields.nextTid :
               routeRule.intermediaryForwardFields.nextTid
           }}
+          {{
+            getLabel(routeRule.forwardFields ?
+            routeRule.forwardFields.nextTid :
+            routeRule.intermediaryForwardFields.nextTid)
+          }}
         </div>
       </div>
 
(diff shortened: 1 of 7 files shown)


jdknives (approved):
Works fine.

PR #649, opened 2020-12-22
Additional app stats

### LINKED ISSUE #643: Make it possible to get more stats from the VPN client app
We should have a way for getting the following data from the VPN client app (this is needed for the UI):

- The data transmission stats (upload/download speed and total data uploaded/downloaded)
- The latency.
- For how long the app has been running.

Maybe the changes made in https://github.com/skycoin/skywire/pull/630 can be used as the foundation for this.

Darkren:
Yes
Fixes #643 

 Changes:	
- `/visors/{pk}/apps/{app}/stats` endpoint was added. It returns general app stats including its start time in UTC, and connections summary;
- Connections summary now has separate upload/download speeds instead of single throughput;
- Connections summary now has `bandwidth_received` field.


taras-skycoin (approved):
Good job!

Senyoret1:
Not sure about the reason, but I’m having 2 problems:

- Callíng `GET /visors/{pk}/apps/vpn-client/stats` for getting data about the VPN client app is always returning this to me:

```
{
    "connections": null
}
```

- In VPN client UI I’m calling the `GET visors/{pk}/apps/vpn-client/connections` every 2 seconds and the behavior appears to be a bit strange. At first, the total uploaded data starts climbing until reaching 558, but the upload speed stays at 0 while that happens. After a moment, the upload speed goes to about 367 for a moment, but the total uploaded data number stays the same. After some time, the upload speed may go up again from 0, but the total uploaded data stays the same. At the same time, the download speeda may go up and down, but the total downloaded data atays at 0.

No sure if the behavior is due to my dev enviroment. For testing, I used `skywire-services` and then connected 2 additional visors to it (visor 1 and visor 2) with all the Skywire apps installed. What I do is to configure the VPN client app in visor 1 to connect to the VPN Server app of visor 2. Maybe having the 2 apps in the same machina is causing the problems.

Appart from that, just a quick question: is the data returned by the `visors/{pk}/apps/vpn-client/connections` endpoint in bits or bytes?

Darkren:
@Senyoret1 to be fair, can't say for sure. tested it, works for me just fine. but one thing is - you shouldn't setup both apps on the same machine, it's not intended to work and behavior is unknown to me. so, please, try moving the server to some remote machine. 

and regarding your question - data is in bytes

Senyoret1:
I checked using 2 different virtual machines, one for the client and the other for the server, using the current develop branch and it worked well. There was just a problem, at least in my tests: the `/api/visors/{pk}/apps/vpn-client/connections` API endpoint returned almost all fields with data that appears to be correct, except for the `bandwidth_received` field, witch always had 0. @Darkren could you please check? I'm still not sure if it is a problem with my setup or something else.

PR #652, opened 2021-01-05
Make the UI show the last version of the selected channel

Senyoret1:

 Changes:	
- Before this PR, if a custom channel for getting updates was set in the settings page, the manager used it for downloading the update, but the UI always showed the last version of the `stable` channel as the one that was going to be installed, which was obviously wrong if the user selected other channel. This PR solves that problem.

How to test this PR:
Set a custom channel for getting updates, using the form in the settings page, and try to update a visor. The updater modal window should show the last version of the selected channel as the one that will be installed and the dev tools should show a call to `/api/visors/{pk}/update/available/{channel}` instead of `/api/visors/{pk}/update/available`.


jdknives (approved):
Good job.

PR #655, opened 2021-01-13
Improvements for the manager

Senyoret1:

 Changes:	
- Improvements were made to how the visor list data is managed, to avoid unnecessary loading times.
- A new start page, for redirecting the user to the real appropriate initial page and avoid some weird behaviors, was added.
- A new delayed data refresh procedure was added to the app list, because some changes may take more time than expedted and that made the app list to remain outdated after making some changes.
- Some calls to data clearing procedures the were not being called were added.

How to test this PR:
- When entering to the visor details page and returning to the visor list, the visor list should be displayed inmediatelly, without the big loading animation. This happened mainly when the app settings were changad to update the data automatically every 3 seconds.
- If the auth options are disabled, opening the root URL of the app should redirect to the visor list page, instead of opening the loging page and then be redirected to the visor list.
- Alter starting or stopping an app, its state should be updated in the app list. The faulty cases were not very easy to reproduce, but some times starting the VPN server app made the problem to appear.

### DIFF
--- a/static/skywire-manager-src/src/app/app-routing.module.ts
+++ b/static/skywire-manager-src/src/app/app-routing.module.ts
@@ -1,6 +1,7 @@
 import { NgModule } from '@angular/core';
 import { RouterModule, Routes } from '@angular/router';
 
+import { StartComponent } from './components/pages/start/start.component';
 import { LoginComponent } from './components/pages/login/login.component';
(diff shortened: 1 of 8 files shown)


jdknives:
Works very well. Good job!

PR #656, opened 2021-01-14
Feature/add min hops config

### LINKED ISSUE #319: Allow to set min hops for route requests
**Feature description**

We support setting a min hop parameter when requesting routes so that users can setup routes that go through an intermediary before going to the destination visor. It would be good to allow to set that value for all route requests in the config.

i-hate-nicknames:

Fixes #319

 Changes:	
- Add min_hops visor router config option

How to test this PR:

### DIFF
--- a/pkg/router/router.go
+++ b/pkg/router/router.go
@@ -61,6 +61,8 @@ type Config struct {
 	RouteGroupDialer setupclient.RouteGroupDialer
 	SetupNodes       []cipher.PubKey
 	RulesGCInterval  time.Duration
+	MinHops          uint16
+	MaxHops          uint16
 }
 
 // SetDefaults sets default values for certain empty values.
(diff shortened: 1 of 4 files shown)


Darkren (approved):
Great job!

jdknives:
@i-hate-nicknames since this is changing the config of a visor, we need to think about whether we need to add this to the config migration and whether we need to bump the config version.

i-hate-nicknames:
@jdknives I'm not sure if migration is needed, and what is the procedure. I need to research it.

As of now, I can't think of issues that can be introduced by this change. 

If new code reads old config format, the values are not there and config struct gets zero values. Zero value of minHops is the default value, so no problem there. Zero value of maxHops is a problem, but it is reset to default value (50) [here](https://github.com/skycoin/skywire/pull/656/files#diff-06404baf0ccd8477ead1249463f8f394030eb63c6c3bcff3fe1fea2368150673R83) for such case.

If old code reads new config file, the values will be ignored (no change in the structure is made, only new fields are added).

PR #663, opened 2021-01-22
Fix/check empty password

### LINKED ISSUE #660: Inconsistent behavior when manipulating app args
**Describe the bug**

When setting the secure mode for the VPN-server via HypervisorUI and without setting a password, the hypervisor UI adds these values

```
		"name": "vpn-server",
				"args": [
					"-passcode",
					"",
					"-secure",
					"true"
				],
```

This leads to the VPN-server not registering itself in the service discovery because the check we perform is not sufficient. This needs to be fixed. 

When unchecking the box of secure mode again, this is the updated config value: 

```
	"name": "vpn-server",
				"args": [
					"-secure",
					"false"
				],
```

This is probably not the best way to deal with it. Imo, it sounds cleaner to generate the default values at config generation and always update all of the values when manipulating at least one of them via the hypervisor. The check for whether the VPN is password protected needs to be improved to actually check whether the password is non-empty instead of just checking if the arg is present.

jdknives:

Yes

Fixes #660 at least the bug. The inconsistency remains. 

 Changes:	
- adds check to verify if password is empty. If password is empty, the appupdater is being created. Previously empty string password would lead to appdiscupdater not being created which is not the correct behavior. 

How to test this PR:

- run VPN-server with passcode arg set to empty string against prod deployment
- verify the VPN server gets registered here: https://service.discovery.skycoin.com/api/services?type=vpn

### DIFF
--- a/pkg/app/appdisc/factory.go
+++ b/pkg/app/appdisc/factory.go
@@ -127,7 +127,7 @@ func argEqualsFlag(arg, flag string) bool {
 
 func argVal(args []string, flag string) string {
 	for idx, arg := range args {
-		if argEqualsFlag(arg, flag) {
+		if argEqualsFlag(arg, flag) && idx+1 < len(args) {
 			return args[idx+1]
 		}
 	}
--- a/pkg/app/appdisc/factory.go
+++ b/pkg/app/appdisc/factory.go
@@ -65,7 +65,7 @@ func (f *Factory) AppUpdater(conf appcommon.ProcConfig) (Updater, bool) {
 	log := f.Log.WithField("appName", conf.AppName)
 
 	// Do not update in proxy discovery if passcode-protected.
-	if containsFlag(conf.ProcArgs, "passcode") {
+	if containsFlag(conf.ProcArgs, "passcode") && argVal(conf.ProcArgs, "passcode") != "" {
 		return &emptyUpdater{}, false
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/app/appdisc/factory.go
```diff
@@ -124,3 +124,13 @@ func argEqualsFlag(arg, flag string) bool {
 
 	return arg == flag
 }
+
+func argVal(args []string, flag string) string {
+	for idx, arg := range args {
+		if argEqualsFlag(arg, flag) {
+			return args[idx+1]
```
Darkren: I'm not sure, but if we pass `--passcode` as the last flag without value, wouldn't it panic with the `index out of bounds`? probably it's worth checking in here that `id < len(args)`. otherwise the value should be considered empty

Darkren (changes requested):
Great job! But I'm afraid it may panic. Not sure though

pkg/app/appdisc/factory.go
jdknives: Added check.

i-hate-nicknames (commented):
Great job! I would move argVal function to pkg/app/appcommon/proc_config.go, since it's a generic arg operation

jdknives:
> Great job! I would move argVal function to pkg/app/appcommon/proc_config.go, since it's a generic arg operation

Probably a good idea. will do that.

Darkren (approved):

PR #666, opened 2021-01-26
Feature/extra summary

### LINKED ISSUE #646: Include more info in the API endpoint for getting the visor list
Currently, for populating the visor list, the Skywire manager has make the following API calls:

- `GET api/visors`: for getting the list.
- `GET api/dmsg`: for knowing the dmsg server to which each visor is connected.
-  `GET api/visors/{pk}/health` (one call for each visor): for knowing if the visor has a problem with a service.
- `GET api/about`: for knowing which visor is working as hypervisor.

It would be better if all data could be obtained just by one API call, because having to make several calls make the UI much slower. This could work like the `GET api/visors/{pk}/summary` API endpoint.

Note: this is not prioritary.

i-hate-nicknames:
Yes

Fixes #646 

 Changes:
- Add `GET /api//visors-summary` endpoint, allowing to fetch detailed information about each visor

How to test this PR:
Run a hypervisor with some visors and query `GET /api//visors-summary`.


Darkren (approved):

pkg/visor/hypervisor.go
```diff
@@ -489,6 +490,99 @@ func (hv *Hypervisor) getVisorSummary() http.HandlerFunc {
...
+}
+
+func (hv *Hypervisor) getVisorsExtraSummary() http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		hv.mu.RLock()
+		wg := new(sync.WaitGroup)
+		wg.Add(len(hv.visors))
+
+		i := 0
+		if hv.visor != nil {
```
jdknives: Is there currently any possiblity of this code executing when `hv.Visor` would be nil?

i-hate-nicknames: Indeed there isn't. I took the code for getting visors and modified it to have additional info in response. However, we never start hypervisor without visor, except in tests. However, some tests rely on nil response on an "empty" hypervisor. Not sure if it's worth the trouble to change the tests and remove nil-checks everywhere.

Senyoret1:
This is working well. I would just remove the `dmsg` field, because it appears to always return `null` and the data needed is in the `dmsg_stats` field, and the `routes` field, because it is not needed for the visor list and I think it could sometimes contain a lot of data, which would just unnecessarily degrade the performance. However, if there are reasons for keeping them there, then there is no problem

i-hate-nicknames:
@Senyoret1 no other reasons besides just composing existing structures instead of declaring new ones. If you don't need that info it's no problem to avoid sending it.

Senyoret1:
I checked and https://github.com/skycoin/skywire/pull/699 contines working after the changes.

PR #671, opened 2021-01-29
Parallelize health requests

### LINKED ISSUE #670: Fix /health timeouts
**Describe the bug**
Currently we get a lot of `/health` timeouts from the hypervisor because the performance of the request on the visor is quite bad. Even if we call the endpoint locally, we frequently get timeouts. We should improve the performance of the call. Likely we just need to process individual calls in parallel. Maybe there are other issues that decrease performance.

i-hate-nicknames:

Fixes #670

 Changes:	
- Improve performance of `/api/visors/{pk}/health`

How to test this PR:
Run a hypervisor and hit `GET /api/visors/{pk}/health` for a pk that runs under that hypervisor

### DIFF
--- a/pkg/visor/api.go
+++ b/pkg/visor/api.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"os/exec"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/google/uuid"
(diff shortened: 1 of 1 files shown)


jdknives (approved):
Good work. The performance improvement is really felt in the UI.

Darkren (approved):
Great job!

PR #672, opened 2021-02-01
Public visors advertising

### LINKED ISSUE #574: Allow visors to advertise themselves as running on Public IP
**Feature description**

We should make it easier for users to establish transports automatically. The best scenario would be to make them establish `stcpr` transports to visors on public IPs. We should allow visors on public IPs to register themselves in the service discovery so other visors can establish transports to them upon startup. 

Currently, if visors want to open up a route to another visor, they need to establish transports manually. 

I would suggest we go with a semi-manual way of solving this issue by allowing operators of visors to set a config value that indicates their visor is reachable for `stcpr` transports. When this is set, the visor registers itself in the service discovery. Otherwise we would have to automate the detection of whether visor is reachable for `stcpr`. In any case, we should verify in the service disc whether entries are valid.

Darkren:

Fixes #574 	

 Changes:	
- On start visor registers itself in the service discovery. Visor attaches IPs of its local active network interface to the request;
- If `connect_to_public_visors` field is set to `true` in config, visor will try to connect to 5 known visors on its start.


Darkren:
@jdknives This one is blocked by Traefik2, need it for this to work, just a reminder

PR #677, opened 2021-02-04
Fix hanging health

### LINKED ISSUE #674: /health timeouts when hypervisor board runs VPN-client
**Describe the bug**
On the last pre-release version of Skybian (which uses the pre-release v0.4.0-rc8 in turn), running a VPN client on the hypervisor board makes the visor go offline in the UI permanently. 
More specifically it actually seems to stay green in the UI but throws a 408 timeout error on health requests when the details page is opened. Previously we had similar issues on boards that only ran a regular visor but these timeouts have been fixed by performance improvements in #670 . 

**Environment information:**

- OS: Skybian Skywire v0.2.0-rc4

**Steps to Reproduce**

1. Run VPN client on hypervisor machine. 
2. Check visor details page in UI. 

**Comments**

I have a feeling this could be related to the fact that the connection established between a visor and hypervisor are different if the visor is on the same board vs when it is on a different machine. Maybe the VPN client interferes and breaks this connectino somehow.

Darkren:

Fixes #674 

 Changes:	
- Uptime Tracker IP is now properly passed to the VPN client. So client now can establish direct connection to it.

How to test this PR:
1. Run visor with the hypervisor on the same machine.
2. Run VPN client.
3. Try accessing current visor in the list. Page should open with no timeout.

### DIFF
--- a/internal/vpn/client.go
+++ b/internal/vpn/client.go
@@ -80,7 +80,7 @@ func NewClient(cfg ClientConfig, appCl *app.Client) (*Client, error) {
 
 	utIP, err := uptimeTrackerIPFromEnv()
 	if err != nil {
-		return nil, fmt.Errorf("error getting UT IP: %w")
+		return nil, fmt.Errorf("error getting UT IP: %w", err)
 	}
 
 	stcpEntities, err := stcpEntitiesFromEnv()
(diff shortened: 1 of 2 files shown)


jdknives (approved):
Looks good. Will test and then merge.

PR #683, opened 2021-02-11
Add separate timeout for /health requests done by visor

### LINKED ISSUE #676: /health endpoint times out consistently if call to service times out
**Describe the bug**
With the current setup of the `/health` endpoint, the whole call times out if an individual call of the visor to one of the services times out. This is caused by the timeout of the individual call being higher (or close to) the timeout of the overall call. This should not happen. The service should just be shown as unavailable if the call could not complete within a smaller timeout period. 

**Steps to Reproduce**

See reproduction steps in #674 

Because of a bug in the VPN setup itself, the calls to the uptime tracker could not complete.

Darkren:

Fixes #676 

 Changes:	
- Now `/health` requests done by the visor have separate timeout which is lower than for the hypervisor's `/health` request.

### DIFF
--- a/pkg/visor/rpc.go
+++ b/pkg/visor/rpc.go
@@ -22,8 +22,12 @@ import (
 const (
 	// RPCPrefix is the prefix used with all RPC calls.
 	RPCPrefix = "app-visor"
-	// HealthTimeout defines timeout for /health endpoint calls.
+	// HealthTimeout defines timeout for /health endpoint calls done from hypervisor.
 	HealthTimeout = 5 * time.Second
+	// InnerHealthTimeout defines timeout for /health endpoint calls done from visor.
+	// We keep it is less than the `HealthTimeout`, so that the outer call would
+	// definitely complete.
+	InnerHealthTimeout = 4 * time.Second
 )
 
 var (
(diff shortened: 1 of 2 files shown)


pkg/visor/rpc.go
```diff
@@ -22,8 +22,12 @@ import (
 const (
 	// RPCPrefix is the prefix used with all RPC calls.
 	RPCPrefix = "app-visor"
-	// HealthTimeout defines timeout for /health endpoint calls.
+	// HealthTimeout defines timeout for /health endpoint calls done from hypervisor.
 	HealthTimeout = 5 * time.Second
+	// InnerHealthTimeout defines timeout for /health endpoint calls done from visor.
+	// We keep it is less than the `HealthTimeout`, so that the outer call would
+	// definitely complete.
+	InnerHealthTimeout = 4 * time.Second
```
jdknives: Should probably be reduced to 3 seconds just to be sure. A DMSG roundtrip can take up to 1.2 seconds in certain circumstances so there should be some more space imo.

jdknives (commented):
Good job. Just a small comment.

jdknives:
Attaching further logs here. We experience issues with the uptime tracker right now wherefore I used the opportunity to re-test this PR. We still get health timeouts and the page does not load consistently. Note that I reduced the inner timeout to 2 seconds for testing. 

[log-timeout-health-request.txt](https://github.com/skycoin/skywire/files/6038849/log-timeout-health-request.txt)

jdknives (changes requested):
Still not solved completely.

PR #686, opened 2021-02-16
Feature/transport labels

### LINKED ISSUE #662: Add transport label field
**Feature description**

We are working on automatic transport setup for visors in different forms (automatic and manual but set by Skycoin to improve network topology). In order to be able to discern the creator of the transport, we should add a field to transports allowing to set labels. 

At the moment, there are three labels that we should support

- User 
- Automatic
- Skycoin

These labels should be transmitted to the hypervisor as well so the UI can display the label of the visor.

i-hate-nicknames:

Fixes #662 

 Changes:	
- Add labels to transport, based on who added the transport

How to test this PR:

### DIFF
--- a/pkg/transport/entry.go
+++ b/pkg/transport/entry.go
@@ -41,11 +41,11 @@ type Entry struct {
 	Label Label `json:"label"`
 }
 
-// NewEntry constructs *Entry
-func NewEntry(localPK, remotePK cipher.PubKey, tpType string, public bool) *Entry {
-	return &Entry{
-		ID:     MakeTransportID(localPK, remotePK, tpType),
-		Edges:  SortEdges(localPK, remotePK),
+// MakeEntry creates a new transport entry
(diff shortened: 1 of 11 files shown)


pkg/transport/entry.go
```diff
@@ -14,6 +14,14 @@ var (
 	ErrEdgeIndexNotFound = errors.New("edge index not found")
 )
 
+type Label string
+
+const (
+	LabelUser      Label = "user"
+	LabelAutomatic       = "automatic"
+	LabelSkywire         = "skywire"
```
jdknives: Should be `Skycoin`

jdknives (commented):
Seems like a good start so far. Just a small change and obviously there is some logic missing but this can wait until we can finally merge the rest of the code.

jdknives (approved):
Looks good.

Issue #688, opened 2021-02-23
Bandwidth meter faulty


jdknives:
**Describe the bug**
The bandwidth logging in the hypervisorUI sometimes breaks and is stuck. No matter how much data is streamed through the VPN for instance, the bandwidth meter does not change. Usually it is stuck at a non-nil value, but there does not seem any other pattern I could discern so far. This has been observed on MacOS as well as Linux. 

As this is not easy to reproduce, I will try to collect more hints for reproduction and possible causes here over time.

PR #689, opened 2021-02-23
Fix remote throughput calculation

Darkren:
Yes
Might be part of #688 

 Changes:	
- Remote throughput previously had a data lag between `atomic.LoadValue` and `atomic.StoreValue`. Not very crucial, but under load, calculation might have been a mess. Now this is done with a single atomic call


jdknives (approved):
Well spotted.

PR #695, opened 2021-02-24
Fix bool flag config representation

### LINKED ISSUE #691: Flag order affects VPN behavior
**Describe the bug**
If the `-killswitch` flag is set first, before `-srv` or `-passcode`, the VPN-client does not work as the flags after the `-killswitch` are not read bz or not provided to the VPN-client. 

Consequently this fails 

```
[
					"-killswitch",
					"true",
					"-passcode",
					"",
					"-srv",
					"02e346d917500aadb71c8b0eafb4125c0e03723a0671dc7352cdd106ab215ef8e5"
				],
```

While this works:

```
[
					"-passcode",
					"",
					"-srv",
					"02e346d917500aadb71c8b0eafb4125c0e03723a0671dc7352cdd106ab215ef8e5",
					"-killswitch",
					"true"
				],
```

Darkren:

Fixes #691 

 Changes:	
- All boolean flags that we support are now represented in format of `--flag_name=value` in the app args.

### DIFF
--- a/pkg/visor/visorconfig/v1.go
+++ b/pkg/visor/visorconfig/v1.go
@@ -203,7 +203,7 @@ func updateStringArg(conf *V1Launcher, appName, argName, value string) bool {
 }
 
 // updateBoolArg updates the cli boolean flag of the specified app config and also within the launcher.
-// All flag names and values are formatted as "name=value" to allow arbitrary values with respect to different
+// All flag names and values are formatted as "-name=value" to allow arbitrary values with respect to different
 // possible default values.
 // The updated config gets flushed to file if there are any changes.
 func updateBoolArg(conf *V1Launcher, appName, argName string, value bool) bool {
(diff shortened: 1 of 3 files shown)


pkg/visor/visorconfig/v1.go
```diff
@@ -151,32 +160,108 @@ func (v1 *V1) UpdateAppArg(launch *launcher.Launcher, appName, argName, value st
...
+
+	configChanged := false
+
+	for i := range conf.Apps {
+		if conf.Apps[i].Name != appName {
+			continue
+		}
+
+		// we format it to have two dashes, just to unify representation
+		fmtedArgName := argName
```
jdknives, line 220: Would it not be cleaner to stick with single dashes? Maybe it does not look as nice, but works just as well.

jdknives (approved):
Looking good. Just one suggestion.

jdknives:
[log.txt](https://github.com/skycoin/skywire/files/6037160/log.txt)

We are also running into a panic when updating the secure mode on an empty VPN server arg list.

pkg/visor/visorconfig/v1.go
Darkren: It doesn't really matter at all, so will do

PR #697, opened 2021-02-25
Feature/transport setup

i-hate-nicknames:
Add support for transport setup service


pkg/transport/setup/visor.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// NewTransportListener makes a TransportListener from configuration
+func NewTransportListener(ctx context.Context, conf *visorconfig.V1, dmsgC *dmsg.Client, tm *transport.Manager) (*TransportListener, error) {
+	log := logging.MustGetLogger("transport_setup")
+	go dmsgC.Serve(ctx)
+	log.WithField("local_pk", conf.PK).Info("Connecting to the dmsg network.")
+	select {
+	case <-dmsgC.Ready():
+		log.Info("Connected!")
```
Darkren: Could you please expand this log line a bit? Will it be clear enough in the overall logs flow?

pkg/visor/init.go
```diff
@@ -220,6 +222,24 @@ func initTransport(v *Visor) bool {
...
 
+func initTransportSetup(v *Visor) bool {
+	report := v.makeReporter("transport_setup")
+	ctx, cancel := context.WithCancel(context.Background())
+	ts, err := ts.NewTransportListener(ctx, v.conf, v.net.Dmsg(), v.tpM)
+	if err != nil {
+		cancel()
+		return report(err)
+	}
+	go func() {
```
Darkren: Can we do just `go ts.Serve(ctx)` here?

Darkren (commented):
Great job! Nothing crucial, just minor comments. Also could you please add some more empty lines between pieces of code? It's actually a very personal stuff, yeah, I just think that it should make code easier to understand

.goreleaser.yml
```diff
@@ -6,7 +6,7 @@ release:
   # Default is extracted from the origin remote URL or empty if its private hosted.
   # Note: it can only be one: either github or gitlab or gitea
   github:
-    owner: skycoin
+    owner: i-hate-nicknames
```
alexadhy: probably should be switched back to `skycoin`?

pkg/visor/visor_test.go
```diff
@@ -1,94 +1,82 @@
...
-	"encoding/json"
-	"io/ioutil"
-	"net/http"
-	"net/http/httptest"
-	"os"
-	"testing"
+// todo: tests here seems to be at very least of integration flavor: taking multiple
+// parts and checking if they work together
+// either remove them completely and rely on tests that setup infrastructure
+// or rewrite them to use mocks
```
alexadhy: I agree with this, this sounds like an integration test.

alexadhy (commented):
awesome work lgtm

.goreleaser.yml
i-hate-nicknames: true, good catch

pkg/visor/visor_test.go
jdknives: In this case, lets remove them and add them to the list of integration tests you are working on in `services` @alexadhy @i-hate-nicknames

pkg/transport/setup/rpc.go
```diff
@@ -0,0 +1,71 @@
...
+// BoolResponse is a simple boolean wrapped in structure for RPC responses
+type BoolResponse struct {
+	Result bool
+}
+
+// AddTransport specified by request
+func (gw *TransportGateway) AddTransport(req TransportRequest, res *TransportResponse) error {
+	ctx := context.Background()
+	gw.log.WithField("PK", req.RemotePK).WithField("type", req.Type).Info("Adding transport")
+	mt, err := gw.tm.SaveTransport(ctx, req.RemotePK, req.Type, transport.LabelSkycoin)
```
jdknives: Logic itself is fine. But I believe this does not return an error if the transport could not be setup or is not up but has been dialed. We probably need to update the logic of this function itself.

pkg/transport/setup/visor.go
```diff
@@ -0,0 +1,89 @@
...
+		}
+		return tl, nil
+	case <-ctx.Done():
+		return nil, fmt.Errorf("failed to connect to dmsg network")
+	}
+}
+
+// Serve transport setup rpc to trusted nodes over dmsg
+func (ts *TransportListener) Serve(ctx context.Context) {
+	ts.log.WithField("dmesg_port", skyenv.DmsgTransportSetupPort).Info("starting listener")
```
jdknives: `dmsg` instead of `dmsg`.

jdknives (commented):
Good job. Just a few small comments.

jdknives:
There is a missing piece of logic here: 

- we should only be able to delete tps that we have created
- we need a function to retrieve all transports setup via `trs`

pkg/visor/visor_test.go
alexadhy: yes, good point. I will.

i-hate-nicknames:
We can theoretically query transport discovery, but I added a method to visor too. Also data from visor should be more up to date. Will add changes to the transport setup service too

pkg/visor/visor_test.go
jdknives: @i-hate-nicknames please remove todo and remove the tests

jdknives (approved):

PR #706, opened 2021-03-01
Replace rakyll/static with native embedding

### LINKED ISSUE #703: Embed frontend files with go embed
**Feature description**

It seems like Go released a built in way to embed static files into a binary. We are currently using a 3rd party tool called `rakyll/statik` for doing that. Using the Go built in tool should be preferable as we can deprecate the 3rd party tool and maybe can remove the need to have a separate build target that rebuilds the frontend for the visor. 

https://blog.carlmjohnson.net/post/2021/how-to-use-go-embed/

Darkren:
Yes
Fixes #703 

 Changes:	
- `rakyll/static` was replaced with native embedding;
- `make build-ui` was modified to make use of native embedding.


cmd/skywire-visor/commands/root.go
```diff
@@ -135,7 +139,15 @@ var rootCmd = &cobra.Command{
...
 // Execute executes root CLI command.
-func Execute() {
+func Execute(ui embed.FS) {
+	uiFS, err := fs.Sub(ui, "static")
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+
+	uiAssets = uiFS
```
jdknives, line 156: Why are we not just passing in `uiFS` below?

jdknives (approved):

jdknives (changes requested):
@darkren good job overall. We need to bump Travis to Go 1.16, update the README and please answer my query above. Then this can go. Well done!

cmd/skywire-visor/commands/root.go
Darkren: @iketheadore `Execute` is a cobra lib function, we can't pass it in there

Makefile
```diff
@@ -51,9 +52,15 @@ BUILD_OPTS_DEPLOY?="-ldflags=$(BUILDINFO) -w -s"
...
+move-built-frontend:
+	rm -rf ${MANAGER_UI_BUILT_DIR}
+	mkdir ${MANAGER_UI_BUILT_DIR}
+	cp -r ${MANAGER_UI_DIR}/dist/. ${MANAGER_UI_BUILT_DIR}
+	rm -rf ${MANAGER_UI_DIR}/dist/
 
-build-static: host-apps-static bin-static ## Build apps and binaries. `go build` with ${OPTS}
+build: dep move-built-frontend host-apps bin ## Install dependencies, build apps and binaries. `go build` with ${OPTS}
+
+build-static: move-build-frontend host-apps-static bin-static ## Build apps and binaries. `go build` with ${OPTS}
```
jdknives: Will not work because of spelling issue.

Makefile
```diff
@@ -51,9 +52,15 @@ BUILD_OPTS_DEPLOY?="-ldflags=$(BUILDINFO) -w -s"
 
 check: lint test ## Run linters and tests
 
-build: host-apps bin ## Build apps and binaries. `go build` with ${OPTS}
+move-built-frontend:
+	rm -rf ${MANAGER_UI_BUILT_DIR}
+	mkdir ${MANAGER_UI_BUILT_DIR}
+	cp -r ${MANAGER_UI_DIR}/dist/. ${MANAGER_UI_BUILT_DIR}
+	rm -rf ${MANAGER_UI_DIR}/dist/
```
Senyoret1: I’m not sure if this is needed, as that folder is already removed from the repo. It indeed prevents having duplicated files, but also has the potential of forcing to rebuild the UI if the folder is needed again for any reason.

Makefile
```diff
@@ -51,9 +52,15 @@ BUILD_OPTS_DEPLOY?="-ldflags=$(BUILDINFO) -w -s"
...
 
-build-static: host-apps-static bin-static ## Build apps and binaries. `go build` with ${OPTS}
+build: move-built-frontend host-apps bin ## Install dependencies, build apps and binaries. `go build` with ${OPTS}
```
Senyoret1: I think having `move-built-frontend` here is a problem. With this, now the `dist` folder will have to be present every time anybody wants to build the binaries. As it was removed from the repo, that would mean the need to have all the front-end tools intalled and build the UI each time befor building the binaries, which would be inconvinient. Also, the process would be complicate even having the front-end tools, because `make build-ui` calls `make move-built-frontend`, which would remove the `dist` folder just after finishing, so a call to `make build` just after calling `make build-ui` would fail. I think the solution would be to remove `move-built-frontend` from here.

Maybe the same applies to `build-static`.

Senyoret1 (commented):
I’m not sure about how the embedding procedure works, but it worked well in my test, so appart from my comments I don’t find any problems :+1:

Makefile
Darkren: @Senyoret1 agreed. but we definitely need `dist` in `cmd/skywire-visor` to build binaries now. native embedding requires that files being embedded are ready at compile time. I removed this command but left it in `build-ui`. so that built ui in the end lies where needed

Makefile
Darkren: removed the last line

PR #685, opened 2021-02-16
Move `run` command from `vpn` to `osutil`, add `RunWithResult`

Darkren:
Did you run `make format && make check`?

 Changes:	
- `run` func was made public and moved to `osutil`;
- `RunWithResult` was added to `osutil` to run OS command and fetch output reader.

PR #699, opened 2021-02-26
Improvements for the visor list

Senyoret1:
Did you run `make format && make check`?
The go code was not changed. `npm run lint` and `npm run build` were used.

 Changes:	
- Now the UI uses the new `GET /api//visors-summary` API endpoint, added in https://github.com/skycoin/skywire/pull/666 , for populating the visor list. This change allows to load the visor list faster.

- A problem with the offline visors in the DMSG tab was solved. Before the changes, part of the data of the offline visors in the DMSG tab was shown like this:
D

How to test this PR:
For the first change, refresh the visor list, the data should load faster. For the second change, open the DMSG tab in the visor list while there is an offline node on it, the list should not show invalid data.

NOTE: this PR needs the changes from https://github.com/skycoin/skywire/pull/666 to work well.

PR #716, opened 2021-03-12
Display local IP

### LINKED ISSUE #389: Display local IP
**Feature description**

It would be helpful to see the local ip-address of the visor inside the details.
Also for skychat it would be nice to have a hyperlink inside the visor-app-list to the ip-address:port-of-skychat. 
For example a world-logo like in the picture below.
So a hyperlink column for applications where you need web-interface.

Tried skychat and had no idea which pubkey/visor belongs to which ip-address:8001 without doing further research.

image

Darkren:
Yes
Fixes #389 

**Works only on Linux and MacOS systems**

This PR also contains all the commits from #685 . So if this gets merged, #685 may be closed for good.

 Changes:	
- `DefaultNetworkInterface`, `LocalNetworkInterfaceIPs`, `NetworkInterfaceIPs` were moved from `vpn` to `util/netutil`;
- `DefaultNetworkInterface` was implemented for MacOS;
- `local_ip` field was added to the response from `/api/visors/{pk}.

How to test this PR:
1. Run visor with hypervisor
2. Open network console in browser
3. Open visor page
4. Check response from `/api/visors/{pk}`. It should contain `local_ip` field with the corresponding value for you default network interface IP


jdknives (approved):
Works well on Mac.

Senyoret1:
Worked well in my test using Linux. However, I’m not sure if it will work well with https://github.com/skycoin/skywire/pull/699 and the PR it depends on. Also, quick question: if it works only in Mac and Linux, what is the expected content of the `local_ip` field when using Windows?

Darkren:
> Worked well in my test using Linux. However, I’m not sure if it will work well with #699 and the PR it depends on. Also, quick question: if it works only in Mac and Linux, what is the expected content of the `local_ip` field when using Windows?

@Senyoret1 previously it couldn't be built on Windows. but I fixed it, now it will return empty string for Win

PR #721, opened 2021-03-19
Show the IP in the UI

### LINKED ISSUE #719: Use the ip returned by the API
After https://github.com/skycoin/skywire/pull/716 is merged, the IP will be available for the UI. After that, it will be possible to solve https://github.com/skycoin/skywire/issues/389 and the IP should be used as the default label of the visors.

Senyoret1:

Fixes #719

 Changes:	
- The visor IP is now used as the default label for visors without label.
- The visor details page now shows the visor IP.
- Now the apps list shows a link for opening the UI of the Skychat app in a new tab. For building the URL, the code uses the visor IP and tries to get the port from the app arguments, but if the port is not there, `8001` is used as the default port.

How to test this PR:
To see the default label, just open the visor list, the IP should be shown as the label of all visors without a label set (previously, the default label was just a part of the public key). For checking the IP of a visor, open the visor details page, the IP should be shown in the right bar. For opening the Skywallet UI, open the app list of a visor and press the link button at the right of the Skychat row.


jdknives:
This works well. Will merge. Need to revisit the implementation on the backend though for MacOS.

PR #723, opened 2021-03-24
Update mainnet_rules.md

fray:
I corrected some typos and grammatical errors, and added clarification where necessary and appropriate. (Particular attention was paid to clarifying the language around the rule that a whitelisted user can only receive rewards for a DIY miner *or* one or more official Skyminers, but not both.)

### DIFF
--- a/mainnet_rules.md
+++ b/mainnet_rules.md
@@ -30,34 +30,34 @@ This article represents the central source of information for the Skywire mainne
 ## Rules
 This section covers the rules of the mainnet. Dependent on the amount of miners that are associated with your emails, some of these rules may apply to you or not.
 
-As of now, each person participating in the Skywire mainnet may own one DIY miner and multiple official Skyminers wrt. to being eligible for rewards.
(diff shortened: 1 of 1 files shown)


mainnet_rules.md
```diff
@@ -30,34 +30,34 @@ This article represents the central source of information for the Skywire mainne
 ## Rules
 This section covers the rules of the mainnet. Dependent on the amount of miners that are associated with your emails, some of these rules may apply to you or not.
 
-As of now, each person participating in the Skywire mainnet may own one DIY miner and multiple official Skyminers wrt. to being eligible for rewards.
-<div align="center"><b>"A DIY miner cannot be in the same location as another DIY or official Skyminer.</b></div><br>
+Currently, each person participating in the Skywire mainnet may own one DIY miner or multiple official Skyminers to be eligible for rewards.
```
Asgaror: Distinction between and/or is necessary. It's not just _or_ and not just _and_ but it could be either of them dependent on the location of the Skyminers.

fray: Yes, I agree. I've changed that to an "and/or" in my pull request.

Asgaror (approved):
All good to go.

PR #724, opened 2021-03-24
Init refactoring

### LINKED ISSUE #710: Improve startup performance
**Feature description**

Currently the startup of the `skywire-visor` takes a considerable amount of time because we run the startup steps sequentially. A lot of modules can be started up in parallel.

i-hate-nicknames:

Fixes #710 	

 Changes:	
- Split visor initialization into multiple units of work, and run them concurrently to improve startup performance

How to test this PR:
Build with -race and run skywire-visor


pkg/visor/init.go
```diff
@@ -34,54 +35,132 @@ import (
...
+	if err != nil {
+		return err
+	}
+	report := v.makeReporter("address-resolver")
+	conf := v.conf.Transport
+
+	arClient, err := arclient.NewHTTP(conf.AddressResolver, v.conf.PK, v.conf.SK)
+	if err != nil {
+		err := fmt.Errorf("failed to create address resolver client: %w", err)
+		report(err)
```
Darkren: Sometimes we return plain error, sometimes we return `report(err)`, sometimes we just call `report(err)` but return plain error, like here. Shouldn't we make this consistent? I know that `report` has been here before you started on this PR, so probably the `report` func should be re-visited. Probably it's not even needed with the new approach

pkg/visor/init.go
```diff
@@ -34,54 +35,132 @@ import (
...
-		initHypervisors,
-		initUptimeTracker,
-		initTrustedVisors,
-		initHypervisor,
-	}
+type visorCtxKey int
+
+const visorKey visorCtxKey = iota
+
+var up, ebc, ar, disc, sn, pty, tr, rt, launch, cli, hvs, ut, trv, hv vinit.Module
```
Darkren: Probably would be good to group these with braces and add some comments on each with the full name. A lot abbreviations may seem scary especially if one looks at this for the first time

pkg/visor/init.go
```diff
@@ -34,54 +35,132 @@ import (
...
+		&tr, &rt, &launch, &cli, &hvs, &ut, &trv)
+	hv = vinit.MakeModule("hypervisor", initHypervisor, logger, &vis)
 }
 
-func initUpdater(v *Visor) bool {
-	report := v.makeReporter("updater")
+// -------------------------------------------------------------------
 
+func initUpdater(ctx context.Context, log *logging.Logger) error {
+	v, err := getVisor(ctx)
```
Darkren: This call is present in all init funcs, and they all the same signature. Can we change this signature and add some kind of wrapper which will fetch visor from context and pass it to init func? Should reduce copy-paste amount

Darkren (commented):
I'm not sure about my comments, these are mostly questions, so I don't approve or request changes explicitly. But anyway, that is an impressive job right here!

pkg/visor/init.go
i-hate-nicknames: Yes, it is weird. I'm strongly leaning towards removing report and corresponding calls to it. Basically, I left it untouched for the time being. Currently it is used to check for errors in concurrent code, but in my opinion it's faulty and can skip errors. So that's definitely an area to improve yet

pkg/visor/init.go
i-hate-nicknames: Thanks, good advice

pkg/visor/init.go
i-hate-nicknames: These functions require visor, and ideally need visor argument. But it's not possible because that would result in circular import: the modules package will depend back on visor.

I initially used wrapping, but ended up with a much more cumbersome method that what it is currently:
https://github.com/skycoin/skywire/blob/92e84034eea0081c29d0ad2b2fe6e543856345a9/pkg/visor/init.go

Extra indentation and multiple context arguments were just too much, so I just made it dead-simple. But now that I think about it there seems to be a better way of doing it than I did the first time. I'll try to move it somewhere to avoid code repetition, thanks.

ersonp (approved):
The changes seem good.

pkg/visor/visorinit/module.go
```diff
@@ -0,0 +1,172 @@
...
+// first to last and depth first
+// If any of the underlying dependencies, or this module initialize with error, return that error
+func (m *Module) InitSequential(ctx context.Context) error {
+	// early quit if initialized
+	ok := m.start()
+	if !ok {
+		return nil
+	}
+	defer m.stop()
+	m.log.Infof("Starting")
```
alexadhy: I think this and line 141 can just use `Info` rather than `Infof`?

pkg/visor/init.go
```diff
@@ -181,10 +259,11 @@ func initTransport(v *Visor) bool {
 		DiscoveryClient: tpdC,
 		LogStore:        logS,
 	}
-
-	tpM, err := transport.NewManager(v.MasterLogger().PackageLogger("transport_manager"), v.net, &tpMConf)
+	managerLogger := v.MasterLogger().PackageLogger("transport_manager")
+	tpM, err := transport.NewManager(managerLogger, v.net, &tpMConf)
```
mrpalide, line 263: Does any reason for this change? [a personal question]

mrpalide (commented):
I have a good feeling after reviewing this PR. Awesome code. 
I just had a quick [personal] question that I left a comment on there [visor/init.go].

pkg/visor/visorinit/module.go
```diff
@@ -0,0 +1,172 @@
...
+		return true
+	default:
+		return false
+	}
+}
+
+// InitSequential initializes all module dependencies recursively and sequentially, one by one
+// first to last and depth first
+// If any of the underlying dependencies, or this module initialize with error, return that error
+func (m *Module) InitSequential(ctx context.Context) error {
```
jdknives: The whole function can be removed imo. I dont see a case where we want to initialize the visor sequentially given that the PR works.

jdknives (approved):

pkg/visor/init.go
i-hate-nicknames: No, I don't think there is, maybe it's a bit shorter and looks better. Could be that there were more changes that required logger variable but later I removed them.

PR #735, opened 2021-04-20
Feature/runtime logs

### LINKED ISSUE #725: Allow to retrieve full logs from UI
**Feature description**

We need to provide users with an easier way to retrieve the full logs of the visor. Currently we can display application logs but no system logs. 

**Ideas**

If the visor is run as `systemd` service we can simply return the last few thousand lines of logs to the hypervisor. If we are not running it as `systemd` we need to think about a different solution.

i-hate-nicknames:

Fixes #725 

 Changes:	
- Add `GET /api/visors/{pk}/runtime-logs` endpoint

How to test this PR:
Run visor with hypervisor. Check output of the request.

Response is an array of json-encoded log entries. Entries are created by logger clients and can have custom fields with values, e.g. PK field that provide additional info.
Each log entry should have the following fields:
`time`: formatted timestamp of the log entry
`msg`: message of the entry
`level`: level of the entry
`log_line`: log entry line in the global log. Since the capacity of log store is limited, some old entries could be dropped.
All other keys may be absent in some entries.


pkg/visor/logstore/logstore.go
```diff
@@ -0,0 +1,86 @@
...
+	store := &store{max: int64(max), entries: entries, formatter: formatter}
+	return store, store
+}
+
+type store struct {
+	max       int64
+	n         int64
+	entries   []*logrus.Entry
+	formatter logrus.Formatter
+}
```
mrpalide: What is **n** exactly? In the GetLogs function use as number of log lines, isn't?

pkg/visor/logstore/logstore.go
```diff
@@ -0,0 +1,86 @@
...
+	return logrus.AllLevels
+}
+
+func (s *store) Fire(entry *logrus.Entry) error {
+	idx := s.n % s.max
+	e := entry.WithField(LogRealLineKey, s.n)
+	s.entries[idx] = e
+	s.n++
+	return nil
+}
```
mrpalide: What is the **Fire** function exactly?

pkg/visor/logstore/logstore.go
i-hate-nicknames: It's the number of logs processed and the number of the next log entry to process.
It's used to calculate the index in the limited capacity of the store and also could be used to append actual log entry number to the response. I refactored names a bit, n is probably too generic

pkg/visor/logstore/logstore.go
i-hate-nicknames: Fire is a part of logrus.Hook interface that I used to get log entries from the logger. Basically this function is called when a new entry is created by logger

Senyoret1:
I found some things:

- I tested it and got 50+ elements in the response, but all of them had `level` set to `panic`. Some entries just had info about API responses, so I think there is a problem with that.

- The response appears to be JSON, but the format is not correct, because the elements are not really inside an array. Opening it with Firefox causes this error to appear:
er

Appart from that, the response is somewhat complex, as there are a lot of possible properties and not all entries have the same ones. If the manager is just going to open the whole response in a new tab, there is no problem, but if the manager needs to show the data in a more processed way then we will need to plan what will be shown, to check if the data currently being returned will allow to do that.

i-hate-nicknames:
Thanks. Another issue that I just found is that not logs get to the hook for some reason. Apparently, new subloggers created by e.g. dmsg do not inherit hooks.

@Senyoret1 log entries are inherently heterogeneous, because callers add their own data that makes sense to them to the entries. JSON representation just gets all the data that is there.
I thought about returning just textual representation of an entry, but this way there will be no coloring and it might get long.
Would it be possible to use the generic parts of a log entry for presentation (level, date, log line, message), and show every other field as "additional fields" somewhere, like a dropdown?

jdknives:
@i-hate-nicknames this endpoint and the returned logs do not need to be colored to be honest. We will mostly use this for diagnostics so we may even just return a file to the user with plain text. Colored and nicely formatted text in the UI would be cool but not really necessary so I would keep it simple.

pkg/visor/logstore/logstore.go
```diff
@@ -0,0 +1,94 @@
...
+// log entries are formatted with store formatter and concatenated together
+func (s *store) GetLogStr() (string, error) {
+	logs, _ := s.GetLogs()
+	var sb strings.Builder
+	for _, entry := range logs {
+		bs, err := s.formatter.Format(entry)
+		if err != nil {
+			return "", err
+		}
+		sb.WriteString(string(bs))
```
alexadhy: Cool, just wonder why is it not a `bytes.Buffer` instead of `strings.Builder` here, might save conversions from `[]byte to string`

i-hate-nicknames: True. I ended up storing strings instead of log entries since only strings are ever used and can avoid excessive format calls, so this code is no more

i-hate-nicknames:
@Senyoret1 ended up using json formatter, because standard text formatter only pretty-prints entries when using colors. I will update PR description with fields that are included in every entry

pkg/visor/logstore/logstore.go
```diff
@@ -0,0 +1,81 @@
+package logstore
+
+import (
+	"github.com/sirupsen/logrus"
+)
+
+// LogRealLineKey is a key in the log enry that denote real log line number
```
jdknives: `entry` and `denotes`

pkg/visor/logstore/logstore.go
```diff
@@ -0,0 +1,81 @@
...
+// and ending at to (not inclusive)
+func (s *store) collectLogs(from, to int64) []string {
+	logs := make([]string, 0)
+	for i := from; i < to; i++ {
+		logs = append(logs, s.entries[i])
+	}
+	return logs
+}
+
+// GetLogs returns most resent log lines (up to cap log lines is stored)
```
jdknives: `recent`

jdknives (approved):
Good job.

PR #743, opened 2021-04-27
Feature/public visors advertising

### LINKED ISSUE #574: Allow visors to advertise themselves as running on Public IP

i-hate-nicknames:
Did you run make format && make check?

Fixes #574

Changes:

On start visor registers itself in the service discovery. Visor attaches IPs of its local active network interface to the request;
If `public_autoconnect` field is set to true in config, visor will try to connect to 5 known visors on its start.


i-hate-nicknames:
One thing I dislike is that I had to rely on callbacks in [autoconnect](https://github.com/skycoin/skywire/pull/743/files#diff-c467701eaae6064f56b622ff8ea5c73e25a64661555af80245142e1c0f456659). This is due to the reason that autoconnect needs transport manager, but importing transport manager introduces dependency cycle: transport package depends on servicedisc package.

I'm thinking about restructuring it a better way, maybe splitting some packages. Input on this subject is welcome.

ersonp:
@i-hate-nicknames I was testing out the PR and `skywire-cli visor gen-config` generates config for `v1.1.0` and not for `v1.1.1`.

i-hate-nicknames:
Weird, I just checked and it generated 1.1.1 for me. Is it possible that you didn't build with `make build` and accidentally ran an old binary?

ersonp:
`make build; make install` solved it. And `connect_to_public_visors` mentioned above. I could not find it in the config. Perhaps you meant `public_autoconnect`.

pkg/servicedisc/client.go
```diff
@@ -98,8 +101,12 @@ func (c *HTTPClient) Auth(ctx context.Context) (*httpauth.Client, error) {
 }
 
 // Services calls 'GET /api/services'.
-func (c *HTTPClient) Services(ctx context.Context) (out []Service, err error) {
-	req, err := http.NewRequestWithContext(ctx, http.MethodGet, c.addr("/api/services", c.entry.Type), nil)
+func (c *HTTPClient) Services(ctx context.Context, quantity int) (out []Service, err error) {
+	addr := c.addr("/api/services", c.entry.Type)
+	if quantity != 0 {
```
jdknives: Should we not use `net/http` functionality for building query parameter?

pkg/servicedisc/autoconnect.go
```diff
@@ -0,0 +1,99 @@
...
+	connector.log = log
+	connector.conns = make(map[cipher.PubKey]struct{})
+	return connector
+}
+
+// Run implements Autoconnector interface
+func (a *autoconnector) Run(ctx context.Context, connector ConnectFn, checker CheckConnFN) error {
+	retrier := netutil.NewRetrier(fetchServicesDelay, 0, 2)
+	for {
+		time.Sleep(PublicServiceDelay * 2)
```
jdknives: Why are doubling the value here instead of just setting a different value? Afaics this constant is only used here.

pkg/servicedisc/autoconnect.go
```diff
@@ -0,0 +1,99 @@
...
+			} else {
+				a.conns[pk] = struct{}{}
+			}
+		}
+	}
+}
+
+// check if existing connections are still active using checker
+// and delete those that are not
+func (a *autoconnector) checkConns(checker CheckConnFN) {
```
jdknives: I think it is preferable to use the transport module's built in functions to actually check if a given transport is still up. The connections/transports we are establishing are just that: transports. So we should not need custom checks. If something with the functionality of the transport module is insufficient, we need to fix things there imo.

pkg/servicedisc/autoconnect.go
```diff
@@ -0,0 +1,99 @@
...
+// Autoconnector continuously tries to connect to services
+type Autoconnector interface {
+	Run(context.Context, ConnectFn, CheckConnFN) error
+}
+
+type autoconnector struct {
+	client   *HTTPClient
+	maxConns int
+	log      *logging.Logger
+	conns    map[cipher.PubKey]struct{}
```
jdknives: I guess we are currently keeping a list of conns here to be able to differentiate between autoconnect transports and manually added user transports. If thats the only case to keep them here, then this PR might profit from merging #686 being merged first. I think we could just use the transport labels to differentiate between the autoconnect tps instead of keeping a separate structure for this purpose.

pkg/servicedisc/autoconnect.go
```diff
@@ -0,0 +1,99 @@
...
+			return err
+		}
+
+		for _, service := range services {
+			pk := service.Addr.PubKey()
+			if _, ok := a.conns[pk]; ok {
+				continue
+			}
+			err := connector(ctx, pk)
+			if err != nil {
```
jdknives: Just log and retry with the next iteration of the `autoconnector.Run` being run should be fine I guess.

pkg/servicedisc/types.go
```diff
@@ -104,11 +104,12 @@ type Stats struct {
...
 // Service represents a service entry in service-discovery.
 type Service struct {
-	Addr    SWAddr       `json:"address"`
-	Type    string       `json:"type"`
-	Stats   *Stats       `json:"stats,omitempty"` // TODO: Have this implemented.
-	Geo     *GeoLocation `json:"geo,omitempty"`
-	Version string       `json:"version,omitempty"`
+	Addr     SWAddr       `json:"address"`
+	Type     string       `json:"type"`
+	Stats    *Stats       `json:"stats,omitempty"` // TODO: Have this implemented.
```
jdknives: Not part of your PR, but the basics of `Stats` has been implemented for other services already. So I guess we dont need this comment anymore?

pkg/visor/init.go
```diff
@@ -487,28 +488,66 @@ func initUptimeTracker(v *Visor) bool {
...
-					WithField("pk", pk).
-					WithField("type", trustedVisorsTransportType).
-					Warnf("Failed to add transport to trusted visor via")
-			} else {
-				v.log.
-					WithField("pk", pk).
-					WithField("type", trustedVisorsTransportType).
-					Infof("Added transport to trusted visor")
-			}
+	_, portStr, err := net.SplitHostPort(v.conf.STCP.LocalAddr)
```
jdknives: Why is the local address of a visor needed in this case. Setting up `stcpr` should work without the IP address of the local or the remote being specified in config. Usually no one defines teh `stcp` config fields because its manual and error prone wherefore we added `stcpr` to avoid having to define that manually.

jdknives (commented):
Overall good work although some of the implementation details might be improvable.

pkg/servicedisc/autoconnect.go
i-hate-nicknames: The problem is that transport package depends on appdisc package, so I cannot directly use transport manager in autoconnector due to circular dependency.
Instead, init creates callbacks that will use transport manager, but autoconnector just runs callbacks for connection and checking connection.
I skimmed through appdisc and transport packages quickly, but couldn't find an easy way to take them apart. This design is certainly not ideal, maybe it would be better to find a way to split that dependency.

pkg/servicedisc/autoconnect.go
i-hate-nicknames: Only autoconnect transport PKs are added here. This set is used to limit the number of connections we keep, otherwise we would just ask for more and more connections. Making use of transport package facilities should be better, I need to think about it.

pkg/servicedisc/client.go
i-hate-nicknames: I'm not sure there are any url builders in standard go library, at least I couldn't find them. There is only net/url that allows building query parameters. Can you please point me in the right direction?
In any case `addr` function can use some improvements, thanks

pkg/visor/init.go
i-hate-nicknames: It's been like that in commits before me, I only tested that it works and restructured it later.
I read through the code a bit and I believe the problem is that appdisc client needs port value to be constructed, which in turn is required because every such client also advertises its service. It currently doesn't support just fetching public services. Needs closer look

pkg/servicedisc/client.go
jdknives: https://golang.org/pkg/net/url/#example_URL Actually in `net/url`. But the example looks liek what we are doing here, right?

pkg/servicedisc/client.go
```diff
@@ -8,20 +8,23 @@ import (
...
 	"github.com/skycoin/skywire/internal/httpauth"
+	nu "github.com/skycoin/skywire/internal/netutil"
 	"github.com/skycoin/skywire/pkg/util/buildinfo"
+	"github.com/skycoin/skywire/pkg/util/netutil"
 )
 
 var (
 	// ErrVisorUnreachable is returned when visor is unreachable.
 	ErrVisorUnreachable = errors.New("visor is unreachable")
+	updateRetryDelay    = 5 * time.Second
```
alexadhy: Very good code, I think this would be better as a `const` perhaps? since on `Retrier` we copy the exponentialBackoff duration value so we don't modify this?

pkg/util/netutil/net_darwin.go
```diff
@@ -0,0 +1,26 @@
...
+package netutil
+
+import (
+	"bytes"
+	"fmt"
+	"os/exec"
+)
+
+const (
+	defaultNetworkInterfaceCMD = "netstat -rn | sed -n '/Internet/,/Internet6/p' | grep default | awk '{print $4}'"
```
alexadhy, line 12: could also use `route -n get default | awk 'FNR == 5 {print $2}'`, which is a bit shorter I guess.

i-hate-nicknames: Unfortunately, I do not have a mac available for me to test this properly so can't change this. Doesn't sound like a big issue, so will keep as it is

i-hate-nicknames:
Possible improvement: once https://github.com/skycoin/skywire/pull/686 is merged, try making use of transport labels and check all transport manager connections instead of keeping a separate set in autoconnect.go

pkg/servicedisc/autoconnect.go
```diff
@@ -42,83 +41,83 @@ func MakeConnector(conf Config, maxConns int, tm *transport.Manager, log *loggin
...
-func (a *autoconnector) connect(ctx context.Context, pk cipher.PubKey) {
-	a.log.WithField("pk", pk).Infoln("Adding transport to public visor")
-	if _, err := a.tm.SaveTransport(ctx, pk, tptypes.STCPR, transport.LabelAutomatic); err != nil {
-		a.log.
-			WithError(err).
-			WithField("pk", pk).
-			WithField("type", tptypes.STCPR).
-			Warnln("Failed to add transport to public visor")
-		return
+func (a *autoconnector) calculateAbsent(pkeys []cipher.PubKey, trs []*transport.ManagedTransport) []cipher.PubKey {
```
jdknives: We are removing potential duplicates from the list of to be established transports here, right? Woudl it be better to name it something like `isDuplicate` or something similar?

pkg/servicedisc/autoconnect.go
```diff
@@ -42,83 +41,83 @@ func MakeConnector(conf Config, maxConns int, tm *transport.Manager, log *loggin
...
+		if !tr.IsUp() {
+			a.tm.DeleteTransport(tr.Entry.ID)
+		} else {
+			trsActive = append(trsActive, tr)
 		}
 	}
-	for _, pk := range toDelete {
-		delete(a.conns, pk)
-	}
+	return trsActive
```
jdknives: Usually we use `tps` for transports in variable names. We should probably stick with it.

pkg/servicedisc/autoconnect.go
```diff
@@ -42,83 +41,83 @@ func MakeConnector(conf Config, maxConns int, tm *transport.Manager, log *loggin
...
+		}
+		return nil
+	}
+	if err := retrier.Do(fetch); err != nil {
+		return nil, err
 	}
-	up := t.IsUp()
-	if !up {
-		a.tm.DeleteTransport(t.Entry.ID)
+	var pkeys []cipher.PubKey
```
jdknives: Can you check if we use this name elsewhere? I think `pks` or `keys` is more canonical.

jdknives (approved):

alexadhy (approved):
Awesome code, lgtm.

Issue #747, opened 2021-04-30
Push visor docker image to Dockerhub


jdknives:
**Describe the bug**

We currently do not seem to push the visor docker image we have to dockerhub. We need to push it there with every push to `develop` or `master` with a tag system similar to what we do for services. The image should be pushed to skycoin/skywire.

jdknives:
We also have a dockerfile `skywire-runner`. I am not sure of its purpose. But maybe it could be removed/made obsolete with a cleaner setup.

jdknives:
I added a `DOCKER_USERNAME` and `DOCKER_PASSWORD` to Travis that can be accessed to login to our Dockerhub.

PR #749, opened 2021-04-30
push to dockerhub on push to develop or master

alexadhy:

Fixes: #747 

 Changes:	
- Builds the dockerfile in the `docker/images/visor/Dockerfile`
- Tag and push the image to dockerhub if it's a push to `master` or `develop` branch
- Added a few build-arg to the docker image creation under 

How to test this PR:

- WIP


.travis.yml
```diff
@@ -35,3 +39,10 @@ script:
   - make install-deps-ui
   - make lint-ui
   - make build-ui
+
+deploy:
+  provider: script
+  script: make docker-push
+  on:
+    type: push
+    condition: $TRAVIS_BRANCH =~ ^(develop|master)$
```
jdknives: I would probably follow the `skywire-services` standard here and use the `branch` field as well as the condition to only push if it is not a PR. 

This would also eliminate some code in the Makefile that we otherwise dont need. And the Makefile is already pretty cluttered imo.

docker/images/visor/Dockerfile
```diff
@@ -35,3 +39,9 @@ ENTRYPOINT [ "./skywire-visor" ]
 
 # default target
 FROM  visor-runner
+
+LABEL org.opencontainers.image.created="${BUILD_DATE}" \
```
jdknives: What do we need those labels for?

Makefile
```diff
@@ -159,7 +170,19 @@ build-ui: install-deps-ui  ## Builds the UI
 
 # Dockerized skywire-visor
 docker-image: ## Build docker image `skywire-runner`
-	docker image build --tag=skywire-runner --rm  - < skywire-runner.Dockerfile
+	#docker image build --tag=skywire-runner --rm  - < skywire-runner.Dockerfile
```
jdknives: Probably can be removed. Check if there are any references to the skywire-runner file and remove it if we can.

Makefile
```diff
@@ -159,7 +170,19 @@ build-ui: install-deps-ui  ## Builds the UI
...
+	--tag=skycoin/skywire:${DOCKER_IMAGE_TAG_NAME} \
+	--build-arg BUILD_DATE=${DATE} \
+	--build-arg VCS_REF=${DOCKER_VCS_REF} \
+	--build-arg IMG_NAME=skywire \
+	-f ./docker/images/visor/Dockerfile .
+
+docker-push: docker-image
+	docker login -u ${DOCKER_USERNAME} -p ${DOCKER_PASSWORD}
+	docker tag skycoin/skywire:${DOCKER_IMAGE_TAG_NAME}  skycoin/skywire:${DOCKER_IMAGE_TAG_NAME}
+	#docker container commit ${COMMIT} skycoin/skywire:${DOCKER_IMAGE_TAG_NAME}
```
jdknives: Can this commented line be removed?

.travis.yml
alexadhy: ok, will do so. Thanks

docker/images/visor/Dockerfile
alexadhy: Just to help organize things perhaps? It's not very useful right now I know. I'll remove it

Makefile
alexadhy: I think it's not being used, but I'll investigate more. If it's not being used anywhere, I'll remove it on follow-up commits.

Makefile
alexadhy: ok

Makefile
```diff
@@ -25,8 +25,6 @@ DOCKER_IMAGE?=skywire-runner # docker image to use for running skywire-visor.`go
 DOCKER_NETWORK?=SKYNET
 DOCKER_NODE?=SKY01
 DOCKER_OPTS?=GO111MODULE=on GOOS=linux # go options for compiling for docker container
-DOCKER_VISOR_IMAGE_NAME?=skywire
-DOCKER_VCS_REF?=${BRANCH}-${COMMIT}
 DOCKER_IMAGE_TAG_NAME:=""
 ifeq (${BRANCH},master)
```
jdknives: Can we just pass a var to the command here instead of using the Makefile vars to infer branch?

alexadhy: should we just use (like `skywire-services`) a shell script to do it (take arg)? would be cleaner I guess

jdknives: Sounds fine to me @alexadhy

.travis.yml
```diff
@@ -41,8 +41,14 @@ script:
...
+  - provider: script
+    script: make docker-push
+    on:
+      branch: master
+      condition: $TRAVIS_PULL_REQUEST = false
+  - provider: script
+    script: make docker-push
+    on:
+      branch: develop
+      condition: $TRAVIS_PULL_REQUEST = true
```
jdknives: Why are we setting true here?

alexadhy: fixed in the follow-up commit, sorry it was a silly mistake.

Makefile
```diff
@@ -13,6 +13,7 @@ VERSION := $(shell git describe)
 RFC_3339 := "+%Y-%m-%dT%H:%M:%SZ"
 DATE := $(shell date -u $(RFC_3339))
 COMMIT := $(shell git rev-list -1 HEAD)
+BRANCH := $(shell git rev-parse --abbrev-ref HEAD)
```
jdknives: Can be removed, right?

WinWisely268: yes i will remove it also

alexadhy: done in the followup commit

README.md
```diff
@@ -38,27 +40,51 @@ In order to expose the hypervisor UI, generate a config file with `--is-hypervis
 $ skywire-cli visor gen-config --is-hypervisor
 ```
 
+Docker container will create config automatically for you, should you want to run it manually, you can do:
+
+```bash
+$ docker run --rm -v <YOUR_CONFIG_DIR>:/opt/config \
```
jdknives: We should mount the config to `/opt/skywire/` as I think that is the actual path on our other installations.

jdknives (commented):
Looks good otherwise. Will need to test it.

PR #750, opened 2021-04-30
Solve nil pointer derefrence

### LINKED ISSUE #740: Panic on visor removal from hypervisor
**Describe the bug**
When a visor is being disconnected from a hypervisor, the hypervisor crashes with the following panic

[log.txt](https://github.com/skycoin/skywire/files/6359443/log.txt)

**Environment information:**

- OS: e.g. Arch Linux 
- Branch: `develop`

**Steps to Reproduce**

1. Setup hypervisor
2. Connect visor to hypervisor
3. Disconnect visor
4. Check hypervisor logs

ersonp:
yes

Fixes #740 

 Changes:	
- Initialised `Summary` struct inside `ExtraSummary`

How to test this PR:
1. Setup hypervisor
2. Connect visor to the hypervisor
3. Disconnect visor
4. Check hypervisor logs

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -275,7 +275,6 @@ golang.org/x/crypto v0.0.0-20201208171446-5f87f3452ae9 h1:sYNJzB4J8toYPQTM6pAkcm
 golang.org/x/crypto v0.0.0-20201208171446-5f87f3452ae9/go.mod h1:jdWPYTVW3xRLrWPugEBEK3UY2ZEsg3UU495nc5E+M+I=
 golang.org/x/lint v0.0.0-20181026193005-c67002cb31c3/go.mod h1:UVdnD1Gm6xHRNCYTkRU2/jEulfH38KcIWyp/GAMgvoE=
 golang.org/x/lint v0.0.0-20190313153728-d0100b6bd8b3/go.mod h1:6SW0HCj/g11FgYtHlgUYUwCkIfeOF89ocIRzGO/8vkc=
-golang.org/x/mod v0.2.0 h1:KU7oHjnv3XNWfa5COkzUifxZmxp1TyI7ImMXqFxLwvQ=
 golang.org/x/mod v0.2.0/go.mod h1:s0Qsj1ACt9ePp/hMypM3fl4fZqREWJwdYDEqhRiZZUA=
 golang.org/x/mod v0.3.0/go.mod h1:s0Qsj1ACt9ePp/hMypM3fl4fZqREWJwdYDEqhRiZZUA=
 golang.org/x/mod v0.4.2 h1:Gz96sIWK3OalVv/I/qNygP42zyoKp3xptRVCWRFEBvo=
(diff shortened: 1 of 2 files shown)


jdknives:
@ersonp this solves the panic, but leaves us with another issue: 

Sometimes, the visor that got shutdown is not shown as offline and the frontend cant reload anymore. 

[log.txt](https://github.com/skycoin/skywire/files/6407739/log.txt)

ersonp:
I have managed to fix that issue by sending empty fields. But I think that's unnecessary data, and would be better if there is a fix from the UI side. Having said that, it might not make a lot of difference and we can keep using this fix.

ersonp:
@i-hate-nicknames can we use [extraSummaryResp](https://github.com/skycoin/skywire/blob/6b16ae8e0fabbc1c4d434a54e1573f289ca8d859/pkg/visor/hypervisor.go#L468-L472) with `IsHypervisor` and `DmsgStats` fields instead of using a new object [extraSummaryWithDmsgResp](https://github.com/skycoin/skywire/blob/6b16ae8e0fabbc1c4d434a54e1573f289ca8d859/pkg/visor/hypervisor.go#L492-L501).
Also, I feel like few of the method names can be changed for better clarity like: `getVisorsExtraSummary` to `getAllVisorsSummary` and `makeExtraSummaryResp` tp `generateSummary`.

jdknives:
@ersonp @i-hate-nicknames just a general comment from my side: 

While reviewing the fix from Erson, I thought that the naming for the various summaries could be clearer. Currently it looks to me like there are multiple summaries which leads us to weird constructs such as `ExtraSummaryWithDmsgResponse` etc. In this case I think it may be better to call the most expansive (informative and detailed) struct the `Summary` and call the current `Summary` object `Overview` or `Details`. Maybe there is a better option as well.

i-hate-nicknames:
Yes, feels like it definitely needs some refactoring. There are multiple levels of extra that seems to have been added over time. Visor has summary and extra summary, and hypervisor has his own extra summary, plus response wrappers around visor summaries:
hypervisor provides:
1. `visor.Summary` (via `getVisor()`), wraps in `summaryResp`
2.  `visor.ExtraSummary` (via `getVisorSummary()`), wraps in `extraSummaryResp`
3. its own extra summary (extra-extra now?) (via `getVisorsExtraSummary()`), `extraSummaryWithDmsgResp`
There are some overlaps between those structures but not always, e.g. visor.ExtraSummary Dmsg is different from Dmsg in extraSummaryWithDmsgResp. So we cannot simply embed one into other.

I would propose:
1. Remove Dmsg from visor.ExtraSummary. This field is not populated by visor api, so there is no reason to have it there.
2. Investigate why do we need dmsg stats for all visors in getVisorSummary, since we are asking for stats of a single visor.
    In case we don't need that data, extraSummaryWithDmsgResp and extraSummaryResp become equivalent and the former 
can be removed
3. Factor out TCPAddr and Online fields into its own struct, maybe smth like NetworkStats?
4. Rename summary to overview like @jdknives suggested, and extra summary to just summary

PR #753, opened 2021-05-04
Feature/makefile cleanup

### LINKED ISSUE #747: Push visor docker image to Dockerhub

alexadhy:

Fixes #747 

 Changes:	
- Makefiles cleanup (many recipes are actually for skywire-services instead of `skywire`, one such example is the `setup-node container`)
- Remove skywire-runner, as it is not being used anywhere
- Added buildinfo ldflags to docker build script 

How to test this PR:
- Follow the instruction on the README.md for running the docker container.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,10 +20,6 @@ OPTS?=GO111MODULE=on
 STATIC_OPTS?= $(OPTS) CC=musl-gcc
 MANAGER_UI_DIR = static/skywire-manager-src
 MANAGER_UI_BUILT_DIR=cmd/skywire-visor/static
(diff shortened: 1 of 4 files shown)


jdknives:
@alexadhy did you verify that `skywire-services` still work as expected (setting up integration env for instance) after removing these targets?

jdknives:
@alexadhy just verified that it all works. Good job.

jdknives (approved):

PR #759, opened 2021-05-06
Refactor Summaries

### LINKED ISSUE #757: Refactor summaries of visor
Original comment from @i-hate-nicknames :

There are multiple levels of extra that seems to have been added over time. Visor has summary and extra summary, and hypervisor has his own extra summary, plus response wrappers around visor summaries:
hypervisor provides:
1. `visor.Summary` (via `getVisor()`), wraps in `summaryResp`
2.  `visor.ExtraSummary` (via `getVisorSummary()`), wraps in `extraSummaryResp`
3. its own extra summary (extra-extra now?) (via `getVisorsExtraSummary()`), `extraSummaryWithDmsgResp`
There are some overlaps between those structures but not always, e.g. visor.ExtraSummary Dmsg is different from Dmsg in extraSummaryWithDmsgResp. So we cannot simply embed one into other.

I would propose:
1. Remove Dmsg from visor.ExtraSummary. This field is not populated by visor api, so there is no reason to have it there.
2. Investigate why do we need dmsg stats for all visors in getVisorSummary, since we are asking for stats of a single visor.
    In case we don't need that data, extraSummaryWithDmsgResp and extraSummaryResp become equivalent and the former 
can be removed
3. Factor out TCPAddr and Online fields into its own struct, maybe smth like NetworkStats?
4. Rename summary to overview like @jdknives suggested, and extra summary to just summary

ersonp:
yes

Fixes #757 

 Changes:	
- Changed `Summary` to `Overview` and `ExtraSummary` to `Summary`.
- Factored out TCPAddr and Online fields into its own struct `NetworkStats`.
- Removed `extraSummaryWithDmsgResp` and updated the new `Summary` accordingly.
- Renamed `getVisorsExtraSummary` to `getAllVisorsSummary`.
- Removed ` TCPAddr` field from `makeSummaryResp`

How to test this PR:
1. Run `make build-ui; make move-built-frontend ; make build; make install`
2. Start Hypervisor and check ui.


ersonp:
`dmsg stats` is needed by both `getVisorSummary` and `getAllVisorsSummary`.

pkg/visor/api.go
```diff
@@ -78,8 +78,8 @@ type HealthCheckable interface {
 	Health(ctx context.Context) (int, error)
 }
 
-// Summary provides a summary of a Skywire Visor.
-type Summary struct {
+// Overview provides a overview of a Skywire Visor.
```
jdknives: Maybe be bit more verbose `Overview provides a range of basic information about a Visor`

jdknives (changes requested):
Seems to me like TCP address is not actually completely removed. If it is, then we can probably get rid of the whole network stats structure because

1. `TCPAddr` is gone
2. `Online` field is not actually filled by visor iirc so it does not need to be part of the summary.

ersonp:
Actually `NetworkStats` are used in `getVisors()`(I do not know where `/visors, getVisors()` is being used for),  `getVisor()` uses the port from `TCPAddr` and `Online` from `NetworkStats` is used by `getAllVisorsSummary`.

ersonp:
`"/visors", getVisors()` and `"/visors/{pk}", getVisor()` uses `NetworkStat` but i do not know where the endpoints are used. I removed `NetworkStat` from them and everything still works but I will need more info on them. `Online` is used by `getAllVisorsSummary()` and the port in `TCPAddr` is used by `getVisorSummary` so I replaced `TCPAddr` with `Port`.  Both `getAllVisorsSummary()` and `getVisorSummary` user `Overview` so `Online` and `Port` can't be taken out.

pkg/visor/api.go
```diff
@@ -114,38 +114,41 @@ func (v *Visor) Summary() (*Summary, error) {
...
-	Routes  []routingRuleResp               `json:"routes"`
+// Summary provides detailed info including overview and health of the visor.
+type Summary struct {
+	Overview     *Overview                      `json:"overview"`
+	Health       *HealthInfo                    `json:"health"`
+	Uptime       float64                        `json:"uptime"`
+	Routes       []routingRuleResp              `json:"routes"`
+	IsHypervisor bool                           `json:"is_hypervisor,omitempty"`
+	DmsgStats    *dmsgtracker.DmsgClientSummary `json:"dmsg_stats"`
+	Port         uint16                         `json:"port"`
```
jdknives: Not sure what we display that for. Is it the `dmsg` port?

ersonp: Yes it's the `dmsg` port.

Senyoret1 (commented):
Working well, just 2 things:
- With the changes made, some parts of the UI code are no longer needed, but I checked and added them to a list, so I can make the changes after this PR is merged, to avoid delays with this one.
- I'm not a fan of updating the compiled UI files in PRs that were made for adding other changes, but if nobody else has problems with this, there is no problem.

jdknives:
> Working well, just 2 things:
> 
> * With the changes made, some parts of the UI code are no longer needed, but I checked and added them to a list, so I can make the changes after this PR is merged, to avoid delays with this one.
> * I'm not a fan of updating the compiled UI files in PRs that were made for adding other changes, but if nobody else has problems with this, there is no problem.

Nice. I usually dont like it either, but will merge this one.

jdknives (approved):

Issue #543, opened 2020-10-11
Installer for MacOS


jdknives:
**Feature description**

We will need an installer for MacOS for Skywire and the VPN. 

Requirements:
- package Visor, VPN-client, Skywire-CLI and task bar app
- run visor as root by default
- generate a hypervisor config by default

For the taskbar app this library has been simple enough to use and work with. https://github.com/getlantern/systray

PR #761, opened 2021-05-08
Systray launcher

alexadhy:

Part of #543 

 Changes:	
- Making cross-building on goreleaser works (via `golang-cross`)
- Build darwin binaries on darwin host only 

How to test this PR:

https://github.com/skycoin/skywire/blob/284937e4a966b49e7b268e2733b02f7aca67779c/docs/systray-builds.md

Running on linux:

Screen Shot 2021-07-06 at 16 13 58

Tested on ubuntu 20.04, gnome DE.

Running on Mac:

Screen Shot 2021-07-06 at 16 16 22


Makefile
```diff
@@ -104,6 +110,30 @@ format: tidy ## Formats the code. Must have goimports and goimports-reviser inst
...
+		alexadhyatma/golang-cross:$(GO_BUILDER_VERSION) --snapshot --skip-publish --rm-dist
+
+snapshot-clean: ## Cleans snapshot / release
+	rm -rf ./dist
+
+sysroot:
+	mkdir -p ./sysroot
+	@echo "getting sysroot for cross compilation"
+	if [[ ! -f /tmp/snapshot-05-12-2021.tar.gz ]]; then \
+  		curl -L -o /tmp/snapshot-05-12-2021.tar.gz "https://alexadhy-git.s3-ap-southeast-1.amazonaws.com/snapshot-05-12-2021.tar.gz"; \
```
i-hate-nicknames: Can you elaborate on these links, what is the contents? Ideally I think everything should be self-contained or rely on public tools that are maintained by someone, and I believe this is self-hosted by you?

alexadhy: It’s just a sysroot from a running ubuntu 20.04 container containing libgtk-3-dev and libappindicator3-dev and all of their dependencies: https://github.com/alexadhy/sysroot/blob/master/scripts/rsync.sh

It contains the headers, pc, .so files for arm64, armhf, and amd64.

I was hosting it with git-lfs previously but it was just too big, so for now i hosted it in my s3 bucket, would be nice to find other alternatives.

internal/gui/gui_linux.go
```diff
@@ -0,0 +1,6 @@
+//+build linux
+
+package gui
+
+// TODO (darkrengarius): change path
+const iconPath = "/Applications/Skywire.app/icon.png"
```
alexadhy: @jdknives, @i-hate-nicknames need an input on where this iconPath should be or if this should be hardcoded like this, same case in `darwin` and `windows`

jdknives:
@alexadhy I cant build with the latest commit of this PR:

[error-build.log](https://github.com/skycoin/skywire/files/6633726/error-build.log)


Regarding the location of the icon: If I am correct, the location for the `darwin` icon is pretty much predetermined in the current directory specified. Thats where the app is installed to with installer and where the system expects the icons to be as well. 

For Windows I have not idea at all. Someone needs to look into creating a Windows installer and find out where these are usually put. 

For Linux, we could search for the icon in the same place we expect the config to be. That'd be `./icon.png` for source installs and `/opt/skywire/` for package installs.

alexadhy:
> @alexadhy I cant build with the latest commit of this PR:
> 
> [error-build.log](https://github.com/skycoin/skywire/files/6633726/error-build.log)

@jdknives: For now to build it with CGO enabled, you have to run `make snapshot` as per the instruction on the top. If all is working for you, I'll change the makefile recipe for build (a bit hesitant to do this, since not all binaries required systray, dockerized one for example)

jdknives:
@alexadhy notes for this PR to be merged:

1. We need to be able to build the systray app optionally. If we make this part of the default build command, a lot of people will not be able to build anymore and we need to have an option to build with just go properly configured. 
2. The instructions for how to build need to be improved
3. When running `skywire-visor` with `--systray=true` it seems to override me passing a config path with `-c` flag for some reason.

alexadhy:
> @alexadhy notes for this PR to be merged:
> 
> 
> 
> 1. We need to be able to build the systray app optionally. If we make this part of the default build command, a lot of people will not be able to build anymore and we need to have an option to build with just go properly configured. 
>

I'll dissect it and add build tag for systray, such that if anyone wants to build it with systray enabled, He/She may continue to do so.

> 2. The instructions for how to build need to be improved
> 

> 3. When running `skywire-visor` with `--systray=true` it seems to override me passing a config path with `-c` flag for some reason. 

Ok will fix it today

mrpalide:
I reviewed codes from [`#cc98403`](https://github.com/skycoin/skywire/pull/761/commits/cc9840397164b5003ee2048ec45eec47052927bc), I don't have any problem with codes, but the problem that @jdknives mention exist yet.

> 3. When running skywire-visor with --systray=true it seems to override me passing a config path with -c flag for some reason.

ersonp:
Everything looks good except the exit button in the systray. The visor exits before all the modules can shut down. When uninstalling it shuts down all the modules just like using ctrl + c. It's a minor thing but maybe we should look at it.

mrpalide:
The shutting modules down process on `exit` button work once (also complete 8/11), but not work more.
Also for KDE desktop, we should install related packages. On Debian based distros, you can use these command:
```
sudo apt install libappindicator3-dev gir1.2-appindicator3-0.1 libgtk-3-dev pkg-config
```

jdknives (commented):
Good work. A few comments:

`make build-systray` builds `setup node` and `skywire-cli` with `-systray` which it should not. 

Documentation needs to be changed to indicate that systray app can and needs to be run with `sudo`. Running without it breaks the VPN as there is no prompt to gain root privileges for the `vpn-client`.

The systray app does not open the hypervisor in the default browser for some reason. It always seems to open up Safari on MacOS. Not sure if thats an issue with the browser launch library we are using. But I remember that in Skycoin, which uses the same library, it opens the one defined as standard. 

If the systray app is not run as package, we should not allow the uninstall option but that I would probably tackle separately together with a few other issues we need to look into to improve the task bar app. I will make a separate ticket for that.

ersonp:
@alexadhy @jdknives systray with `sudo` is not working for me.

alexadhy:
> @alexadhy @jdknives systray with `sudo` is not working for me.

right now, under `Linux`, you don't have to run it with `sudo`, just run it normally, it will ask for your sudo password, instruction:

1. Modify app launch config, specifically `sudo_required` part to `true` in `vpn-client` for example

for darwin / macos you can run the whole visor with sudo 


> sudo apt install libappindicator3-dev gir1.2-appindicator3-0.1 libgtk-3-dev pkg-config
> ```

I'll add it to the docs

> The shutting modules down process on `exit` button work once (also complete 8/11), but not work more.

I'll look into it.

> The systray app does not open the hypervisor in the default browser for some reason. It always seems to open up Safari on MacOS. Not sure if thats an issue with the browser launch library we are using. But I remember that in Skycoin, which uses the same library, it opens the one defined as standard.

hmm, I have a weird case which there are 2 browsers opened when I run it (set `firefox` as default browser). I'll look into it.

ersonp:
@alexadhy VPN is working on linux. The prompt for the password comes up just as intended. Also the `sudo_required` for `vpn_client` should be true by default.

ersonp:
The vpn cannot be stopped from the UI. 
```
[2021-08-16T16:17:53+05:30] WARN StopApp [launcher]: Failed to stop app. app_name="vpn-client" error="operation not permitted"
```

docs/systray-builds.md
```diff
@@ -0,0 +1,78 @@
...
+To build `skywire-visor` with systray feature enabled, you have to had these installed on your system:
+
+### Prequisites
+
+#### Linux
+
+- Debian / Ubuntu and its derivations
+
+```bash
+$ sudo apt install libgtk-3-dev libappindicator-3-dev
```
ersonp: `libappindicator3-dev` is the correct name of the package.

pkg/skyenv/values_windows.go
```diff
@@ -0,0 +1,13 @@
+//+build windows
+
+package skyenv
+
+const (
+	// TODO (darkrengarius): change path
+	packageSkywirePath = "/opt/skywire"
```
ersonp, line 7: We should probably change this to `C:` drive.

alexadhy: this doesn't currently run on windows, and I have no machine to test it. Hence, I'm going to propagate this issue to the Windows Build PR. Thanks for the suggestion.

.goreleaser.yml
```diff
@@ -163,16 +152,14 @@ builds:
...
-        goarch: arm64
     env:
       - CGO_ENABLED=0
     main: ./cmd/apps/vpn-client/
     ldflags: -s -w -X github.com/skycoin/dmsg/buildinfo.version={{.Version}} -X github.com/skycoin/dmsg/buildinfo.commit={{.ShortCommit}} -X github.com/skycoin/dmsg/buildinfo.date={{.Date}}
 archives:
   - format: tar.gz
     wrap_in_directory: false
-    name_template: 'skywire-v{{ .Version }}-{{ .Os }}-{{ .Arch }}{{ if .Arm }}v{{ .Arm }}{{ end }}'
+    name_template: 'skywire-{{ .Version }}-{{ .Os }}-{{ .Arch }}{{ if .Arm }}v{{ .Arm }}{{ end }}'
```
jdknives: Why are we changing the naming conventions here? That will break updating.

.goreleaser.yml
```diff
@@ -31,12 +31,11 @@ builds:
     ignore:
       - goos: darwin
         goarch: 386
-      - goos: darwin
```
jdknives, line 34: Why are we removing the `arm64` architecture? We should release for `arm64` as I believe Apple will or has already switched over their architecture for newer machines.

.goreleaser-systray.yml
```diff
@@ -0,0 +1,246 @@
...
+    env:
+      - CGO_ENABLED=0
+    main: ./cmd/apps/vpn-client/
+    ldflags: -s -w -X github.com/skycoin/dmsg/buildinfo.version={{.Version}} -X github.com/skycoin/dmsg/buildinfo.commit={{.ShortCommit}} -X github.com/skycoin/dmsg/buildinfo.date={{.Date}}
+    flags:
+      - -tags='systray'
+archives:
+  - format: tar.gz
+    wrap_in_directory: false
+    name_template: 'skywire-{{ .Version }}-{{ .Os }}-{{ .Arch }}{{ if .Arm }}v{{ .Arm }}{{ end }}'
```
jdknives: I would keep the naming convention we currently use for regular releases and just add a `systray` label so we would have 

`skywire-systray-v0.4.2-linux-amd64`

.travis.yml
```diff
@@ -14,6 +14,10 @@ addons:
     packages:
       # For building MUSL static builds on Linux.
       - musl-tools
+      #- gcc-mingw-w64-i686
```
jdknives: Do we need these packages in Travis for the systray building?

Makefile
```diff
@@ -100,13 +108,47 @@ format: tidy ## Formats the code. Must have goimports and goimports-reviser inst
 dep: tidy ## Sorts dependencies
 	${OPTS} go mod vendor -v
 
+snapshot-systray: sysroot ## create snapshot release
```
jdknives, line 113: What purpose do the `snapshot` targets have? Are you using them for local testing/debugging only?

Makefile
jdknives: Will host that ourselves in an S3 like bucket. Will provide details early next week.

Makefile
```diff
@@ -134,12 +181,21 @@ build-deploy: ## Build for deployment Docker images
...
+	docker run --rm --privileged \
+		-v $(CURDIR):/go/src/github.com/skycoin/skywire \
+		-v /var/run/docker.sock:/var/run/docker.sock \
+		-v $(GOPATH)/src:/go/src \
+		-v $(CURDIR)/sysroot:/sysroot \
+		-w /go/src/github.com/skycoin/skywire \
+		alexadhyatma/golang-cross:$(GO_BUILDER_VERSION) -f /go/src/github.com/skycoin/skywire/.goreleaser-systray.yml --rm-dist
 
-github-release: ## Create a GitHub release
+github-release:
```
jdknives: Thinking about this, I believe we can and should move both release targets and goreleaser files together. Going forward, we will release systray and regular binaries for every release so I dont think we need to have separate files and targets.

cmd/skywire-cli/commands/visor/gen-config.go
```diff
@@ -38,6 +45,12 @@ func init() {
 	genConfigCmd.Flags().BoolVarP(&testEnv, "testenv", "t", false, "whether to use production or test deployment service.")
 	genConfigCmd.Flags().BoolVar(&hypervisor, "is-hypervisor", false, "whether to generate config to run this visor as a hypervisor.")
 	genConfigCmd.Flags().StringVar(&hypervisorPKs, "hypervisor-pks", "", "public keys of hypervisors that should be added to this visor")
+	genConfigCmd.Flags().BoolVar(&hasVPNClient, "has-vpn-client", false, "generate config for VPN client")
```
jdknives: I assume you want to add all of the app-config flags to be able to just generate config for VPN-client for packages and systray builds. I think it may be preferable to just add one more flag for systray specifc config instead of 6 new flags. I do not see much use for the app specific flags apart from the systray config generation and we keep the logic and codebase a bit cleaner imo.

internal/gui/gui.go
```diff
@@ -0,0 +1,231 @@
...
+		initOpenHypervisorBtn(conf)
+		initUninstallBtn()
+		initQuitBtn()
+
+		go handleUserInteraction(conf)
+	}
+}
+
+// OnGUIQuit is executed on GUI exit.
+func OnGUIQuit() {
```
jdknives, line 68: Is there anything specific we should do here? Could not find any place where we use it.

internal/gui/gui.go
```diff
@@ -0,0 +1,231 @@
...
+	stopVisorFnMx.Unlock()
+
+	if stop != nil {
+		stop()
+	}
+}
+
+func isHypervisorRunning(addr string) bool {
+	// we check if it's up by querying `health` endpoint
+	resp, err := http.Get(addr + "/api/health")
```
jdknives, line 279: Maybe we should use `http.Head` here so we dont need to read from the response?

internal/gui/gui_darwin.go
```diff
@@ -0,0 +1,8 @@
+//+build darwin,systray
+
+package gui
+
+const (
+	iconPath        = "/Applications/Skywire.app/Contents/Resources/tray_icon.tiff"
```
jdknives, line 14: Is it possible to use a regular `png` or `jpg` here? Just curious as Tiff is an annoying format iirc.

pkg/util/stdio/arm_dup.go
```diff
@@ -7,5 +7,5 @@ import "syscall"
 // DupTo duplicates old fd into the new fd
 // see dup2 and dup3 system calls
 func DupTo(oldfd, newfd int) error {
-	return syscall.Dup3(oldfd, newfd, 0)
+	return syscall.Dup2(oldfd, newfd)
```
jdknives, line 10: Why are we switching back here? I do not recall all the context of the original implementation of the call here.

pkg/util/stdio/arm_nondarwin.go
```diff
@@ -0,0 +1,11 @@
+// +build linux,arm64 windows
```
jdknives, line 1: Is there any use for Windows on arm? Never heard of it to be honest.

pkg/visor/init_cli.go
```diff
@@ -0,0 +1 @@
+package visor
```
jdknives: Why do we require this file and the `init_cli.go` pendant?

pkg/visor/visorconfig/config.go
```diff
@@ -158,14 +173,18 @@ func MakePackageConfig(log *logging.MasterLogger, confPath string, sk *cipher.Se
...
+	conf.LocalPath = skyenv.PackageAppLocalPath()
+	conf.Launcher.BinPath = skyenv.PackageAppBinPath()
 
 	if conf.Hypervisor != nil {
 		conf.Hypervisor.EnableAuth = skyenv.DefaultEnableAuth
 		conf.Hypervisor.EnableTLS = skyenv.PackageEnableTLS
-		conf.Hypervisor.TLSKeyFile = skyenv.PackageTLSKey
-		conf.Hypervisor.TLSCertFile = skyenv.PackageTLSCert
+		if runtime.GOOS == "darwin" {
+			// disable TLS by default for OSX
```
jdknives: Why are we disabling TLS on OSX?

jdknives (commented):
Well done. Just a few small comments and questions from my side.

.goreleaser.yml
alexadhy: because the generated package name is `skywire-vv4.0.0-Linux-amd64` instead of `skywire-v4.0.0` i tests, not sure if that's what is intended, i'll change it back to it if required.

.goreleaser.yml
alexadhy: that's the ignore part, I think I added arm64 build for Apple M1 (no longer ignored arm64)

.travis.yml
alexadhy: not sure, that was added in https://github.com/skycoin/skywire/commit/16623503cf4b23aef2a32a0a6b530eb59c827839, have to investigate further if it's indeed needed.

Makefile
alexadhy: yes, it's only for local test / debugging purpose only, or only to build goreleaser package locally.

internal/gui/gui_darwin.go
alexadhy: Good point, I can just convert the hi-res png to tiff format via library.

pkg/util/stdio/arm_nondarwin.go
alexadhy: There's an emulation layer for devices running windows on arm64 (such as `surface pro x`) to run x64 binaries, but the latest windows (10 & 11) does support devices on arm64, not quite sure if it's needed. Adding it only if it is.

pkg/visor/init_cli.go
alexadhy: good catch, It was an accidental committed file, I'll remove it.

pkg/util/stdio/arm_dup.go
alexadhy: just that Dup2 is not supported on all platforms, i.e. windows

Makefile
```diff
@@ -20,6 +22,7 @@ DMSG_BASE := github.com/skycoin/dmsg
 OPTS?=GO111MODULE=on
 STATIC_OPTS?= $(OPTS) CC=musl-gcc
 MANAGER_UI_DIR = static/skywire-manager-src
+GO_BUILDER_VERSION=v1.16.4
```
mrpalide: It's not specific for skywire, but may we should use v1.17 from now or not? @jdknives

internal/gui/gui.go
mrpalide: It's on `systray.go` on skywire-visor commands [link](https://github.com/alexadhy/skywire/blob/f4489dd7d02a06e9768c328fe227891d07f23a6d/cmd/skywire-visor/commands/systray.go#L45)

internal/gui/gui.go
```diff
@@ -0,0 +1,322 @@
...
+}
+
+// OnGUIQuit is executed on GUI exit.
+func OnGUIQuit() {
+
+}
+
+// ReadSysTrayIcon reads system tray icon.
+func ReadSysTrayIcon() ([]byte, error) {
+	if err := preReadIcon(); err != nil {
```
mrpalide, line 74: `preReadIcon` seems not declared.

internal/gui/gui.go
```diff
@@ -0,0 +1,322 @@
...
+		return nil, err
+	}
+	contents, err := ioutil.ReadFile(iconPath)
```
mrpalide, line 77: `iconPath` seems not declared.

internal/gui/gui.go
```diff
@@ -0,0 +1,322 @@
...
+		return
+	}
+	if cond {
+		mOpenHypervisor.Disable()
+		mUninstall.Disable()
+		mQuit.Disable()
+
+		stopVisor()
+
+		if err := platformExecUninstall(); err != nil {
```
mrpalide, line 258: `platformExecUninstall` seems not declared.

pkg/util/stdio/arm_nondarwin.go
```diff
@@ -0,0 +1,11 @@
...
+
+package stdio
+
+import "syscall"
+
+// DupTo duplicates old fd into the new fd
+// see dup2 and dup3 system calls
+func DupTo(oldfd, newfd int) error {
+	return syscall.Dup3(oldfd, newfd, 0)
```
mrpalide, line 10: Should we use `Dup2` here again or not?

cmd/skywire-cli/commands/visor/init_windows.go
```diff
@@ -0,0 +1,7 @@
+//+build windows
+
+package visor
+
+func appendExecPlatform(envs []string) {
```
mrpalide: What is this function exactly? I didn't see used anywhere.

pkg/visor/init_windows.go
```diff
@@ -12,3 +12,7 @@ func initDmsgpty(ctx context.Context, log *logging.Logger) error {
 	log.Error("dmsgpty is not supported on windows.")
 	return nil
 }
+
+func appendExecPlatform(envs []string) {
```
mrpalide: What is this function exactly? I didn't see used anywhere. (Same as `init_windows.go` in `skywire-cli` command visor)

mrpalide (commented):
Same as my old mistake, seems adding new line at end of codes is off on your IDE.
And just some little comments.

mrpalide:
I want to open browser from systray, but I get following error.
```
[2021-08-23T22:35:42+04:30] INFO []: Opening hypervisor at http://localhost:8000
[2021-08-23T22:35:42+04:30] ERROR []: Failed to open hypervisor error="failed to open link: webbrowser: can't open browser"
```
I'm on KDE desktop with Firefox default browser.


------------------------------
Other thing that I see is a sudden interruption without completing the shutdown process.

internal/gui/gui.go
alexadhy: It is declared, try build it. `OnGUIQuit` is a function that will be called on `GUI` quit, it's empty right now. It will be populated once for example, we stop polling available vpn server, or doing rpc calls to check the stats of vpn connection, or to stop the vpn connection altogether when we quit the gui.

internal/gui/gui.go
alexadhy: it is.

internal/gui/gui.go
alexadhy: please see these files: `gui_linux.go` or `gui_darwin.go` or `gui_windows.go` for declaration for each platform. Same as `iconPath` and `preReadIcon` function above.

mrpalide: Yes, they declare there.

pkg/util/stdio/arm_nondarwin.go
alexadhy: There's literally no difference between both calls, see [http://manpages.ubuntu.com/manpages/trusty/man2/dup.2.html](dup2 and dup3), and `Dup3` is available for windows as well, hence I chose `Dup3`.

pkg/visor/init_windows.go
alexadhy: good catch, I'll delete it. Thanks man

alexadhy:
> I want to open browser from systray, but I get following error.
> 
> ```
> [2021-08-23T22:35:42+04:30] INFO []: Opening hypervisor at http://localhost:8000
> [2021-08-23T22:35:42+04:30] ERROR []: Failed to open hypervisor error="failed to open link: webbrowser: can't open browser"
> ```
> 
> I'm on KDE desktop with Firefox default browser.
> 
> Other thing that I see is a sudden interruption without completing the shutdown process.

Have you set default browser in `xdg-settings`?

I'll fix shutdown process today.

pkg/util/stdio/arm_nondarwin.go
mrpalide: Oh! Really? I see you said to Rudi that Dup3 not support windows at [this](https://github.com/skycoin/skywire/pull/761#discussion_r693944837) comment and ask this question. Sorry for my mistake.

mrpalide:
> 
> Have you set default browser in `xdg-settings`?
> 
> I'll fix shutdown process today.
1. I didn't set with `xdg-settings` in terminal and just use Default Application in System Setting Module. I check it and xdg-settings is firefox too.
   ```
   mohammadreza@workstation:~$ xdg-settings get default-web-browser
   firefox.desktop
   ```
2. Great.

alexadhy:
https://user-images.githubusercontent.com/72057554/130661616-ef4fd1f9-9ac4-44df-82bd-b2e5668a7c22.mov

> > Have you set default browser in `xdg-settings`?
> > I'll fix shutdown process today.
> 
>     1. I didn't set with `xdg-settings` in terminal and just use Default Application in System Setting Module. I check it and xdg-settings is firefox too.
>        ```
>        mohammadreza@workstation:~$ xdg-settings get default-web-browser
>        firefox.desktop
>        ```
> 
>     2. Great.

will test on KDE, and see whether it works. Works on KDE Plasma latest, on ubuntu 20.04 see attached video

pkg/util/stdio/arm_nondarwin.go
jdknives: @alexadhy does not hurt to support it if it is an option. Lets keep it.

pkg/util/stdio/arm_dup.go
jdknives: @alexadhy I see. In this case, can you check that the if we update a visor the output of the update still gets copied over correctly and displayed correctly in the terminal on Linux or Mac? I remember there was a reason we had to switch initially.

.goreleaser.yml
jdknives: @alexadhy not intended. `skywire-v0.4.0` is intended. I just missed that the `v` will be incldued by `.Version`.

internal/gui/gui.go
jdknives: @mrpalide I see. Seems to be just a requirement from the libary/.

Makefile
jdknives: @mrpalide I havent checked out go 1.17. Not sure if there are relevant improvements for us or new features. If not, I guess we can stick with 1.16 for a while.

.travis.yml
jdknives: @alexadhy no worries then. Would just get rid of it.

PR #777, opened 2021-05-27
Fix/shutdown data races

### LINKED ISSUE #774: Race condition in snet and visor
**Describe the bug**
Race conditions are observed when a visor built with `-race` flag is closed with `Ctrl+C`.

**Environment information:**

- OS: Linux
- Platform: Linux 5.8.0-50-generic x86_64

**Steps to Reproduce**

1. Start a visor
2. Close the visor with `Ctrl+C`
3. See data race

**Actual behavior**
[log.txt](https://github.com/skycoin/skywire/files/6533650/log.txt)
[log2.txt](https://github.com/skycoin/skywire/files/6538130/log2.txt)

**Expected behavior**
Should close without data race.

i-hate-nicknames:

Fixes #774

 Changes:	
- Fix data races when manually shutting down visor

How to test this PR:
Run visor, wait until initialization finishes and then hit Ctrl+C

Note: I removed synchronization in router shutdown: [commit](https://github.com/skycoin/skywire/commit/15e0ebca5c4405f3996c7d5194af69a45966ffbb). There was a waitgroup that waits until router `serveSetup` method exists completely  that was causing data race. Apparently, you cannot use `Add` and `Wait` wg methods in different goroutines. Not sure if we need this synchronization there, but should be possible to achieve similar result with a channel. 

Another issue is that visor exits with status 1, i.e. error, so the shutdown process is not considered successful. This is the observable behavior on develop and wasn't introduced by this PR.


jdknives (approved):
Looks good to me!

alexadhy (approved):
Nice work, working fine here.

PR #729, opened 2021-03-30
Add router related config to the manager

Senyoret1:
Did you run `make format && make check`?
The go code was not changed. `npm run lint` and `npm run build` were used.

 Changes:	
- Now the min and max hops are shown in the right bar of the visor details page. Also, under the data there is a button for being able to change the values.

How to test this PR:
Enter to the visor details page and check the data in the right bar. For changing the values, press the `Change configuration` button.

Note: the PR will be finished after the API endpoints for getting and setting the data are added.

PR #780, opened 2021-05-28
Add min hops manipulation endpoint

### LINKED ISSUE #770: Implement `min hops` hypervisor endpoint
**Feature description**

We need to implement the endpoint for getting and setting the visor config min hops value as requested and implemented (frontend) here:

- https://github.com/skycoin/skywire/pull/729

alexadhy:

Fixes #770 

Related to:
- #729 

 Changes:	
- Added `min_hops` endpoint to the http API and rpc one (@jdknives @Senyoret1 not sure about the latter, since there's `route-finder`)

How to test this PR:
- run it in the integration test (using v111 config)
- do an http call to the `/visors/{pk}/min-hops` endpoint (`POST` and `GET`) like:
```bash
$ curl -k https://localhost:8000/api/visors/${PK}/min-hops
$ curl -k -d '{"min_hops": 1}' -H "Content-Type: application/json" -X POST  https://localhost:8000/api/visors/${PK}/min-hops
```
- or do an rpc call via `skywire-cli visor route get-min-hops` and `skywire-cli visor route set-min-hops <some_number>`

### DIFF
--- a/cmd/skywire-cli/commands/visor/routes.go
+++ b/cmd/skywire-cli/commands/visor/routes.go
@@ -22,6 +22,8 @@ func init() {
 		ruleCmd,
 		rmRuleCmd,
 		addRuleCmd,
+		getMinHopsCmd,
+		setMinHopsCmd,
 	)
 }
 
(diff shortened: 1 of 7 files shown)


cmd/skywire-cli/commands/visor/routes.go
```diff
@@ -22,6 +22,8 @@ func init() {
 		ruleCmd,
```
jdknives: I dont see a point for us to add get/set commands to the cli. The cli needs to run locally so the user might as well just change the config value himself. The http endpoint is needed to change this value remotely mostly.

jdknives (approved):
Not sure if we should add the CLI commands just to be consistent. I cant see a use for it effectively so we may get rid of it. But also does not hurt to have it.

i-hate-nicknames (approved):

Senyoret1 (commented):
I have 2 comments about the implementation:
- It would be good if the `GET /api/visors/{pk}/summary` API endpoint returns the data, that way the UI would not have to make an additional API call, which would make the UI a bit slower.
- Is there also a max hops configuration? If that is the case, it sould be added to the API too.

alexadhy:
@Senyoret1:

- first point makes sense, done.
- currently there's no max hops configuration

Senyoret1 (approved):
Great, I implemented the changes in https://github.com/skycoin/skywire/pull/729, thanks 👍

PR #784, opened 2021-05-31
Fix a bug in the app list

Senyoret1:

 Changes:	
- Currently, if a visor has no arguments for the skychat app, it is not possible to use the Skywire manager for seeing the app list, as there is an error. This PR solves that problem.

- When a visor does not have a port explicitly defined for the skychat app in the config file, the link shown in the Skywire manager for showing the skychat UI does not work. This PR solves that problem.

How to test this PR:
Use the Skywire manager to open a visor without any argument for the skychat app and open the app list, it should be displayed normally. Also, press the icon in the right part of the table to open the skychat UI, it should be openned normally.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ before_install:
   - nvm install 10.16
 
 install:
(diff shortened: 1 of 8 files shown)


Senyoret1:
In the last commits, the changes made in https://github.com/skycoin/skywire/pull/780 were added to this PR, as are nedded for the Travis tests.

PR #794, opened 2021-06-04
Feature/docker refactorings

### LINKED ISSUE #793: Docker Issues
**Describe the bug**

- ~~Docker Images are not being pushed to the registry (dockerhub) on push to master or develop, have checked and it is because of incorrect docker credentials (secrets) in travis-ci~~ DONE
- Visor image should not hardcode the filepath used to find config

** Possible Implementation**

- Fix docker entrypoint to use non hardcoded config path, or only use it if no config is specified in the args.

alexadhy:

Fixes #793 

 Changes:	
- Refactor docker image according to the issue

How to test this PR:
- See README on how to run skywire-visor

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,7 @@ VERSION := $(shell git describe)
 RFC_3339 := "+%Y-%m-%dT%H:%M:%SZ"
 DATE := $(shell date -u $(RFC_3339))
 COMMIT := $(shell git rev-list -1 HEAD)
+BRANCH := $(shell git rev-parse --abbrev-ref HEAD)
 
 PROJECT_BASE := github.com/skycoin/skywire
 DMSG_BASE := github.com/skycoin/dmsg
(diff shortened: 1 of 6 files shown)


jdknives:
@alexadhy running without volume defined does not work for me: 

[log.txt](https://github.com/skycoin/skywire/files/6615365/log.txt)

jdknives:
I am still running into the same problem as before. In addition, can we add a make target for building the docker image? 

[log.txt](https://github.com/skycoin/skywire/files/6626263/log.txt)

alexadhy:
@jdknives: I think it was because you built the image with `git rev-parse --abbrev-ref HEAD` tag while running it with `skywire:test` tag, the docker image tag should be `latest` for master branch, `test` for develop branch, and the same branch name if it's other branch.

Should I change that behaviour?

jdknives:
@alexadhy works well. We should also copy over `Skychat` and the `skysocks` app to the container. Afterwards this can be merged.

Makefile
```diff
@@ -138,6 +138,9 @@ build-deploy: ## Build for deployment Docker images
 	${OPTS} go build ${BUILD_OPTS_DEPLOY} -o /release/apps/skysocks ./cmd/apps/skysocks
 	${OPTS} go build ${BUILD_OPTS_DEPLOY} -o /release/apps/skysocks-client ./cmd/apps/skysocks-client
 
+build-docker: ## Build docker image
+	./ci_scripts/docker-push.sh -t develop -b
```
jdknives: Can we use a var here instead of hardcoding the tag? May be more helpful this way.

jdknives (approved):

PR #810, opened 2021-06-10
Change perm of pid and logstore

### LINKED ISSUE #808: Apps-pid.txt does not get removed and has wrong permissions
The `apps-pid.txt` file is not deleted and is created with the old permissions, meaning sudo is required to remove it which is problem when for instance, someone wants to run `go mod vendor` or the docker build script. 

We should check if there is a reason we have a file instead of in memory implementation and we should change permissions if we need a file.

ersonp:
Yes

Fixes #808 

 Changes:	
- Changed `ownerRW` to `ownerRWOtherRW`
- Changed file perm of `pidFile`, `ownerRWOtherRW` and `NewBBoltLogStore`  from `600` to `606`
- Added `make run`
- Combined `make move-built-frontend` and `make build-ui` into `make build-ui`.

How to test this PR:
1. Delete local folder `sudo rm -rf local/`
2. Run visor `sudo ./skywire-visor skywire-config.json`
3. Ctrl+C
4. Run `./ci_scripts/docker-push.sh -t $(git rev-parse --abbrev HEAD) -b`
5. No file permission error should be observed

### DIFF
--- a/pkg/util/pathutil/util.go
+++ b/pkg/util/pathutil/util.go
@@ -11,7 +11,7 @@ import (
 
 const (
 	tmpSuffix      = ".tmp"
-	ownerRW        = 0600
+	ownerRW        = 0606
 	userRWXGroupRX = 0750
 )
 
--- a/pkg/util/pathutil/util.go
+++ b/pkg/util/pathutil/util.go
@@ -11,7 +11,7 @@ import (
 
 const (
 	tmpSuffix      = ".tmp"
-	ownerRW        = 0606
+	ownerRWOtherRW = 0606
 	userRWXGroupRX = 0750
 )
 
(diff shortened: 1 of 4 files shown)


pkg/util/pathutil/util.go
```diff
@@ -11,7 +11,7 @@ import (
 
 const (
 	tmpSuffix      = ".tmp"
-	ownerRW        = 0600
+	ownerRW        = 0606
```
jdknives: The const name seems misleading, given the change we are making here.

jdknives (commented):
Looking good moslty.

pkg/util/pathutil/util.go
ersonp: Changed to `ownerRWOtherRW`

jdknives (approved):

PR #811, opened 2021-06-10
Feature/snet rewrite

### LINKED ISSUE #760: Review `snet` implementation
**Describe the feature**

While reviewing #724 it was discovered that a lot of modules depend on `snet` module for working, which may indicate the module implementation could be improved. We should review the implementation and check if we can decouple snet from other modules a bit more.

i-hate-nicknames:
Rewriting network initialization to gain init performance, as well as cleaner code and better organization in snet and transport packages.

not yet

Fixes #760

 Changes:	
-	

How to test this PR:


pkg/snet/snettest/env.go
```diff
@@ -1,23 +1,12 @@
 package snettest
 
+// todo: this seems like an integration test, we set up a real dmsg server here
```
jdknives: I would just pull this out together with all other tests in here that are effectively integration tests. We should run them as docker tests in services imo.

pkg/visor/init.go
```diff
@@ -180,37 +181,34 @@ func initDiscovery(ctx context.Context, v *Visor, log *logging.Logger) error {
...
-		ARClient:       v.arClient,
-		NetworkConfigs: nc,
-	}
-
-	n, err := snet.New(conf, v.ebc, v.MasterLogger())
-	if err != nil {
-		return err
+func initDmsg(ctx context.Context, v *Visor, log *logging.Logger) error {
+	if v.conf.Dmsg == nil {
+		// todo: this is preserved behavior from snet. Do we need it?
```
jdknives: We should just stop execution here. Without dmsg configured there is virtually nothing a visor can be used for.

jdknives (commented):
Overall good job. I think the naming renaming of certain structures and usage of `dmsgC` instead of generic `snet` structures is helpful. In general the refactor seems to go in the right direction and works. I have not checked if the performance is significantly changed but that is not very important anyways. 

Two comments:

1. I am not yet sure if there is more work to be done on structs we define to wrap connections yet. If I see correctly we have a `type Conn interface` and a `type Client interface`. The naming of Conn may be somewhat confusing as it is a wrapper for a `net.Conn` and maybe should be called `Transport`? I am also thinking about how to simplify each of these structures (including managed transport). 

One of the ideas I had and discussed with Synth was to get rid of redialing and status logic as described in #830 

Will have a look and see if there are further ways to reduce complexity in these interfaces and structures. 

2. All of the tests resembling integration tests should be torn our here and moved to the services repo. There is an open PR where we can just comment on with tests that should be redone there with docker setup.

cmd/skywire-cli/commands/visor/transports.go
```diff
@@ -113,15 +113,15 @@ var addTpCmd = &cobra.Command{
...
 		} else {
-			transportTypes := []string{
-				tptypes.STCP,
-				tptypes.STCPR,
-				tptypes.SUDPH,
+			transportTypes := []network.Type{
+				network.STCP,
+				network.STCPR,
+				network.SUDPH,
 				dmsg.Type,
```
ersonp: Imo it would be better to either use `dmsg.Type` or `network.DMSG` everywhere.

PR #813, opened 2021-06-17
Fix stcpr transport establishment issues

### LINKED ISSUE #806: stcpr transport cannot be established after being down
**Describe the bug**
Happened when running visor+vpn-server on a separate machine from vpn-client

**Environment information:**
client:
Linux 4.19.0-13-amd64 x86_64

server:
Linux 5.12.4-arch1-2 x86_64

**Steps to Reproduce**
1. started vpn client. Client repeats an error that there is no transport found:

```
[2021-06-09T16:49:24+03:00] INFO [proc:vpn-client:ac1ffcfbcf5d45c7ade1404b70761093]: Request processed. _elapsed="636.214461ms" _method="Dial" _received="4:49PM" error="route finder: transport not found" input=029081ab2f1a9087f7e59b3b58ca2959530335f861725d31aac1b5a3aff150f832:44 output=&{ConnID:0 LocalPort:0}
```

2. start vpn server, add transport manually (no entry before that). vpn works

3. stop client manually and start again. Same error as at (1), no transport found (even though ls-tp shows it's there, but it's no in up state).

4. calling add-tp returns the same transport (same ID too) entry with error that transport established but not up. 

```
[2021-06-09T16:48:36+03:00] FATAL [skywire-cli]: Established stcpr transport to 029081ab2f1a9087f7e59b3b58ca2959530335f861725d31aac1b5a3aff150f832 with ID 7f6c3a10-855a-051f-885d-6eeeb007b0e9, but it isn't up
```

Сalling rm-tp and then add-tp shows the same error. Restarting visor with vpn client doesn't help, with and without removing transport.

5. However, the following sequence works:
- Remove transport
- Stop vpn server
- Start vpn server
- Add transport

**Actual behavior**
1. Transport is not up upon visor start
2. Removing transport and establishing it again doesn't help, so client has no control over fixing process.

**Expected behavior**
1. Transport should become Up when both ends of it are online.
2. Transport should be correctly established to Up state when removed and added again

i-hate-nicknames:

Fixes #806

 Changes:	
- stcp and stcpr single redial try now has a timeout of 3 seconds
- changed transport entry structure: first key is original connection initiator
- connection initiator is now responsible for redialing
- added error responses to transport handshake
- prevent visors that do not have public IP to perform stcpr bind operation

How to test this PR:
Run two visors v1 and v2. v1 should have public IP, v2 shouldn't. Public Key of v1 should be less than Public Key of v2. 
Set up an stcpr transport from v2 to v1 manually, then stop v2. Restart v2, remove transport and establish it again. Transport should be added in UP state.
Stop v2 again. Start it again, transport should automatically switch to UP state on both ends.

### DIFF
--- a/pkg/transport/entry.go
+++ b/pkg/transport/entry.go
@@ -60,13 +60,6 @@ func MakeEntry(initiator, target cipher.PubKey, tpType string, public bool, labe
 	return entry
 }
 
-// SetEdges sets edges of Entry
-func (e *Entry) SetEdges(initiator, target cipher.PubKey) {
-	e.ID = MakeTransportID(initiator, target, e.Type)
-	e.Edges[0] = initiator
-	e.Edges[1] = target
-}
-
 // RemoteEdge returns the remote edge's public key.
 func (e *Entry) RemoteEdge(local cipher.PubKey) cipher.PubKey {
 	for _, pk := range e.Edges {
--- a/pkg/transport/entry.go
+++ b/pkg/transport/entry.go
@@ -33,7 +33,8 @@ type Entry struct {
 	// ID is the Transport ID that uniquely identifies the Transport.
 	ID uuid.UUID `json:"t_id"`
 
-	// Edges contains the public keys of the Transport's edge nodes (should only have 2 edges and the least-significant edge should come first).
+	// Edges contains the public keys of the Transport's edge nodes
+	// (should only have 2 edges and the first edge is transport original initiator).
 	Edges [2]cipher.PubKey `json:"edges"`
 
 	// Type represents the transport type.
@@ -47,20 +48,23 @@ type Entry struct {
 }
 
 // MakeEntry creates a new transport entry
-func MakeEntry(pk1, pk2 cipher.PubKey, tpType string, public bool, label Label) Entry {
-	return Entry{
-		ID:     MakeTransportID(pk1, pk2, tpType),
-		Edges:  SortEdges(pk1, pk2),
+func MakeEntry(initiator, target cipher.PubKey, tpType string, public bool, label Label) Entry {
+	entry := Entry{
+		ID:     MakeTransportID(initiator, target, tpType),
 		Type:   tpType,
 		Public: public,
 		Label:  label,
 	}
+	entry.Edges[0] = initiator
+	entry.Edges[1] = target
+	return entry
 }
 
 // SetEdges sets edges of Entry
-func (e *Entry) SetEdges(localPK, remotePK cipher.PubKey) {
-	e.ID = MakeTransportID(localPK, remotePK, e.Type)
-	e.Edges = SortEdges(localPK, remotePK)
+func (e *Entry) SetEdges(initiator, target cipher.PubKey) {
+	e.ID = MakeTransportID(initiator, target, e.Type)
+	e.Edges[0] = initiator
+	e.Edges[1] = target
 }
 
 // RemoteEdge returns the remote edge's public key.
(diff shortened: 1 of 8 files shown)


pkg/transport/entry.go
```diff
@@ -47,20 +48,23 @@ type Entry struct {
...
+	entry.Edges[0] = initiator
+	entry.Edges[1] = target
+	return entry
 }
 
 // SetEdges sets edges of Entry
-func (e *Entry) SetEdges(localPK, remotePK cipher.PubKey) {
-	e.ID = MakeTransportID(localPK, remotePK, e.Type)
-	e.Edges = SortEdges(localPK, remotePK)
+func (e *Entry) SetEdges(initiator, target cipher.PubKey) {
```
jdknives: Where are we using this function? I see it being used in tests but seems to be absent from codebase otherwise. Do we need it?

pkg/transport/entry.go
```diff
@@ -47,20 +48,23 @@ type Entry struct {
 }
 
 // MakeEntry creates a new transport entry
-func MakeEntry(pk1, pk2 cipher.PubKey, tpType string, public bool, label Label) Entry {
-	return Entry{
-		ID:     MakeTransportID(pk1, pk2, tpType),
-		Edges:  SortEdges(pk1, pk2),
+func MakeEntry(initiator, target cipher.PubKey, tpType string, public bool, label Label) Entry {
+	entry := Entry{
+		ID:     MakeTransportID(initiator, target, tpType),
```
jdknives, line 53: If I read this correctly, we may well end up with a scenario where we have two entries with the same ID, given that we still sort by  PubKey value for generating IDs but create entries by initiator. So if two visors established transports to each other, the ID would be the same. 
1. May that create trouble getting transports by ID from transport discovery? 
2. Are we currently checking (and preventing) a transport to be created if a transport of the same type between the same two edges already exists?

pkg/transport/handshake.go
```diff
@@ -110,23 +123,33 @@ func MakeSettlementHS(init bool) SettlementHS {
 		if _, err := io.ReadFull(conn, accepted); err != nil {
 			return fmt.Errorf("failed to read response: %w", err)
 		}
-		if accepted[0] == 0 {
+		switch hsResponse(accepted[0]) {
```
jdknives, line 126: That will break settlement between older and newer visors, correct?

pkg/transport/managed_transport.go
```diff
@@ -36,10 +36,11 @@ var (
...
 const (
-	tpInitBO = time.Millisecond * 500
-	tpMaxBO  = time.Minute
-	tpTries  = 0
-	tpFactor = 2
+	tpInitBO  = time.Millisecond * 500
+	tpMaxBO   = time.Minute
+	tpTries   = 0
+	tpFactor  = 2
+	tpTimeout = time.Second * 3 // timeout for a sigle try
```
jdknives: `single`

pkg/transport/handshake.go
i-hate-nicknames: No, it shouldn't. 0 is still success and 1 is a generic error like it was before, I just added more values and send those values

pkg/transport/entry.go
i-hate-nicknames: Yes, it is possible to create two entries with the same ID but different ordering of edges, but since IDs are used to discriminate entries it shouldn't matter. 

1. I don't think so. ID is the same for entries (a, b) and (b, a). Transport discovery uses ID to store the entry, and if there is such ID, it does nothing. Other parts of TD just use the ID.

2. Yes, upon startup visor gets all transports from transport discovery and saves them internally. When there is a request to establish new transport, it first checks if transport already exists, using transport ID to determine it.

I can imagine a scenario when two visors try to establish a transport simultaneously to each other, and each holding its own version of entry. There will be no double connection: when visor accepts a new connection that corresponds to an existing transport it will automatically close one of them (this time using smallest PK logic). They will probably both try to redial when the connection is lost since both think they are the initiator, and probably will close one of the connections when there are two for one transport after redial. This is speculating from the code, I haven't performed any tests.

This sounds unlikely though since two visors should create a transport at the same time.

pkg/transport/entry.go
i-hate-nicknames: Don't think we do, good catch

ersonp (approved):
Everything looks good.

PR #816, opened 2021-06-21
Fix improve transport setup logic issue

### LINKED ISSUE #771: Improve transport setup logic
**Describe the bug**

When we call `saveTransport`, the call does not error even if the transport is not setup (i.e timeout while dialing). A transport is saved that is `down`. This is not optimal, as the transport is not actually established and requires additional checks in functions calling `saveTransport`. We also display down transports where we should return an error to users. We should also not register tps in discovery that were not established successfully in the first place.

mrpalide:

Fixes #771 	

Changes:
- Check transport pk before serving instead after that.

How to test this PR:

### DIFF
--- a/pkg/transport/manager.go
+++ b/pkg/transport/manager.go
@@ -306,37 +306,11 @@ func (tm *Manager) SaveTransport(ctx context.Context, remote cipher.PubKey, tpTy
 	for {
 		mTp, err := tm.saveTransport(remote, true, tpType, label)
 		if err != nil {
-			return nil, fmt.Errorf("save transport: %w", err)
-		}
-
-		tm.Logger.Debugf("Dialing transport to %v via %v", mTp.Remote(), mTp.netName)
-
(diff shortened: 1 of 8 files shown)


pkg/transport/manager.go
```diff
@@ -305,37 +305,11 @@ func (tm *Manager) SaveTransport(ctx context.Context, remote cipher.PubKey, tpTy
 	for {
 		mTp, err := tm.saveTransport(remote, tpType, label)
```
jdknives: Should we pass context here that we accept in `SaveTransport` instead of creating a new context in `saveTransport`?

jdknives (commented):
Looks good overall.

pkg/transport/manager.go
mrpalide: Yes, your right, I had forgotten context in `SaveTransport`.

PR #818, opened 2021-06-22
Fix improve transport setup logic issue

### LINKED ISSUE #771: Improve transport setup logic

mrpalide:

Fixes #771 	

Changes:
- Check transport pk before serving instead after that.

How to test this PR:

### DIFF
--- a/pkg/transport/manager.go
+++ b/pkg/transport/manager.go
@@ -192,7 +192,7 @@ func (tm *Manager) initTransports(ctx context.Context) {
 			tpID   = entry.Entry.ID
 		)
 		isInitiator := tm.n.LocalPK() == entry.Entry.Edges[0]
-		if _, err := tm.saveTransport(remote, isInitiator, tpType, entry.Entry.Label); err != nil {
+		if _, err := tm.saveTransport(ctx, remote, isInitiator, tpType, entry.Entry.Label); err != nil {
 			tm.Logger.Warnf("INIT: failed to init tp: type(%s) remote(%s) tpID(%s)", tpType, remote, tpID)
 		} else {
 			tm.Logger.Debugf("Successfully initialized TP %v", *entry.Entry)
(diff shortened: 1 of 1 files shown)


ersonp (approved):
Working as expected. Can see error messages for all 4 transport types if they are not available.

pkg/transport/manager.go
```diff
@@ -192,7 +192,7 @@ func (tm *Manager) initTransports(ctx context.Context) {
 			tpID   = entry.Entry.ID
 		)
 		isInitiator := tm.n.LocalPK() == entry.Entry.Edges[0]
-		if _, err := tm.saveTransport(remote, isInitiator, tpType, entry.Entry.Label); err != nil {
+		if _, err := tm.saveTransport(ctx, remote, isInitiator, tpType, entry.Entry.Label); err != nil {
```
i-hate-nicknames, line 195: if I understand correctly, you moved dialing part from `SaveTransport` to `saveTransport`. The naming is probably not the best, but `SaveTransport` is called via rpc and api, and is essentially "I want to add new transport to there". 
`saveTransport` is actually "save this transport and run it". Here it is saved when a transport entry is obtained from transport discovery. Original implementation includes measures to prevent both ends of transport of dialing each other (initially by using PK value and recently by using initiator logic). Currently, it seems that both ends can get into dialing state easily. I'm not sure if this won't introduce weird connectivity issues when a visor comes online with a number of transports for him in TD, while other ends of them are also online.
Can you please elaborate on this?

mrpalide: The situation that you explain is same as that I have a visor with a served transport (for example a VPN as dmsg), and get offline, and get back online after two hours for example. For now in [initTransport](https://github.com/skycoin/skywire/blob/e2193fb9fefb89f21bf985351bee2801c9979fdb/pkg/visor/init.go#L276), that called when visor get back online, the `Serve` run again, but without dialing discovery, that means transport(s) added to visor both in online or offline status.

PR #825, opened 2021-06-28
Add stun NAT type check for sudph

### LINKED ISSUE #776: Add holepunching check
**Feature description**

We currently do not perform any checks whether any given visor is able to perform holepunching. This leads to a number of issues

- user may try to setup holepunch tp without success 
- other users may try to setup a holepunch tp to a visor without success (requires timeout to hit)

It would be advantageous for users and for us to check upon startup if a visor is available for holepunching. The main factor in determining this is the type of NAT a given user is behind. There is a utility called `PyStun` which allows gathering that information. We should check how they perform that check and see if we can implement something similar. If a visor is not available for holepunching because it is behind a symmetric NAT, we should not register it for holepunching and not allow setting up this transport in frontend either. 

Potentially we could expose `sudph` available visors in AR and perform a check in frontend if remote visor is available for holepunching before user opens up the transport.

ersonp:
yes

Fixes #776 

 Changes:	
- Added a `STUN` check for the `sudph` transport.
- Separated transports into their own modules so that they can be initialized individually and the visor will start only when at least one transport has been initialized. 
- `dmsg` and `sudph` initialize separate from other modules so that the visor doesn't have to wait for them. 
- Removed the unused `public` field from AddTransports.

How to test this PR:
1. Run `./skywire-visor skywire-config.json`
2. If your visor is under a symmetric NAT you should not be able to see a `sudph` transport. (check the logs to see your NAT type)


jdknives:
Looks good @ersonp . Will test and merge once we have a Stun server deployed ourselves.

ersonp:
Few things are remaining such as:
1. Add stun servers to config 
2. Put public IP and Nat type in visor summary
Will push them soon.

pkg/snet/network.go
```diff
@@ -165,8 +168,16 @@ func New(conf Config, eb *appevent.Broadcaster, masterLogger *logging.MasterLogg
 			SK:              conf.SecKey,
 			AddressResolver: conf.ARClient,
 		}
-
-		clients.Direct[tptypes.SUDPH] = directtp.NewClient(sudphConf, masterLogger)
+		// SUDPH wont be initilised only if the visor is under any symmetric NAT
```
jdknives: Would shorten to just `// sudph will only be initialized if NAT is determined not to be symmetric`

pkg/visor/api.go
```diff
@@ -90,11 +91,16 @@ type Overview struct {
 	Transports      []*TransportSummary  `json:"transports"`
 	RoutesCount     int                  `json:"routes_count"`
 	LocalIP         string               `json:"local_ip"`
+	NATType         string               `json:"nat_type"`
+	PublicIP        string               `json:"public_ip"`
```
jdknives, line 98: Not sure if we need to provide a NAT type here. Maybe it would be sufficient to just return if it is available for `sudph` and `stcpr`.

pkg/visor/init.go
```diff
@@ -186,10 +187,39 @@ func initSNet(ctx context.Context, v *Visor, log *logging.Logger) error {
 		STCP: v.conf.STCP,
 	}
 
+	var nat stun.NATType
+	var host *stun.Host
+	var err error
+stunLoop:
```
jdknives: It is somewhat annoying that this is in `snet` package as the Stun check takes a substantial amount of time and blocks most other modules from initializing with it. @i-hate-nicknames does your refactor make it possible to move this to some other module that is not the main dependency for most modules?

jdknives: I guess with the current list of Stun servers, the best thing to do is iterate over them and used the named loop like you do. But I think it should be fine for now to just have one server down the road for this so we can also get rid of this.

i-hate-nicknames: @jdknives Yes I think. Looks to me as the following:
 
1. Make a stun module similarly to other modules
2. Make some stun "fetcher" object that will that can initialize itself with this heavy logic and then store results.
3. Create init function for the module in (1). Create object from (2) there and call the heavy logic. Then assign this object into visor field.

Similarly to what other modules do essentially.

i-hate-nicknames: @ersonp can we avoid using goto though?

pkg/snet/network.go
```diff
@@ -165,8 +168,16 @@ func New(conf Config, eb *appevent.Broadcaster, masterLogger *logging.MasterLogg
...
+		// along with the other NAT types all other errors will also initilise the SUDPH
+		// as without a proper response from a STUN server we cannot determine if the
+		// visor is un der any symmetric NAT
+		switch conf.NATType {
+		case stun.NATSymmetric, stun.NATSymmetricUDPFirewall:
+			break
```
i-hate-nicknames: I would add an Info level log message here that sudph transport is not available because of NAT

pkg/visor/init.go
```diff
@@ -186,10 +187,39 @@ func initSNet(ctx context.Context, v *Visor, log *logging.Logger) error {
...
+		if err != nil {
+			log.Warn(err)
+		}
+
+		switch nat {
+		case stun.NATError, stun.NATUnknown, stun.NATBlocked:
+			log.Warn(nat.String())
+			// incase we fail to connect to all the STUN servers give a warning
+			if stunServer == v.conf.StunServers[len(v.conf.StunServers)-1] {
+				log.Warn("All STUN servers offline")
```
i-hate-nicknames: What will be the values of `nat` and `host` variables in this case? And what is expected behavior of sudph if we cannot determine nat type?

ersonp: `Nat` and `host` will have the value of whatever the result was with the STUN check. In that case, we don't need to show the host value anyway since we don't have any info on the client.

pkg/snet/network.go
ersonp: Good catch. Will do.

pkg/visor/init.go
ersonp: I'll try to find a better way without using goto. Let me know if you have any suggestions @i-hate-nicknames

pkg/snet/network.go
ersonp: Will make the change.

pkg/visor/api.go
ersonp: Okay

ersonp: About `public_ip`; i am currently sending the public IP and port together. Is the port required or should I send just the IP? @jdknives

pkg/visor/init.go
ersonp: We initilize the sudph if we are unable to determine the nat type.

pkg/visor/api.go
ersonp: Also i haven't done any check regarding `sctpr` because it depends on only one side being on a public ip. So i think the NAT info should be send to the address resolver so it can be used to determin if the connection is possible before attempting it.

ersonp:
@jdknives @i-hate-nicknames I have turned stun into a module but corrently all it's related code is in `pkg/visor/init`. I think it should go under `transport` or `transport/network` the second one being from the snet refactor. Let me know if you think otherwise.

pkg/snet/network.go
```diff
@@ -252,6 +281,22 @@ func (n *Network) IsNetworkReady(netType string) bool {
 	return ok
 }
 
+// WaitTillNetworkReady waits till any one of the network is ready.
+func (n *Network) WaitTillNetworkReady(log *logging.Logger) {
+	n.netsMu.Lock()
```
i-hate-nicknames: I would use a done channel and sync.Once instead of active polling

pkg/visor/init.go
```diff
@@ -84,6 +95,8 @@ var (
 	pvs vinit.Module
 	// Public visor: advertise current visor as public
 	pv vinit.Module
+	// Transport check module
```
i-hate-nicknames: interesting idea to implement non-required dependencies as a regular module that is not included in visor but is initialized too, didn't think of it.
I would add a comment here that this is not a functional module but a grouping of all transport types initializations

pkg/visor/init.go
```diff
@@ -62,8 +63,18 @@ var (
...
+	// SUDPH client
+	sudph vinit.Module
+	// SCTPR client
+	sctpr vinit.Module
+	// SCTP client
+	sctp vinit.Module
 	// Snet (different network types)
 	sn vinit.Module
+	// SnetCheck (check if atleast one network is available)
+	snCheck vinit.Module
```
i-hate-nicknames: I don't think we need this as a module. The whole point of modules is to run two initialization functions in parallel to gain performance. Basically this module initialization is "is snet already usable?". And anyone that wants snet functionality would sure want it to be usable, so will have to depend on this module instead of snet, and snet and snCheck will have to initialize sequentially all the time.

So I would put the check if snet is ready to be used in snet initialization function

pkg/visor/init.go
```diff
@@ -766,3 +819,40 @@ func getErrors(ctx context.Context) chan error {
 	}
 	return errs
 }
+
+func stunDetails(stunServers []string, log *logging.Logger) *snet.StunClient {
```
i-hate-nicknames: Can we move it to a separate package? Perhaps into snet itself, since it's used by it.

pkg/snet/network.go
```diff
@@ -252,6 +281,22 @@ func (n *Network) IsNetworkReady(netType string) bool {
...
+	nets := n.nets
+	n.netsMu.Unlock()
+	if len(nets) == 0 {
+		print(len(nets))
+		time.Sleep(200 * time.Millisecond)
+		n.WaitTillNetworkReady(log)
+	} else {
+		for net := range nets {
+			log.Infof("Transport %v initialized", net)
```
i-hate-nicknames: This will allow dmsg to allow snet to get initialized. Do we consider dmsg to be sufficient in terms of running a visor @jdknives ?

i-hate-nicknames (changes requested):
Good job overall! Also I would just remove the code that you commented out. Most likely it will be gone when snet refactoring is merged, but still a good habit

pkg/visor/init.go
ersonp: Okay, will do.

pkg/visor/init.go
ersonp: Yes I'll move the stun code to snet.

pkg/snet/network.go
ersonp: Good idea. didn't think of it.

pkg/visor/init.go
ersonp: All the transport modules depend on the snet module so without it no transports will be initialized.

ersonp: Okay.

pkg/visor/init.go
```diff
@@ -257,6 +259,7 @@ func initSctpClient(ctx context.Context, v *Visor, log *logging.Logger) error {
 }
 
 func initDmsgCtrl(ctx context.Context, v *Visor, log *logging.Logger) error {
+	<-v.isNetConf
```
i-hate-nicknames: You don't need this synchronization. Since dmsgCtrl module depends on snet module, `initDmsgCtrl` is guaranteed to run after `initSnet`. 

After snet refactoring (the changes just made it to develop), snet module and snet.Network are removed, and dmsg is taken out into its own module and dmsgCtrl depends on dmsg directly.

i-hate-nicknames (commented):
Just a small comment, otherwise looks good.

Also, now that snet refactoring got merged there is no snet package and snet.Network entity, but the changes should fit perfectly into transport manager.

pkg/visor/init.go
ersonp: Yes solving conflicts with develop rn, will make the changes.

pkg/snet/network.go
jdknives: @i-hate-nicknames Either the code here changed or I am missing some context because I cant relate your comment to the code. But I consider dmsg running to be sufficient to let the visor run and start all other modules.

ersonp:
need to merge #842

pkg/router/router.go
```diff
@@ -317,7 +317,7 @@ func (r *router) Serve(ctx context.Context) error {
 
 	go r.serveSetup()
 
-	r.tm.Serve(ctx)
+	// r.tm.Serve(ctx)
```
mrpalide: Why not deleted complete?

mrpalide (approved):
Everything (just a little comment) is OK. Ready to merge.

pkg/router/router.go
ersonp: Left it by mistake. Will take it out.

pkg/visor/api.go
```diff
@@ -107,6 +112,15 @@ func (v *Visor) Overview() (*Overview, error) {
 			newTransportSummary(v.tpM, tp, true, v.router.SetupIsTrusted(tp.Remote())))
 		return true
 	})
+	if v.stunClient != nil {
+		switch v.stunClient.NATType {
+		case stun.NATNone, stun.NATFull, stun.NATRestricted, stun.NATPortRestricted:
```
jdknives, line 120: This seems to be the wrong way. If this is the case, we assign `isSymmetricNAT` true.

pkg/transport/manager.go
```diff
@@ -67,55 +71,58 @@ func NewManager(log *logging.Logger, arClient addrresolver.APIClient, ebc *appev
...
-		go tm.acceptTransports(ctx, lis)
+	tm.mx.Lock()
+	client := tm.netClients[netType]
+	tm.mx.Unlock()
+	tm.Logger.Infof("Serving %s network", client.Type())
+	err := client.Start()
+	if err != nil {
+		tm.Logger.WithError(err).Errorf("Failed to listen on %s network", client.Type())
 	}
+	lis, err := client.Listen(skyenv.DmsgTransportPort)
```
jdknives: @ersonp not part of your PR specifically, but can we check why we are initializing all transport client listeners with dmsg transport port values? As the name suggests, these should only be used for dmsg.

pkg/visor/init.go
```diff
@@ -105,20 +116,27 @@ func registerModules(logger *logging.MasterLogger) {
 	ebc = maker("event_broadcaster", initEventBroadcaster)
 	ar = maker("address_resolver", initAddressResolver)
 	disc = maker("discovery", initDiscovery)
+	tr = maker("transport", initTransport, &ar, &ebc)
+
+	sc = maker("stun_client", initStunClient)
+	sudph = maker("sudph", initSudphClient, &sc, &tr)
+	sctpr = maker("sctpr", initSctprClient, &tr)
```
jdknives: `stcpr` as in Skywire TCP with automatic address Resolution. Same for `stcp`.

pkg/visor/visor.go
```diff
@@ -57,11 +58,13 @@ type Visor struct {
...
-	arClient addrresolver.APIClient
-	router   router.Router
-	rfClient rfclient.Client
+	stunClient *network.StunDetails
+	tpM        *transport.Manager
+	arClient   addrresolver.APIClient
+	router     router.Router
+	rfClient   rfclient.Client
 
+	isNetConf   chan bool             // net config check
```
jdknives: Naming here is not clear to me. What are we doing with the value?

jdknives (commented):
Looks mostly good.

jdknives:
Please pull in `develop` and fix conflicts.

pkg/visor/visor.go
ersonp: That was leftover code. Will take it out.

pkg/visor/init.go
ersonp: My bad. Will make the changes.

pkg/visor/api.go
ersonp: Thanks for pointing that out. I have updated the code to include error/unknown cases. Let me know if it isn't required.

pkg/transport/manager.go
ersonp: Seems like it has been named incorrectly compared to it's comment and usage.

ersonp:
@jdknives Currently we are not able to use `sudph` in integration env in skywire-services because of the STUN check. So I think we should add either a `test` flag in or a `test` field in config to turn off the STUN check.

PR #831, opened 2021-07-01
Dynamic servers address selected

mrpalide:

Fixes: Issue [426](https://github.com/SkycoinPro/skywire-services/issues/426) in SkycoinPro/skywire-services

 Changes:
- [x] Create new servers.yml file that should fetch from skycoin.com or load from local file
- [x] Add two variable to skywire-config.json
    - `ServersListAddress` : address for fetching servers list data
    - `IsTest` : checking config is for test or production
- [x] Add new flag for skywire-visor `--local-server` that force to connect local servers
- [x] Check timezone automate for choosing best server 

How to test this PR:
Every command like before, just for choosing local servers manually, add `--local-server true` to skywire-visor running command like this:

> `sudo ./skywire-visor -c skywire-config.json --local-server true`

### DIFF
--- a/pkg/skyenv/values.go
+++ b/pkg/skyenv/values.go
@@ -8,8 +8,9 @@ import (
 
 // Constants for skywire root directories.
 const (
-	DefaultSkywirePath        = "."
-	PackageSkywirePath        = "/opt/skywire"
+	DefaultSkywirePath = "."
+	PackageSkywirePath = "/opt/skywire"
+	// TODO make a decision about actual name of the domain
 	DefaultServersListAddress = "http://servers-list.skycoin.com"
 )
 
--- a/pkg/skyenv/values.go
+++ b/pkg/skyenv/values.go
@@ -8,8 +8,9 @@ import (
 
 // Constants for skywire root directories.
 const (
-	DefaultSkywirePath = "."
-	PackageSkywirePath = "/opt/skywire"
+	DefaultSkywirePath        = "."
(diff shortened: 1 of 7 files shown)


servers.yml
```diff
@@ -0,0 +1,39 @@
+test:
```
jdknives, line 1: Does yml allow comments? If it does, we should add some basic documentation for users how to deal with this file and explain the non-obvious fields such as capacity.

pkg/skyenv/values.go
```diff
@@ -8,8 +8,9 @@ import (
 
 // Constants for skywire root directories.
 const (
-	DefaultSkywirePath = "."
-	PackageSkywirePath = "/opt/skywire"
+	DefaultSkywirePath        = "."
+	PackageSkywirePath        = "/opt/skywire"
+	DefaultServersListAddress = "http://servers-list.skycoin.com"
```
jdknives, line 14: I will speak to Murphy about the actual name of the domain. Add a TODO here for now please.

mrpalide: Added

servers.yml
mrpalide: It's possible for comments, but I don't know add what and how docmuentation.

alexadhy:
Can we use toml / json instead? saw that the `servers.yml` config doesn't contain that many children / indentations. Just my 2 cents, correct me if I'm wrong.

mrpalide:
Because of some problems, this PR continues on #918.

PR #863, opened 2021-08-21
add dmsgpty to skywire-cli

0pcom:
add dmsgpty-cli and dmsgpty-ui comands to skywire-cli.

This PR depends on:
https://github.com/skycoin/dmsg/pull/101

I've been able to test this by manually changing the dependencies to reflect the pull request in dmsg.
The dmsgpty-cli works, but the dmsgpty-ui does not register as a command, yet there are no errors or any indication why it does not appear.

The ability to launch the dmsgpty-ui from skywire-cli would be nice but is not essential for my purposes

dmsgpty-cli has been missing from skywire-cli; currently dmsgpty-ui is included with the visor.
this is a necessary step for further integration of update functionality into skywire-cli.

 Changes:	
-	add dmsgpty commands

How to test this PR:
```
skywire-cli visor dmsgpty-cli
```

### DIFF
--- a/cmd/skywire-cli/commands/visor/dmsgpty.go
+++ b/cmd/skywire-cli/commands/visor/dmsgpty.go
@@ -1,0 +1,13 @@
+package visor
+
+import (
+dmsgcli	"github.com/skycoin/dmsg/cmd/dmsgpty-cli/commands"
+dmsgui	"github.com/skycoin/dmsg/cmd/dmsgpty-ui/commands"
+)
+
+func init() {
+	RootCmd.AddCommand(
+		dmsgui.RootCmd,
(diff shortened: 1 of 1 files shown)


jdknives:
@0pcom just gave this a run and it will not work as is:

- dmsgpty cli expects a config file to connect to the dmsgpty host 
- the dmsgpty config is not the same as the config file we use in skywire

We do not currently generate a `dmsgpty` config in Skywire when running a visor (or when configuring a visor). We just take various values we need from the main config in Skywire and pass them to dmsgpty. 

I would not like to add support in dmsg for the Skywire config as dmsg should be independent of Skywire completely. So we should add support for it in Skywire repo.

PR #872, opened 2021-09-07
Update service discovery references

ersonp:

Fixes #	

 Changes:	
- Change `proxyDiscovery` to `serviceDiscovery`

How to test this PR:

### DIFF
--- a/pkg/visor/visorconfig/v1.go
+++ b/pkg/visor/visorconfig/v1.go
@@ -100,15 +100,15 @@ type V1UptimeTracker struct {
 	Addr string `json:"addr"`
 }
 
-// V1Disc configures Skywire App Discovery Clients.
-type V1Disc struct {
+// V1ServiceDisc configures Skywire App Discovery Clients.
+type V1ServiceDisc struct {
 	UpdateInterval Duration `json:"update_interval,omitempty"`
 	ServiceDisc    string   `json:"service_discovery"`
 }
 
 // V1Launcher configures the app launcher.
 type V1Launcher struct {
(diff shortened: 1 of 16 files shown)


pkg/visor/visorconfig/v1.go
```diff
@@ -100,15 +100,15 @@ type V1UptimeTracker struct {
 	Addr string `json:"addr"`
 }
 
-// V1AppDisc configures Skywire App Discovery Clients.
-type V1AppDisc struct {
+// V1Disc configures Skywire App Discovery Clients.
+type V1Disc struct {
```
jdknives: `ServiceDisc` maybe better? We have three discoveries at this point overall.

jdknives (commented):
Should we rename the package `appdisc` as well? Or is just responsible for updating app statuses in service discovery? Also do we need to define a config migration, right ? Also, please add changes to changelog. Will need to start doing that more diligently.

pkg/visor/visorconfig/v1.go
ersonp: Okay

ersonp:
- We can name `appdisc` to `updatedisc` instead since that is what it's doing since `appdisc` can be confusing. 
- The `updateUrls` in `parse` takes care of the migration. Only thing is that the URL is changed to `http://sd.skycoin.com` with it.
- Will add all the changes of config `v1.1.0` to Changelog. Also, I will need some inputs for all the changes introduced in the current version of `skywire@develop`.

jdknives (approved):

PR #883, opened 2021-09-10
Dmsg delete entry on shutdown

ersonp:

Fixes #	

 Changes:	
- Added `dmsgC.Close()` in `initDmsg` in modules
- Removed `go dmsgC.Serve(ctx)` from `NewTransportListener` as it was being called twice.

How to test this PR:
1. Checkout dmsg repo on PR [#110](https://github.com/skycoin/dmsg/pull/110).
2. In skywire uncomment `replace github.com/skycoin/dmsg => ../dmsg` in go.mod .
3. Run `make dep`
4. In `skywire-services` make `log:"debug"` in `integration/generic/visorA.json`
5. Start integration env in `skywire-services`
6. Run in different terminal
```bash
$ redis-cli
> get <PK-OF-VISORA>
```
7. Wait for 5 mins and check again. It should still be there.
8. Close visor A with `Ctrl+C`
9. Check for the pk again in redis. The entry should be gone.

This can be tested without integration env on `http://dmsg-discovery.myownsky.xyz`.


pkg/visor/init.go
```diff
@@ -212,9 +212,22 @@ func initDmsg(ctx context.Context, v *Visor, log *logging.Logger) error {
...
 	dmsgC := dmsgc.New(v.conf.PK, v.conf.SK, v.ebc, v.conf.Dmsg)
 
-	time.Sleep(200 * time.Millisecond)
-	go dmsgC.Serve(context.Background())
-	time.Sleep(200 * time.Millisecond)
+	wg := new(sync.WaitGroup)
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		time.Sleep(200 * time.Millisecond)
```
alexadhy: why the need for Sleep? is it for coordination or something else?

ersonp: It was already there. I Will check if it's still required or not and remove it if isn't.

ersonp:
@jdknives I was looking at the code and it seems like we need a nil check for [TransportSetup](https://github.com/skycoin/skywire/blob/2a76c6a19326ff8f78d791d740252895ac15be6a/pkg/visor/init.go#L346) module since `transport_setup_nodes` in the config is always null. And if I am not wrong then it only works when `transport_setup_nodes` has a list of pk's.

jdknives:
> @jdknives I was looking at the code and it seems like we need a nil check for [TransportSetup](https://github.com/skycoin/skywire/blob/2a76c6a19326ff8f78d791d740252895ac15be6a/pkg/visor/init.go#L346) module since `transport_setup_nodes` in the config is always null. And if I am not wrong then it only works when `transport_setup_nodes` has a list of pk's.

We need to define a PubKey for that service before we push out the release so it is not null by default. If it is though by user intervention, we should not initialize the module.

pkg/visor/init.go
```diff
@@ -212,9 +212,20 @@ func initDmsg(ctx context.Context, v *Visor, log *logging.Logger) error {
...
-	time.Sleep(200 * time.Millisecond)
+	wg := new(sync.WaitGroup)
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		dmsgC.Serve(context.Background())
+	}()
+
+	v.pushCloseStack("dmsg", func() error {
+		if err := dmsgC.Close(); err != nil {
```
jdknives, line 223: We are closing twice. Once here and once when we shutown the clients on https://github.com/skycoin/skywire/blob/2a76c6a19326ff8f78d791d740252895ac15be6a/pkg/transport/manager.go?plain=1#L547 if I am not mistaken.

ersonp: We need to close dmsg explicitly https://github.com/skycoin/skywire/blob/2a76c6a19326ff8f78d791d740252895ac15be6a/pkg/transport/network/dmsg.go#L65

jdknives (approved):

pkg/router/router.go
```diff
@@ -351,9 +351,9 @@ func (r *router) serveSetup() {
 			log := r.logger.WithError(err)
 			if err == dmsg.ErrEntityClosed {
```
alexadhy: for the sake of consistency I guess, this one matches the error by value, the one on the file below (transport/manager.go) uses proper error type checks (`errors.Is`), probably use that also?

pkg/transport/setup/visor.go
```diff
@@ -62,7 +60,12 @@ func (ts *TransportListener) Serve(ctx context.Context) {
 	for {
 		conn, err := lis.AcceptStream()
 		if err != nil {
-			ts.log.WithError(err).Error("failed to accept")
+			log := ts.log.WithError(err)
+			if err == dmsg.ErrEntityClosed {
```
alexadhy: same with this one?

pkg/visor/hypervisor.go
```diff
@@ -1549,7 +1549,13 @@ func (hv *Hypervisor) serveDmsg(ctx context.Context, log *logging.Logger) {
 	go func() {
 		<-hv.dmsgC.Ready()
 		if err := hv.ServeRPC(ctx, hv.c.DmsgPort); err != nil {
-			log.WithError(err).Fatal("Failed to serve RPC client over dmsg.")
+			log := log.WithError(err)
+			if err == dmsg.ErrEntityClosed {
```
alexadhy: and this one

alexadhy (changes requested):
just proper error checks thing for consistency

alexadhy (approved):

PR #447, opened 2020-07-23
[WIP] Skywire Android port

Darkren:
Did you run `make format && make check`?
Yes

Fixes #138 	

 Changes:	
- Added a separate VPN client for mobile devices
- Added `skywiremob` library to expose Skywire network API to mobile devices
- Added example Android Skywire VPN application
- Visor's startup is now fully interruptable

How to test this PR:
1. Install gomobile from https://github.com/golang/mobile
2. Install Android SDK and NDK. Put its locations to corresponding env vars: ANDROID_HOME for SDK and ANDROID_NDK_HOME for NDK
3. Compile the `skywiremob` library. To do this one should put the skywire repository according to the `GOPATH`, i.e.: `$GOPATH/src/github.com/skycoin/skywire`. Then you need to `cd` to the directory right outside of the `GOPATH`. So, having default `GOPATH` at `$HOME/go` you need to `cd $HOME`. From there you may build the library like this:
```
$ gomobile bind -o ./go/src/github.com/skycoin/skywire/cmd/skywirevisormobile/android/app/skywire.aar -target=android ./go/src/github.com/skycoin/skywire/pkg/skywiremob/
```
Probably it may be successfully built the other way, but this is the only way I got it to work. The above command will put the `skywire.aar` library right to where the Android project is looking for it.
4. Build the apk. May be done from the terminal or Android studio
5. Install and run apk
6. Put correct remote PK of visor running VPN server and passcode for that server and press `Start`. Once VPN is ready you will see the toast message
7. Check your IP address

PR #884, opened 2021-09-13
Remove heartbeat | Add Deregister

mrpalide:

Fixes [Skywire Service #447](https://github.com/SkycoinPro/skywire-services/issues/447)

 Changes:	
- remove updater loop
- change `Delete` entry to `Deregister`
- add `Deregister` on client side (skywire) stop app process and visor shutdown process.

How to test this PR:
- Run visor with vpn-server true, then stop it by hypervisor UI or shutdown visor from CLI. Then check [`sd.skycoin.com/api/services?type=vpn`](sd.skycoin.com/api/services?type=vpn) for related PK.


pkg/app/appserver/proc.go
```diff
@@ -259,6 +259,9 @@ func (p *Proc) Stop() error {
 		}
 	}
 
+	// deregister discovery service
+	p.disc.Stop()
```
jdknives, line 242: Can `disc` be zero for apps without an updater such as `skychat`?

pkg/app/updatedisc/discovery_updater.go
```diff
@@ -52,7 +52,9 @@ func (u *serviceUpdater) Start() {
 
 	u.wg.Add(1)
```
jdknives: Do we still need the waitgroup?

pkg/app/updatedisc/discovery_updater.go
```diff
@@ -52,7 +52,9 @@ func (u *serviceUpdater) Start() {
...
 	go func() {
-		u.client.UpdateLoop(ctx, u.interval)
+		if err := u.client.Update(ctx); err != nil {
+			return
```
jdknives: May the return here lead to the waitgroup never finishing?

jdknives (commented):
It seems this PR is unfinished. There are a few questions regarding the changes you made but more importantly there seem to be a few things that have not been tackled at all yet. We have an updating interval in the `serviceupdater` struct defined despite us getting rid of the loop using that value. 

I also wonder whether we still need the `waitgroup` and the `cancelfunc` given that we get  rid of the update loop which we originally used it for.

pkg/app/appserver/proc.go
mrpalide: For app like skychat that doesn't need service discovery, we associate [empty updater](https://github.com/skycoin/skywire/blob/2a76c6a19326ff8f78d791d740252895ac15be6a/pkg/app/updatedisc/factory.go#L88) and `Stop()` for those is [empty function](https://github.com/skycoin/skywire/blob/2a76c6a19326ff8f78d791d740252895ac15be6a/pkg/app/updatedisc/discovery_updater.go#L29) too. So derigstration never run for them.

pkg/app/updatedisc/discovery_updater.go
mrpalide: I think goroutine for `updaterloop` was for other reason than loop, but now I remove them.

pkg/app/updatedisc/discovery_updater.go
mrpalide: Your right, but waitgroup remove complete.

pkg/app/managedisc/discovery_manager.go
```diff
@@ -0,0 +1,67 @@
+package managedisc
```
jdknives: I agree that a package name change is necessary. Not sure if `managedisc` is really a good name. We just recently changed from `appdisc` to `updatedisc` but I think under these circumstances `appdisc` may be better.

pkg/app/managedisc/discovery_manager.go
```diff
@@ -0,0 +1,67 @@
...
+	u.mu.Lock()
+	defer u.mu.Unlock()
+
+	ctx := context.Background()
+	if err := u.client.DeregisterEntry(ctx); err != nil {
+		return
+	}
+}
+
+func (u *serviceManager) ChangeValue(name string, v []byte) error {
```
jdknives: This function currently is useless as we do not periodically update the entry anymore. So either we get rid of the servicestats in the entry or we update the entry once we the connection stats change. Please check if you can update on change of the connection stats.

pkg/app/managedisc/factory.go
```diff
@@ -14,30 +12,26 @@ import (
 
 // Factory creates appdisc.Updater instances based on the app name.
```
jdknives: Comment needs to be updated as well.

pkg/app/managedisc/factory.go
```diff
@@ -14,30 +12,26 @@ import (
...
-	}
 	if f.ServiceDisc == "" {
 		f.ServiceDisc = skyenv.DefaultServiceDiscAddr
 	}
 }
 
-// VisorUpdater obtains a visor updater.
-func (f *Factory) VisorUpdater(port uint16) Updater {
-	// Always return empty updater if keys are not set.
+// VisorManager obtains a visor manager.
```
jdknives: Visor manager is a bad name as it implies the visor is managed through this structure while it is just a client for updating entries for services it runs.

pkg/app/managedisc/factory.go
```diff
@@ -48,24 +42,23 @@ func (f *Factory) VisorUpdater(port uint16) Updater {
...
-		interval: f.UpdateInterval,
+	return &serviceManager{
+		client: servicedisc.NewClient(f.Log, conf),
 	}
 }
 
-// AppUpdater obtains an app updater based on the app name and configuration.
-func (f *Factory) AppUpdater(conf appcommon.ProcConfig) (Updater, bool) {
-	// Always return empty updater if keys are not set.
+// AppManager obtains an app manager based on the app name and configuration.
```
jdknives: Similarly to above, this does not actually manage apps, but updates entries of apps in a discovery.

pkg/servicedisc/client.go
```diff
@@ -230,8 +230,8 @@ func (c *HTTPClient) postEntry(ctx context.Context) (Service, error) {
 	return entry, err
 }
 
-// DeleteEntry calls 'DELETE /api/services/{entry_addr}'.
-func (c *HTTPClient) DeleteEntry(ctx context.Context) (err error) {
+// DeregisterEntry calls 'DELETE /api/services/{entry_addr}'.
```
jdknives: Maybe `DeleteEntry` is clearer. Sorry for not using the correct term before.

jdknives (approved):

pkg/servicedisc/client.go
mrpalide: Done.

pkg/app/managedisc/factory.go
mrpalide: Done.

pkg/app/managedisc/factory.go
mrpalide: Done.

pkg/app/managedisc/discovery_manager.go
mrpalide: Done!

pkg/app/managedisc/factory.go
mrpalide: After changing from `managedisc` to `appdisc` and change `VisorManager` and `AppManager` to `*Updater` no need to change.

pkg/app/appdisc/factory.go
```diff
@@ -81,8 +74,7 @@ func (f *Factory) AppUpdater(conf appcommon.ProcConfig) (Updater, bool) {
 	switch conf.AppName {
 	case skyenv.VPNServerName:
```
ersonp, line 75: Should we add the proxy-server back since we are getting rid of the heartbeat @jdknives 
Also it would be better if we rename it to `skysocks` instead of `proxy`.

pkg/servicedisc/client.go
```diff
@@ -269,40 +269,10 @@ func (c *HTTPClient) DeleteEntry(ctx context.Context) (err error) {
...
-	}
-}
-
-// Update calls 'POST /api/services' to update service discovery entry
-// it performs exponential backoff in case of errors during update, unless
+// RegisterEntry calls 'POST /api/services' to register service discovery entry
+// it performs exponential backoff in case of errors during register, unless
 // the error is unrecoverable from
-func (c *HTTPClient) Update(ctx context.Context) error {
+func (c *HTTPClient) RegisterEntry(ctx context.Context) error {
```
ersonp: We should name this `Register`, and rename `UpdateEntry` to `RegisterService`.
Edit: As per the comment made by rudi We should name `RegisterEntry` to `Register`, and rename `UpdateEntry` to `RegisterEntry`

ersonp: @jdknives I think it would be better if we are consistent with the naming convention. We are using `entry` here on the client side but `service` on `skycoin-service-discovery` side.

ersonp:
Besides some naming conventions for functions, everything looks good. Everything else is working as expected. I tested by registering `vpn-server` and `public visor`. `vpn-server` entry is deleted on stopping the app as well as with the shutdown logic. `public visor` entry is deleted with the shutdown logic.

pkg/app/appdisc/discovery_manager.go
```diff
@@ -1,15 +1,14 @@
...
 	"context"
 	"strconv"
 	"sync"
-	"time"
 
 	"github.com/skycoin/skywire/pkg/servicedisc"
 )
 
-// Updater updates the associated app discovery
+// Updater update the associated app discovery
```
ersonp: `updates` is the right word here so no need to change it to `update`

mrpalide: Done

pkg/app/managedisc/discovery_manager.go
mrpalide: Completely removed.

jdknives:
> Besides some naming conventions for functions, everything looks good. Everything else is working as expected. I tested by registering `vpn-server` and `public visor`. `vpn-server` entry is deleted on stopping the app as well as with the shutdown logic. `public visor` entry is deleted with the shutdown logic.

Fair point. We can call it `entry` on the service side as well. We do that in the `dmsg-discovery` and `transport-discovery` as well.

pkg/app/appdisc/discovery_manager.go
```diff
@@ -0,0 +1,50 @@
+package appdisc
```
jdknives, line 1: Should the file be named discovery_updater.go again?

pkg/app/appdisc/discovery_manager.go
```diff
@@ -0,0 +1,50 @@
...
+// emptyUpdater is for apps that do not require discovery updates.
+type emptyUpdater struct{}
+
+func (emptyUpdater) Start() {}
+func (emptyUpdater) Stop()  {}
+
+// serviceUpdater updates service-discovery entry of locally running App.
+type serviceUpdater struct {
+	client *servicedisc.HTTPClient
+	mu     sync.Mutex
```
jdknives: What do we need the mutex for here?

pkg/app/appdisc/factory.go
jdknives: @mrpalide @ersonp yes and yes.

pkg/visor/init.go
```diff
@@ -647,7 +646,7 @@ func initPublicVisor(_ context.Context, v *Visor, log *logging.Logger) error {
 	visorUpdater.Start()
 
 	v.log.Infof("Sent request to register visor as public")
-	v.pushCloseStack("visor updater", func() error {
+	v.pushCloseStack("visor manager", func() error {
```
jdknives: I would name it `public visor updater`.

pkg/app/appdisc/discovery_manager.go
mrpalide: Handled.

pkg/app/appdisc/discovery_manager.go
mrpalide: oh! I renamed everything, and forgot about file name.
Done now.

pkg/visor/init.go
mrpalide: Done

pkg/app/appdisc/factory.go
mrpalide: Done.

pkg/servicedisc/client.go
mrpalide: Done.

pkg/servicedisc/types.go
```diff
@@ -11,8 +11,8 @@ import (
 )
 
 const (
-	// ServiceTypeProxy stands for the proxy discovery.
-	ServiceTypeProxy = "proxy"
+	// ServiceTypeProxy stands for the skysocks discovery.
+	ServiceTypeProxy = "skysocks"
```
ersonp: We should rename `ServiceTypeProxy` to `ServiceTypeSkysocks`

mrpalide: Done

pkg/visor/visorconfig/v1.go
```diff
@@ -105,8 +105,7 @@ type V1UptimeTracker struct {
 
 // V1ServiceDisc configures Skywire App Discovery Clients.
 type V1ServiceDisc struct {
-	UpdateInterval Duration `json:"update_interval,omitempty"`
-	ServiceDisc    string   `json:"service_discovery"`
+	ServiceDisc string `json:"service_discovery"`
```
ersonp: Since we only have a URL now we can get rid of the struct and just add `ServiceDisc` field to `V1Launcher`. @jdknives @mrpalide  What are your thoughts?

ersonp:
Everything looks good besides the config, that I have mentioned above. I have tested `visor`, `vpn` and `skysocks` registration and deletion and it works as intended.

pkg/visor/visorconfig/v1.go
mrpalide: I'm agree with you and I do these change.

pkg/servicedisc/client.go
```diff
@@ -112,9 +106,9 @@ func (c *HTTPClient) Auth(ctx context.Context) (*httpauth.Client, error) {
 	return auth, nil
 }
 
-// Services calls 'GET /api/services'.
+// Services calls 'GET /api/entries'.
 func (c *HTTPClient) Services(ctx context.Context, quantity int) (out []Service, err error) {
-	url, err := c.addr("/api/services", c.entry.Type, quantity)
+	url, err := c.addr("/api/entries", c.entry.Type, quantity)
```
jdknives: Please do not change the endpoint for the regsitration/deregistration.

pkg/visor/visorconfig/config.go
```diff
@@ -33,12 +33,10 @@ func MakeBaseConfig(common *Common) *V1 {
 		RouteFinderTimeout: DefaultTimeout,
 	}
 	conf.Launcher = &V1Launcher{
-		Discovery: &V1ServiceDisc{
-			ServiceDisc: skyenv.DefaultServiceDiscAddr,
-		},
-		Apps:       nil,
-		ServerAddr: skyenv.DefaultAppSrvAddr,
-		BinPath:    skyenv.DefaultAppBinPath,
+		ServiceDisc: skyenv.DefaultServiceDiscAddr,
```
jdknives, line 36: @ersonp we do need to define a migration somewhere for that, right?

jdknives (changes requested):

pkg/visor/visorconfig/config.go
ersonp: I think we just need to update the already existing migration in `parse.go/updateUrls()`

pkg/servicedisc/client.go
mrpalide: Done!

pkg/visor/visorconfig/config.go
ersonp: Just tested by running a visor from this PR on config from `master`, and we do not need any extra migration. The already existing one takes care of it.

ersonp (approved):
Everything looks good.

Issue #860, opened 2021-08-12
Improve Systray App 


jdknives:
**Feature description**

The systray application currently does not support:

- stopping/starting the VPN 
- show VPN status (off/on/dialing)
- quick selection of VPN servers

It also does grey out the `uninstall` option if the Systray app is run manually (not package based). 

These features would be valuable for users. 

Concrete wireframes for these features are yet to be done.

PR #889, opened 2021-09-23
Mac installer

alexadhy:

For Skywire-Visor Release 0.5.0-rc*

 Changes:	
- Package skywire-visor as macos application

TODO:
- [x] Fix application not installed correctly 
- [x] Fix updating application
- [x] Fix application icon (not systray icon) not showing properly on `/Applications`
- [x] Also builds arm64 binaries

To Test:

- Run `make mac-installer` and run the installer via `open ./SkywireInstaller.pkg` or see https://github.com/skycoin/skywire/pull/889#issuecomment-925639466


alexadhy:
Instruction to run it:


https://user-images.githubusercontent.com/72057554/134482811-6ab95803-e3b7-465c-a914-65bf4e67f160.mp4


- Then Install it, It should ask for your password to run it (or your fingerprint) 
- To run vpn-client, do:

```bash
$ skywire-cli visor add-tp <YOUR_VPN_SERVER_PK> --type dmsg
$ skywire-cli visor start-app vpn-client
```
or via UI

Click VPN link menu, it should show:

https://user-images.githubusercontent.com/72057554/135220296-52fe233e-7f8c-4f7c-8d08-40b3d6823389.mp4


To create mac-installer with codesign, you need to:

1. Have valid Apple Developer Account
2. Export these environment variables:
```bash
 $ MAC_HASH_APPLICATION_ID="???"    #  Hash for Certificate for Application ID
 $ MAC_HASH_INSTALLER_ID="???" # Hash for Certificate Installer ID
 $ MAC_DEVELOPER_USERNAME="???"              # Apple Developer account email
 $ MAC_DEVELOPER_PASSWORD="???"             # Apple Developer account password
```
3. Run
```bash
$ ./scripts/mac_installer/create-installer.sh -s -n
```

alexadhy:
App Icon should be working just fine now:

Screen Shot 2021-09-23 at 22 12 14

scripts/mac_installer/create_installer.sh
```diff
@@ -0,0 +1,196 @@
...
+    security import "$cert_path" -k skywireBuild.keychain -P "$APP_CERTIFICATE_PASSWORD"
+  fi
+
+  test -x /usr/local/bin/gon || {
+    brew tap mitchellh/gon
+    brew install mitchellh/gon/gon
+  }
+
+  # build skywire binariea
+  make CGO_ENABLED=1 GOOS=darwin GOARCH="${go_arch}" build-systray
```
jdknives, line 60: We are not bundling any binaries other than Skywire-VPN client and visor, right? Maybe easier to build them explicitly.

alexadhy:
Last few commits fixes privileges on macos installer and application runner

jdknives (commented):
I have sent the error log of the installation on Element.

Two more points:

1. Is it possible to make Apple use the version we embed in the binary or do we have to set that somewhere dfiferently. 
2. There is an option to set a readme for the installer. What is that used for or how would the user pull that up on Apple?

alexadhy:
> I have sent the error log of the installation on Element.
> 
> 
> 
> Two more points:
> 
> 
> 
> 1. Is it possible to make Apple use the version we embed in the binary or do we have to set that somewhere dfiferently. 
> 
We can, I'll dynamically generate it / add it to the Info.plist

> 2. There is an option to set a readme for the installer. What is that used for or how would the user pull that up on Apple? 

We can via rtf files converted to b64 image iIrc, will investigate and do it if needed.

cmd/skywire-visor/commands/root.go
```diff
@@ -36,7 +37,7 @@ var uiAssets fs.FS
 var restartCtx = restart.CaptureContext()
 
 const (
-	defaultConfigName    = "skywire-config.json"
+	defaultConfigName    = "skywire-visor.json"
```
jdknives: This should not be changed. We should stick with `skywire-config` please.

jdknives (changes requested):

cmd/skywire-visor/commands/root.go
jdknives: OK. Some issue with the Github interface. My bad.

Makefile
```diff
@@ -174,5 +171,11 @@ build-ui: install-deps-ui  ## Builds the UI
 	mkdir ${MANAGER_UI_BUILT_DIR}
 	cp -r ${MANAGER_UI_DIR}/dist/. ${MANAGER_UI_BUILT_DIR}
 
+mac-installer: ## Create unsigned application
```
jdknives: I think we should make this create a signed installer and application. Otherwise there is not much of a point to the target as we will likely only release signed installers.

cmd/skywire-cli/commands/config/gen.go
```diff
@@ -62,12 +64,15 @@ var genConfigCmd = &cobra.Command{
 
 		//set output for package and skybian configs
 		if packageConfig {
-			if hypervisor {
-				output = "/opt/skywire/skywire.json"
+			var configName string
+			if runtime.GOOS == "linux" {
+				configName = "skywire-visor.json"
```
jdknives: Please name it `skywire-config.json`. We should try to get all installations methods homogenized with these values and technically we are exposing the hypervisor UI.

internal/gui/gui.go
```diff
@@ -44,23 +45,25 @@ var (
 )
 
 var (
+	mAdvancedButton *systray.MenuItem
 	mOpenHypervisor *systray.MenuItem
 	//mVPNClient      *systray.MenuItem
```
jdknives, line 49: I guess a lot of the commented variables here will be used for the advanced systray?

internal/gui/gui.go
```diff
@@ -44,23 +45,25 @@ var (
...
+	mVPNLink   *systray.MenuItem
 	mUninstall *systray.MenuItem
 	mQuit      *systray.MenuItem
 )
 
 // GetOnGUIReady creates func to run on GUI startup.
-func GetOnGUIReady(icon []byte, conf *visorconfig.V1) func() {
+func GetOnGUIReady(icon []byte, conf *visorconfig.V1, vis *visor.Visor) func() {
```
jdknives: I dont think we should pass around the visor. The VPN UI link can be available whether or not the application has started imo. Pretty sure the UI does not require the VPN to be started in order to run correctly.

scripts/mac_installer/create_installer.sh
```diff
@@ -0,0 +1,203 @@
...
+done
+
+# call build_installer twice, once for the original host architecture
+# and one for the other one (x86_64 and arm64 or vice versa)
+
+if [ "$staple_notarization" == false ]; then
+  build_installer
+
+  case ${go_arch} in
+  amd64)
```
jdknives, line 192: This seems to be incorrect. In case of `amd64` we set `go_arch` to `arm64`. Or am I missing anything here?

jdknives (approved):

internal/gui/gui.go
alexadhy: Yes, it will be used for #860 (connecting to vpn via systray, show available vpn servers, etc).

scripts/mac_installer/create_installer.sh
alexadhy: in the case of `amd64` we also create `arm64` and vice versa (we called the `build_installer` function twice basically. One for current arch and the other for the other arch.

jdknives (approved):

Issue #449, opened 2020-07-26
Normalize updates


0pcom:
**Feature description**

The current system of updating via the UI button does not update the system software.

It updates the skywire 'installation' which is a term I use loosely. This is not ok to do if you have installed skywire as a package, and instead of the desired effect of the updates happening the way you want them to, stuff is overwritten in an arbitrary way, and the package manager is disregarded.

For updates to the system software to be done correctly and as automatically as possible, there must remain room for manual intervention, which the current system does not facilitate and which is not an unexpected or irregular occurence.

###Outline of ideal update interaction:
* UI button is clicked to update an individual visor
* full dmsgpty terminal window opens in web browser and a full system update is initiated
* The system update completes with success and the window closes
* The visor should reboot if the linux kernel was updated

Most importantly, the window should not close if the exit status of the update command was not 0 so that the administrator can act to fix things. This is what is termed 'manual intervention' and it remains a fact of life on any linux distribution.

**Is your feature request related to a problem? Please describe.**

Totally solves all current installation and updating issues.
This relates to / encompasses #387

**Describe the solution you'd like**

The idea when phrased differently is to adapt [yay](https://github.com/Jguer/yay) to work on other linux distributions in some way or other. This is where the magic happens with dependency management. Yay is beautifuly written in golang and it is, I believe, the most widely used AUR package and the first one that most people who use Archlinux will install from the AUR.

To summarize the features specifically needed to create a reliable updater as I have described:

* Recognise what linux distribution it is running on
* Provide an interface to the system package manager
* interactive search/install
* Minimize user input
* Automate local package creation from [AUR](aur.archlinux.org) or custom repo hosting builds
* dependancy management

[yay](https://github.com/Jguer/yay) has most of this, and there are other tools which could be ported from arch to make creation of packages easier.

**Describe alternatives you've considered**

It would be an equally valid thing to do to eliminate the updater entirely and let the administrator install skywire via a package. The system could be updated normally from the dmsgpty terminal with instructions for how to do so

**Possible implementation**

If the dmsgpty terminal can open and run a command I can come up  with the specific implementation for updates which will encompass the existing packaging efforts.

jdknives:
@0pcom this has been solved with #911 I guess. Users are instructed to update the package themselves if they have installed via package managers and directed to the official ressources if they installed a package we provide.

jdknives:
Let me know any further concerns you may have.

jdknives:
Two more comments:

[...]

PR #891, opened 2021-09-23
Remove Transport Discovery heatbeat

ersonp:
yes

Fixes [#449-skywire-services](https://github.com/SkycoinPro/skywire-services/issues/449)

 Changes:	
- Removed `transport-discovery` heartbeat
- Removed redundant DeleteTransport code of hypervisor

How to test this PR:
Instructions given [here](https://github.com/SkycoinPro/skywire-services/pull/457).

### DIFF
--- a/pkg/transport/manager.go
+++ b/pkg/transport/manager.go
@@ -478,17 +478,6 @@ func (tm *Manager) DeleteTransport(id uuid.UUID) {
 
 	// Deregister transport before closing the underlying transport.
 	if tp, ok := tm.tps[id]; ok {
-		ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
-		defer cancel()
-
-		// todo: this should probably be moved to tp.close because we want to deregister
-		// a transport completely and not deal with transport statuses at all
-		if err := tm.Conf.DiscoveryClient.DeleteTransport(ctx, id); err != nil {
-			tm.Logger.WithError(err).Warnf("Failed to deregister transport of ID %s from discovery.", id)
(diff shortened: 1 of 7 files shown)


pkg/transport/manager.go
```diff
@@ -478,17 +478,6 @@ func (tm *Manager) DeleteTransport(id uuid.UUID) {
 
 	// Deregister transport before closing the underlying transport.
 	if tp, ok := tm.tps[id]; ok {
-		ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
-		defer cancel()
-
-		// todo: this should probably be moved to tp.close because we want to deregister
-		// a transport completely and not deal with transport statuses at all
```
jdknives, line 485: Are we deregistering anywhere else upon shutdown that I am missing? Seems like we are not removing tps from discovery at all if we delete this.

jdknives (approved):
Looks good to me.

pkg/transport/manager.go
ersonp: We are already deregistering inside tp.close as the todo is sugesting to do. [tp.close()](https://github.com/ersonp/skywire/blob/ef9b1dcd93c7573a30068ac17cba906a2e5de1b2/pkg/transport/managed_transport.go#L192)
[deleteFromDiscovery()](https://github.com/ersonp/skywire/blob/ef9b1dcd93c7573a30068ac17cba906a2e5de1b2/pkg/transport/managed_transport.go#L334) is called inside tp.close()

PR #894, opened 2021-09-23
Cleanup Makefile targets remove systray release targets

jdknives:

Yes

Fixes various outdated warnings when linting, remove obsolete target and remove systray release related targets and files. 

 Changes:	
- remove `lint-extra` as it does not provide any different linting than `lint`
- remove `sysroot`, `systray-snapshot` targets
- remove `goreleaser.yml` for systray release


ersonp (approved):
Looks good.

PR #901, opened 2021-09-30
Fix vpn client

### LINKED ISSUE #783: Invalid VPN authentication handled incorrectly
**Describe the bug**
The vpn client app is not connecting. I ran 2 visors in different virtual machines, using the last changes of the develop branch. When starting the vpn client app of the first visor to connect to the vpn server app of the second visor, the connection is never stablished. A few seconds after starting the app, the client and server apps show this in the console: `Error during close route group: close route group timed out: context deadline exceeded `. Eventually, the visor running the client app panics.

After that, the calls to the `GET /api/visors/{pk}` API endpoint return `Connection failed, reconnecting` in the `detailed_status` property of the vpn client app until the visor is stopped. A problem with this is that, at the same time, the `status` property returns `0`, which the UI interprets as the app being stoped, so the button for stopping the vpn is not shown. Maybe it does not make much sense to report that the app is stopped with the `status` property set to `0` and at the same time report that it is reconnecting with the `detailed_status` property set to `Connection failed, reconnecting`, but if there is a technical reason for that, maybe the UI can be adapted for considering the special case,

### LINKED ISSUE #900: Visor calls route finder without any transports established
**Describe the bug**
The visor currently calls the route finder to return routes for a connection requested by an application even if it does not have any transports established itself. Calling the visor for a route creation should return an error straight away that the visor does not have transports established and should not return the route finder error.

ersonp:
yes

Fixes #783 
Fixes #900

 Changes:	
- Vpn retirer only trys 3 times
- Added errors from server hello to retire whitelist
- Removed conSummarry check in AppStates()
- Added new app state `AppStatusErrored` : ` "status": 2` which gives the error inside `detailed_status`
- Added a transport check before dailing routes adapted from [here](https://github.com/alexadhy/skywire/commit/916f90c73d79a6660cd17a16cb77b42bc645d843)
- Added `RPCErr` to preserve the type of the error we return via RPC (as suggested by @alexadhy) 

How to test this PR:
Test #783 
1. Run visorA with VPN server and passcode
2. Run visorB create a transport to vpn-server and connect to Vpn-server of visor A with the wrong passcode.
3. Visor won't retry after the first try.
4. Check `http://localhost:8000/api/visors/<pk>/summary` and see if `vpn-client` has
```json
{
  "name": "vpn-client",
  "args": [
    "-srv",
    "02d27b250feea1fc2fca518baff5eb30df0d77d43fd2cc323d1a25be6ce3abc7be"
  ],
  "auto_start": false,
  "port": 43,
  "status": 2,
  "detailed_status": "Password didn't match"
}
```

Test #900
1. Run visorB without creating a transport to vpn-server and connect to Vpn-server of visor A.
2. Visor won't retry after the first try.
3. Check `http://localhost:8000/api/visors/<pk>/summary` and see if `vpn-client` has
```json
{
  "name": "vpn-client",
  "args": [
    "-srv",
    "02d27b250feea1fc2fca518baff5eb30df0d77d43fd2cc323d1a25be6ce3abc7be"
  ],
  "auto_start": false,
  "port": 43,
  "status": 2,
  "detailed_status": "transport not found"
}
```


alexadhy:
CMIIW, I actually think that the `DetailedStatus` can be better with it being a struct containing the actual status of the running application (the enums like the current one in `develop`, as well as the current error description like the one you have). Can be beneficial also for tests.

So we can eliminate `SetDetailedStatusError` and instead just use the `SetDetailedStatus` for everything.

ersonp:
`SetDetailedStatusError` and `SetDetailedStatus` are separate because both of them are stored and fetched from different places, i.e. `detailed_status` is fetched from proc whereas the `error` is fetched from `procManager` as the proc is deleted when the app stops after an error. But I think it would be better to just have `SetDetailedStatus` just like you have suggested.

ersonp:
@alexadhy I went through the code again and removing `SetDetailedStatusError`  would make the code needlessly complex so instead renamed all instances of `SetDetailedStatusError` to `SetError` which is more precise and easy to understand.

internal/vpn/client.go
```diff
@@ -188,16 +188,27 @@ func (c *Client) Serve() error {
...
 		if c.isClosed() {
 			return nil
 		}
 
 		if err := c.dialServeConn(); err != nil {
-			c.setAppStatus(ClientStatusReconnecting)
-			fmt.Println("Connection broke, reconnecting...")
-			return fmt.Errorf("dialServeConn: %w", err)
+			switch err {
+			case errHandshakeStatusForbidden, errHandshakeStatusInternalError, errHandshakeNoFreeIPs,
```
jdknives, line 205: Since the timeout is the only error we need to specifically check for redialing, would it be easier to check explicilty for `case errTimeout:`?

internal/vpn/client.go
```diff
@@ -326,6 +337,11 @@ func (c *Client) setupTUN(tunIP, tunGateway net.IP) error {
 func (c *Client) serveConn(conn net.Conn) error {
 	tunIP, tunGateway, err := c.shakeHands(conn)
 	if err != nil {
+		switch err {
+		case errHandshakeStatusForbidden, errHandshakeStatusInternalError, errHandshakeNoFreeIPs,
```
jdknives: Why dont we just return on all errors?

internal/vpn/client.go
```diff
@@ -445,6 +461,11 @@ func (c *Client) dialServeConn() error {
 	}
 
 	if err := c.serveConn(conn); err != nil {
+		switch err {
+		case errHandshakeStatusForbidden, errHandshakeStatusInternalError, errHandshakeNoFreeIPs,
+			errHandshakeStatusBadRequest:
```
jdknives: Same question as above. Why are we checking for these errors here explicitly.

internal/vpn/handshake_status.go
```diff
@@ -32,3 +34,20 @@ func (hs HandshakeStatus) String() string {
...
+	case HandshakeStatusBadRequest:
+		return errHandshakeStatusBadRequest
+	case HandshakeNoFreeIPs:
+		return errHandshakeNoFreeIPs
+	case HandshakeStatusInternalError:
+		return errHandshakeStatusInternalError
+	case HandshakeStatusForbidden:
+		return errHandshakeStatusForbidden
+	default:
+		return errors.New("Unknown code")
```
jdknives: Maybe `Unkown error code` is more explicit.

internal/vpn/handshake_status.go
```diff
@@ -32,3 +34,20 @@ func (hs HandshakeStatus) String() string {
 		return "Unknown code"
 	}
 }
+
+func (hs HandshakeStatus) getError() error {
```
jdknives, line 38: Maybe I missed it but I could not find any reason we need the `func (hs HandshakeStatus) String() string {}`.

pkg/visor/api.go
```diff
@@ -349,9 +350,22 @@ func (v *Visor) SetAppDetailedStatus(appName, status string) error {
 	return nil
 }
 
+// SetAppError implements API.
+func (v *Visor) SetAppError(appName, aErr string) error {
```
jdknives: Maybe `appErr` is clearer since we call it `appName` alreadt?

jdknives (commented):
Just a few questions from my end. It seems like there may be a better way to handle processes so we do not need to set status and errors separately but its not the task of the PR to rewrite the whole library. So apart from the questions this looks good.

internal/vpn/handshake_status.go
ersonp: We were previously using that to print the handshake errors. And we are still using it to print server response
This is with the `String()` func
```bash
[proc:vpn-client:5d27178f0d224432a8bc92f10338e4f2]: Got server hello: {Forbidden <nil> <nil>}Closed TUN
```
And this is without the `String()` func
```bash
[proc:vpn-client:32f4cb6eb0684582810ef0979e95b514]: Got server hello: {4 <nil> <nil>}Closed TUN
```

pkg/visor/api.go
ersonp: Yes, sounds good.

internal/vpn/handshake_status.go
ersonp: Okay.

internal/vpn/client.go
ersonp: Did that so we could send `"error during client/server handshake: %w"` this extra error line if not for the above ones. Will take it out as it seems unnecessarry.

internal/vpn/client.go
ersonp: For that, we will have to make a `WithErrBlacklist` instead, as we are only exiting the retrier if the error is whitelisted.

ersonp:
Should I add a max no of retries to the [retrier](https://github.com/ersonp/skywire/blob/abf6e3537c7539204a93681c568a1c9b8781fc15/internal/vpn/client.go#L119) of [`dialServer`](https://github.com/ersonp/skywire/blob/abf6e3537c7539204a93681c568a1c9b8781fc15/internal/vpn/client.go#L703) too?

internal/vpn/client.go
```diff
@@ -188,16 +182,32 @@ func (c *Client) Serve() error {
...
 
-	r := netutil.NewDefaultRetrier(c.log)
+	errTransportNotFound := appserver.RPCErr{
+		Err: rfclient.ErrTransportNotFound.Error(),
+	}
+
+	r := netutil.NewRetrier(c.log, netutil.DefaultInitBackoff, netutil.DefaultMaxBackoff, 3, netutil.DefaultFactor).
+		WithErrWhitelist(errHandshakeStatusForbidden).WithErrWhitelist(errHandshakeStatusInternalError).
+		WithErrWhitelist(errHandshakeNoFreeIPs).WithErrWhitelist(errHandshakeStatusBadRequest).
+		WithErrWhitelist(errTransportNotFound)
```
mrpalide: `WithErrWhitelist` accept multiple error, so why didn't add all once and use multiple `WithErrWhitelist` method? Something like this:
```Go
r := netutil.NewRetrier(c.log, netutil.DefaultInitBackoff, netutil.DefaultMaxBackoff, 3, netutil.DefaultFactor).
	WithErrWhitelist(
		errHandshakeStatusForbidden,
		errHandshakeStatusInternalError,
		errHandshakeNoFreeIPs,
		errHandshakeStatusBadRequest,
		errTransportNotFound,
	)
```

mrpalide (commented):
Functionality work as well as possible, and code was clear, and just one little comment.

internal/vpn/client.go
ersonp: That was an oversight on my part. Thanks for pointing that out.

pkg/router/router.go
```diff
@@ -218,6 +218,11 @@ func (r *router) DialRoutes(
 	lPK := r.conf.PubKey
 	forwardDesc := routing.NewRouteDescriptor(lPK, rPK, lPort, rPort)
 
+	ok := r.checkIfTransportAvailabel(rPK)
```
jdknives: This check should check if any transports are established for the visor and not if there is a transport between local and remote visor. Routes can be multihop and therefore this check in its current form is overly strict. Please change it and also fix the spelling issue with `Avalailable`.

jdknives: The error consequently would be more appropriately named `ErrNoTransportFound`.

jdknives (changes requested):

pkg/router/router.go
ersonp: I feel like `ErrTransportNotFound` should be renamed to something like `ErrFailedToCreateRoute` which is more informative and `ErrNoTransportFound` and `ErrTransportNotFound` feel too similar.

jdknives (approved):

PR #911, opened 2021-10-06
add build_tag to summary API endpoint

### LINKED ISSUE #910: Embed installation method into binaries
**Feature description**

Going forward we will need a way to discern Skybian installations and package based installations from each other. I would suggest we do that by embedding a tag in the binary for the visor. Upon building the release binaries, the macOS and Debian packages we build with a different tag and fetch that info from the visor and return it to the hypervisorUI.

I suggest we use the following tags

- `skybian` for the Github binary release as it is almost exclusively used there
- `macOS` for the MacOS installer
- `debian` for the Debian package
Depending on the tag returned to the frontend, the user should receive different instructions. Skybian installs continue with the regular updating as currently performed. MacOS installer installs and Debian package installs should receive a notification of  a new release but should not be able to perform the binary update we perform for Skybian. 

As raised initially in https://github.com/skycoin/skywire/issues/509, we need to prevent the binary updates from messing with the package based installs. 

We currently do something similar in Skybian by adding an env var to the systemd file. This should be retained as the actual Skybian version is also helpful for debugging. 

**Additional context**
@0pcom once we have this done for the packages, you can do something similar fairly easily for your AUR package. 

**Possible implementation**
We could use `ldflags` I think. Maye something else works as well.

mrpalide:

Fixes #910 

 Changes:	
- added new ldflags for build tag
- `skybian` tag added for goreleaser 

How to test this PR:
use `BUILDTAG=something` on `make` command. Same as `make build BUILDTAG=something`


pkg/visor/api.go
```diff
@@ -166,10 +167,15 @@ type Summary struct {
 	MinHops              uint16                           `json:"min_hops"`
 	PersistentTransports []transport.PersistentTransports `json:"persistent_transports"`
 	SkybianBuildVersion  string                           `json:"skybian_build_version"`
+	BuildTag             string                           `json:"build_tag"`
 }
 
+// buildTag variable that will set when building binary
```
jdknives: `buildTag variable set at compile time to indicate installation method of visor` may be clearer.

pkg/visor/api.go
```diff
@@ -313,6 +320,18 @@ func (v *Visor) SkybianBuildVersion() string {
 	return os.Getenv("SKYBIAN_BUILD_VERSION")
 }
 
+// BuildTagValidation implements API.
```
jdknives: Seems to be just a helper function.

ersonp:
I think we should print `build_tag` here https://github.com/mrpalide/skywire/blob/124d6f22869dca9c30440d28492b1bd77d2bd46d/cmd/skywire-visor/commands/root.go#L85

mrpalide:
> I think we should print `build_tag` here https://github.com/mrpalide/skywire/blob/124d6f22869dca9c30440d28492b1bd77d2bd46d/cmd/skywire-visor/commands/root.go#L85

Good point. Added in [`c8cc3c0`](https://github.com/skycoin/skywire/pull/911/commits/c8cc3c09e87a64ee6ee0028154b296e8dd5d26f8)

ersonp (approved):

jdknives (approved):

PR #914, opened 2021-10-07
Feature/config subcommand

### LINKED ISSUE #893: Add `config` subcommand to `skywire-cli`
**Feature description**

We have a range of commands that only target the config file and consequently are more appropriately named

`skywire-cli config gen`

or 

`skywire-cli config update`

etc.

alexadhy:

Fixes #893 

 Changes:	
- Config subcommand
- Update docs

How to test this PR:

- `make build`
- `skywire-cli config -h`


jdknives (changes requested):
Works as intended. While you are at it, can you make sure that we return an error and do not print the config when the input to `update -e` is unrecognized? Command does not throw an error if the environment name we pass is not recognized which makes it look like it succeeded.

alexadhy:
@jdknives will add check for it.

PR #916, opened 2021-10-08
Feature/improve health status

### LINKED ISSUE #879: Improve health status
We currently store health request responses in the visor to show the online status in the UI. The current method of just sending requests to `/healh` endpoint is suboptimal for users and for the deployment. For the users, simply querying the http endpoint is not a sufficient way to gauge the health of the service at any given time. 

Instead of sending out health requests periodically to all services, the visor should store and return the response of the last uptime tracker request as a proxy for online status. The frontend consequently should not list individual online statuses per service either.

alexadhy:

Fixes #879

 Changes:	
- [x] update internal state of visor whenever we made request to uptime-tracker update uptime of currently running visor.
- [x] returns it to the API 

How to test this PR:

- Run visor
- make request to `http://localhost:8000/api/visors/{pk}/summary`


jdknives (commented):
What I had in mind is that we simply set/unset the health value depending on whether or not `UpdateVisorUptime()` returns an error [here](https://github.com/jdknives/skywire/blob/3f92f2199faac8a82efd0fd2ef5333bc43328555/pkg/visor/init.go?plain=1#L593-L600). That seems to be much simpler.

Senyoret1:
@alexadhy @jdknives As far as I was able to check, now we would have 3 conditions. Visor offline (not returned by the hypervisor), visor online but with problems (returned by the hypervisor but with and empty `health` property) and visor online and working well (returned by the hypervisor and with `health: {services_health: true}` returned by the API). Am I right?

If that is the case, I’ll make the change in the UI, but the current procedure used by the UI for updating the data (calling the API every X seconds) will remain the same. If this is not enough for fixing #879, then we will need a web socket for getting the value.

Also, what the info message for the second case should be?

alexadhy:
@Senyoret1 there's actually just two cases of `Health` endpoint now:  `health: {services_health: false}`  and the `true` part. I forgot to initialize the false part, so I fixed it in https://github.com/skycoin/skywire/pull/919. Sorry @Senyoret1

Senyoret1:
@alexadhy under what conditions the API would return `health: {services_health: false}`?

alexadhy:
> @alexadhy under what conditions the API would return `health: {services_health: false}`?

@Senyoret1 That one is when the visor is initialized, but either it hasn't been able to update its uptime value to `uptime-tracker` service, or it's able to make the request, but the `uptime` service returned an error. You can check it under #919.

PR #922, opened 2021-10-12
Add stcpr dependency to PublicVisor

ersonp:
The commit commit contains addition to the dependency of the PublicVisor module. StcprClient is added as a dependency.

Fixes #	

 Changes:	
-	

How to test this PR:


pkg/visor/init.go
```diff
@@ -133,7 +133,7 @@ func registerModules(logger *logging.MasterLogger) {
 	pv = maker("public_visors", initPublicVisors, &tr)
```
ersonp: I think we should rename this module to `PublicAutoconnect`. @jdknives

jdknives: @ersonp go for it.

jdknives (approved):

PR #939, opened 2021-10-18
Improvements for the vpn client UI

Senyoret1:

 Changes:	
- The UI now shows the connection time.
- Some improvements were added to the coded which decides when to update the IP info.
- Now the vpn started UI shows the errors returned by the `/api/visors/{pk}/apps/vpn-client/connections` API endpoint.
- Various problems when no server has been selected were solved.
- A generic error msg was added for when the API does not return the details of an error.
- An error indication is now shown when there is a problem updating the status of the app.

How to test this PR:
- Connection time: select a server and start the vpn. The connection time will appear at the top of the current connection status. It is updated when the VPN is connected.
- IP info: the ip should be updated when entering and leaving the connected state.
- Errors returned while connected: not sure about how to tests this in the wild.
- Problems when there is no server selected: remove the server in the vpn-client section of the visor config. The UI should say that there is no server selected and you should be able to continue using the UI notmally (like selecting a server and connecting to it).
- Error indication: use the dev tools of the browser, open the network tab and block the network connections. A warning will appear at the top right corner of the UI.


Senyoret1:
The last commit fixes a problem which made the server name to go outside the box boundaries if it is too long.

Senyoret1:

- Now, when a server does not have a name, `Unnamed` is shown as the server name, instead of the server pk.

PR #945, opened 2021-10-19
Feature/improve redialing public autoconnect

### LINKED ISSUE #938: Improve redialing for automatic transports
**Describe the bug**
The redialing for automatic transport creation is too aggressive. After fetching the public visors from the service discovery, the client tries to redial the AR every second or so with an address it cannot resolve. This is the outcome of a faulty deregistration I believe. But whatever the reason, the client should not try to redial addresses the AR cant resolve. Instead it should fetch new visors or at least increase the redialing time.

alexadhy:

Fixes #938 

 Changes:	
- `servicedisc` client code already filtered out established transports
- All I do is added a map of addresses (`pk`s) with attempt(s) as its value, if the attempt to establish transport to the given `pk` returns an error, then we just record the `pk` and / or increment the attempt. Currently the maximum number of attempt is just 2, I can change it if it deemed necessary.
- We only try to establish 3 transport(s) at any given interval. 

How to test this PR:
- Run the `skywire-visor` with `public_autoconnect` field in the `transport` section defined to true
- See logs

### DIFF
--- a/pkg/servicedisc/autoconnect.go
+++ b/pkg/servicedisc/autoconnect.go
@@ -16,7 +16,7 @@ const (
 	// PublicServiceDelay defines a delay before adding transports to public services.
 	PublicServiceDelay = 10 * time.Second
 
-	fetchServicesDelay = 2 * time.Second
+	fetchServicesDelay = 5 * time.Second
 )
 
 // ConnectFn provides a way to connect to remote service
@@ -83,9 +83,9 @@ func (a *autoconnector) fetchPubAddresses(ctx context.Context) ([]cipher.PubKey,
 	if err := retrier.Do(fetch); err != nil {
 		return nil, err
 	}
-	var pks []cipher.PubKey
-	for _, service := range services {
-		pks = append(pks, service.Addr.PubKey())
+	pks := make([]cipher.PubKey, len(services))
+	for i, service := range services {
+		pks[i] = service.Addr.PubKey()
 	}
 	return pks, nil
 }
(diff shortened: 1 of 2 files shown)


pkg/servicedisc/autoconnect.go
```diff
@@ -46,25 +47,35 @@ func MakeConnector(conf Config, maxConns int, tm *transport.Manager, log *loggin
 }
 
 // Run implements Autoconnector interface
-func (a *autoconnector) Run(ctx context.Context) error {
+func (a *autoconnector) Run(ctx context.Context) (err error) {
+	// failed addresses will be populated everytime any
```
jdknives: Comment seems incomplete.

pkg/servicedisc/autoconnect.go
```diff
@@ -53,14 +53,22 @@ func (a *autoconnector) Run(ctx context.Context) (err error) {
...
 	for {
 		time.Sleep(PublicServiceDelay)
+
+		// successfully established transports
+		tps := a.tm.GetTransportsByLabel(transport.LabelAutomatic)
+
+		// don't fetch public addresses if there are more or equal to the number of maximum transport defined.
+		if len(tps) >= a.maxConns {
+			a.log.Infoln("autoconnect: maximum number of established transports reached: ", a.maxConns)
+			return err
```
jdknives, line 63: Can we make this a debug log? Not very useful in most cases. Also why are we returning `err` here instead of nil?

pkg/servicedisc/autoconnect.go
alexadhy: ah yeah, sorry will fix

pkg/servicedisc/autoconnect.go
alexadhy: we return named error btw, the value of `err` variable should be nil in this case.

jdknives (approved):

Issue #943, opened 2021-10-18
The vpn-client app reports errors after being closed normally


Senyoret1:
If the vpn-client app is started and stopped with `PUT /api/visors/{pk}/apps/vpn-client`, after stopping the app the `/api/visors/{pk}/summary` API endpoint starts returning `2` as the state of the app, indicating that there was an error. The API should return `0`.

PR #949, opened 2021-10-19
fix app stopping error status

### LINKED ISSUE #942: Details are needed for more errors the vpn-client app may have
When there is an error that stops the vpn-client app, the API returns a description about the error in the `detailed_status` field of the vpn-client app when calling `/api/visors/{pk}/summary`. The problem is that the field does not get a description after some errors, like when the server visor is closed while the client is connected to it.

It would be good to consider all the common problems and make sure the API returns error messages for them.

alexadhy:

Fixes #942 , #943

 Changes:	
- Fixes error when vpn-client is closed normally

How to test this PR:
- Start vpn-client normally
- Stop it normally as well.
- See #942 for details on how to run it. It shouldn't return `2` if there are no errors anymore


jdknives (approved):
Works well.

Senyoret1 (approved):
Worked well for me

PR #959, opened 2021-10-22
Fix/change update interval

jdknives:
Change intervals for fetching public visors.


alexadhy (approved):
Looks good to me, UI update is severely needed. Thanks

PR #963, opened 2021-10-23
fix defaults for skywire-cli config gen -p and -s flags

### LINKED ISSUE #962: `skywire-cli config gen -p` defaults
The -p flag of config generation was the matching specification to what is in the packages I maintain and it was created explicitly for the purpose.

since being changed to it's own subcommand, subsequent to my pull requests regarding this, the defaults have been changed in the following ways

tls_enable="true" changed from "false" correct way should be "false" matching skybian.

the default config file name was changed to skywire-config.json.

Not only was the name of the config file changed, but the `-o` flag has no effect any longer and the name cannot be changed with flags, tragically.

my temporary workaround at this point is to move the config to the name generated by default, run the skywire-cli command to update it, then move it back to the correct filename which is specifed by the systemd service.

**Steps to Reproduce**
```
skywire-cli config gen -ipro /etc/skywire.json
```

**Actual behavior**
config is written to /opt/skywire/skywire-config.json

**Desired behavior**
```
skywire-cli config gen -ipr
```
this needs to write to /opt/skywire/skywire.json

preferably, the flags can be integrated __even more__ as 'generating' and 'regenerating' should be the same process at the point the keys are retained.

so the command functionality I would like is:
```
skywire-cli config gen -ip
```
write to /opt/skywire/skywire.json
```
skywire-cli config gen -p
```
write to /opt/skywire/skywire-visor.json

the desired functionality would not be to diminish the `-o` flag and it should still be able to write a different path or filename, but the `-r` flag is always going to be used in this context, so it is preferably integrated.

**Environment information:**
- OS: Linux

**Additional context**
I have not crosschecked the behavior of the -o flag being ignored with the -s flag. I recommend the -s flag be kept consistent in it's behavior with -p while maintaining the desired defaults.

0pcom:
yes

Fixes #962

 Changes:	
-	correct defaults for `skywire-cli config gen -p` and `-s` flags
-	preserve `-o` flag functionality

How to test this PR:

observe the subtle differences in the default with the following flags in isolation and in combination

```
skywire-cli config gen -ipro skywire-test.json
```

the combination of `-ip` or `-ipr` will write / rewrite the config to `/opt/skywire/skywire.json`
`-p` or `-pr` without `-i` will write the config to `/opt/skywire/skywire-visor.json`
`-s` flag will write the config to `/etc/skywire-config.json`
`-o` flag can overwrite these default paths now


0pcom:
this should be ready now, I'm not sure why the commit history looks like it does.

PR #967, opened 2021-10-25
fix urls

mrpalide:

 Changes:	
- Remove `/` from end of URLs of test env.


mrpalide:
Handled by @alexadhy  on #965.

PR #968, opened 2021-10-25
Fix autoconnect retrial logic being too aggressive

jdknives:
Yes

Fixes an issue with the autoconnect logic being too aggressive in retrials. 

 Changes:	
- Fixed the autoconnect retrial logic by increasing backoff factor and number
of retrials.
- The (httpClient) Services() function now also returns an error if the length of
the resutls returned by the service discovery is 0.

How to test this PR:
Run `skywire-visor` against prod and check the logs.


alexadhy (approved):
LGTM

PR #984, opened 2021-11-01
Fix/sudph notify

### LINKED ISSUE #892: Sudph does not close properly
**Describe the bug**
`sudph` transport do not close properly. When one edge of the transport closes the transport the other edge does not get notified. Consequently it treats the transport as alive and may write to it or fail to act on the transport being down. This was first noted by @i-hate-nicknames in https://github.com/skycoin/skywire/pull/842

**Steps to Reproduce**

1. Open `sudph` transport
2. Close the `sudph` transport
3. Observe that the remote edge is not closing the transport on its edge

**Possible implementation**
This is likely the result of `sudph` using `kcp` library which may not support the needed shutdown behavior. Maybe it can be configured to do so.

alexadhy:

Fixes #892 	

 Changes:	
- timeout / deadline timer for closing transport (SUDPH)
- keepalive mechanism using no-op packet
- currently VPN doesn't work on sudph, not sure if it's because this package or not.

How to test this PR:
WIP


0pcom:
[Fix/sudph-notify-984](https://github.com/skycoin/skywire/tree/Fix/sudph-notify-984)

PR #1010, opened 2021-11-21
Advanced autoconnection

### LINKED ISSUE #950: Advanced autoconnection logic
**Feature description**

We should implement a more resilient and performant autoconnection logic for visors. This should include the ability to detect opportunities to connect to other visors via `stcpr` or `sudph` directly and use these one hop routes.  

We currently have a very basic autoconnection logic implemented. Upon startup we pull a number of public visors in and connect. This should create a number of transports that makes it easy enough to setup routes through the networkw without having to manually establish transports. The downside is that given the randomised logic behind picking servers to connect to, the routes established to conncet any two visors may be suboptimal and include multiple hops via Skywire where a single hop route would have been possible if further transports were established. `sudph` or `stcpr` transports directly to the remote are faster than mutli hop routes in most cases. 

Clients could cache the list of available remotes via `stcpr` and `sudph` from AR if a route to one of these remotes is requested, they could establish a transport before requesting the route, possibly even circumventing the need to call the route finder and directly call the setup node. Other implementations may be possible.

alexadhy:

Fixes #950 

 Changes:	
- On initializing visor, fetch available transports from `address-resolver`, cache it. Automatically sets a timer of one hour for the cache to update itself. 
- Added `RouteSetupHooks` to the router, which is an array of hook like function for setting up routes, visor injects these hooks during initialization (can also be added later)
- These hooks are then called during route creation process, which conforms to the spec defined in #950 

Notes:
- We still use routefinder to fetchRoutes during rule creation, these might be subpar, will add follow-up commits to it if necessary.

How to test this PR:
- WIP


pkg/router/router.go
```diff
@@ -1042,7 +1073,7 @@ func (r *router) removeRouteGroupOfRule(rule routing.Rule) {
 func (r *router) checkIfTransportAvalailable() (ok bool) {
 	r.tm.WalkTransports(func(tp *transport.ManagedTransport) bool {
 		ok = true
-		return true
+		return ok
```
ersonp, line 1078: We can just remove `ok = true` and `return true`.

alexadhy: regarding assignment to `ok` it is in the response parameter name, since we cannot return any value from `WalkTransport`

mrpalide: So why we assign OK as return value at all? I think non-assign return value and just its type is better.

mrpalide (commented):
Code is clean and clear.

pkg/router/router.go
alexadhy: @mrpalide this piece of code is not really important btw, also I'm not sure if there's any point in returning just the type, since it will mean we still assign bool value to a new variable which we will return, will be 2 more LoC.

ersonp (approved):
Everything looks good and the auto-connect is working as intended.

pkg/servicedisc/autoconnect.go
```diff
@@ -77,17 +78,26 @@ func (a *autoconnector) Run(ctx context.Context) (err error) {
...
 					failedAddresses[pk]++
 					continue
 				}
-				logger.Infoln("Added transport to public visor")
 			}
 		}
 	}
 }
 
+// tryEstablish transport will try to establish transport to the remote pk via STCPR or SUDPH, if both failed, return error.
```
jdknives, line 92: In its currnet form it does not try to establish an `sudph` transport it seems.

jdknives (approved):

jdknives (approved):

PR #1016, opened 2021-11-29
Start visor with hypervisor UI

### LINKED ISSUE #999: Start hypervisor with default values via flag
### Before the hypervisor and visor binaries were merged it was possible to explicitly start either a visor or hypervisor instance.

When this change happened, the systemd service for the hypervisor was removed from skybian.

In the AUR and debian packages of skywire which I maintain, I sought to preserve the two modes of operation by having systemd services which referenced a configuration file which was defined as either a visor or hypervisor configuration.

Now it seems there is some objection by @jdknives to supporting config generation which by default creates a file with a name which is differentiated based on the presence of a hypervisor configuration or remote hypervisors being set.

Another solution I have posited is to have a hypervisor instance started by a flag provided to skywire-visor. this would rmove the configuration for a local hypervisor out of the config file and basically generate this configuration with that flag 'on the fly' and run the local hypervisor UI instance when the visor is started.

this is a request to return a removed feature, in one sense.

I have already completed the multi-config approach with corresponding systemd services and it is currently used in the packages. I would ask for reconsideration from @jdknives on this, and I would take another different flag if the -p flag needs to serve other purposes.

a flag for skywire-visor to explicitly start the hypervisor UI with the visor

mrpalide:

Fixes #999 

 Changes:	
- add new flag `-f` to run visor with hypervisor UI

How to test this PR:
- generate a visor skywire-config by `./skywire-cli config gen`
- run visor by -f flag to use hypervisor UI too. `./skywire-visor -c skywire-config.json -f`

### DIFF
--- a/cmd/skywire-visor/commands/root.go
+++ b/cmd/skywire-visor/commands/root.go
@@ -28,6 +28,7 @@ import (
 	"github.com/skycoin/skywire/pkg/skyenv"
 	"github.com/skycoin/skywire/pkg/syslog"
 	"github.com/skycoin/skywire/pkg/visor"
+	"github.com/skycoin/skywire/pkg/visor/hypervisorconfig"
 	"github.com/skycoin/skywire/pkg/visor/logstore"
 	"github.com/skycoin/skywire/pkg/visor/visorconfig"
 )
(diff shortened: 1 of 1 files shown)


ersonp (approved):

jdknives:
@0pcom is that PR sufficient for your purposes?

PR #1043, opened 2022-01-07
fixing docker image push

dharmendrakariya:

Fixes #	

Docker build and push to ```skycoin/skywire``` dockerhub

 Changes:	
-	Makefile and appveyor CI

How to test this PR:

- Docker Username and Password need to be set on CI envs

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,8 +10,6 @@ VERSION := $(shell git describe)
 RFC_3339 := "+%Y-%m-%dT%H:%M:%SZ"
 COMMIT := $(shell git rev-list -1 HEAD)
 
-
-
 PROJECT_BASE := github.com/skycoin/skywire
 DMSG_BASE := github.com/skycoin/dmsg
 ifeq ($(OS),Windows_NT)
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -9,7 +9,7 @@
 VERSION := $(shell git describe)
 RFC_3339 := "+%Y-%m-%dT%H:%M:%SZ"
 COMMIT := $(shell git rev-list -1 HEAD)
-BRANCH := latest
+BRANCH := develop
```
mrpalide: I think it's better to remove this part, and add something like `DOCKER_TAG` to instead.
Also add `DOCKER_TAG=master` and `DOCKER_TAG=develop` on AppVeyor yaml file in deploy jobs.
The docker-build command should change to `./ci_scripts/docker-push.sh -t ${DOCKER_TAG} -b`

mrpalide (changes requested):

Makefile
dharmendrakariya: I changed the .appveyor.yaml!

mrpalide (approved):

Makefile
```diff
@@ -213,7 +211,7 @@ github-release:
 
 
 build-docker: ## Build docker image
-	./ci_scripts/docker-push.sh -t ${BRANCH} -b
+	./ci_scripts/docker-push.sh -t latest -b
```
ersonp, line 214: This needs to be `prod`

dharmendrakariya: I dont think so! that is in general docker build process! which should be tagged latest.

Issue #1050, opened 2022-01-12
`skywire-cli config gen` flag to enable RPC server


0pcom:
The auto-peering system for skybian requires the RPC sever to be available on the network from the running visor. There is not currently a way to generate this configuration.

```
{
	"version": "v1.1.0",
	"sk": "0374822cdfdc47c3f648aa828b3038132d8e511a00ac2bbd9a7ad4c85d6c5550",
	"pk": "02ecb0cf4fc3ce5fb1fb85c5e8f0034b1c3165c3e9727be41b1a44627795062ff9",
	"dmsg": {
		"discovery": "http://dmsgd.skywire.skycoin.com",
		"sessions_count": 1
	},
	"dmsgpty": {
		"dmsg_port": 22,
		"cli_network": "unix",
		"cli_address": "/tmp/dmsgpty.sock"
	},
	"skywire-tcp": {
		"pk_table": null,
		"listening_address": ":7777"
	},
	"transport": {
		"discovery": "http://tpd.skywire.skycoin.com",
		"address_resolver": "http://ar.skywire.skycoin.com",
		"public_autoconnect": true,
		"transport_setup_nodes": null
	},
	"routing": {
		"setup_nodes": [
			"0324579f003e6b4048bae2def4365e634d8e0e3054a20fc7af49daf2a179658557"
		],
		"route_finder": "http://rf.skywire.skycoin.com",
		"route_finder_timeout": "10s",
		"min_hops": 0
	},
	"uptime_tracker": {
		"addr": "http://ut.skywire.skycoin.com"
	},
	"launcher": {
		"service_discovery": "http://sd.skycoin.com",
		"apps": [
			{
				"name": "vpn-client",
				"auto_start": false,
				"port": 43
			},
			{
				"name": "skychat",
				"args": [
					"-addr",
					":8001"
				],
				"auto_start": true,
				"port": 1
			},
			{
				"name": "skysocks",
				"auto_start": true,
				"port": 3
			},
			{
				"name": "skysocks-client",
				"auto_start": false,
				"port": 13
			},
			{
				"name": "vpn-server",
				"auto_start": false,
				"port": 44
			}
		],
		"server_addr": "localhost:5505",
		"bin_path": "./apps"
	},
	"hypervisors": [],
	"cli_addr": "localhost:3435",
	"log_level": "info",
	"local_path": "./local",
	"stun_servers": [
		"45.118.133.242:3478",
		"192.53.173.68:3478",
[...]

PR #1052, opened 2022-01-12
new/update flags on skywire-cli

### LINKED ISSUE #1046: prevent erroneous hypervisor configuration
When the config is generated like this:

```
skywire-cli config gen --hypervisor-pks <visor-pk> -ro skywire-config.json
```

If the visor pk in question is the public key contained already in skywire-config.json then the above command should produce the same result as the following

```
skywire-cli config gen -iro skywire-config.json
```

That is, trying to set a visor to use it's own public key as hypervisor should generate the same configuration as intentionally generating the config with hypervisor enabled.

Currently this is not the behavior

need to default to producing a working config as much as possible. This is likely what someone intended to do if they did the first command.

This also represents an edge-case for auto configuration, when it might happen coincidentally or by happenstance. So it is preferable for this to be addressed.

mrpalide:

Fixes #1046 #1050 #1054 #1061

 Changes:	
- added `--public-rpc` flag to change CLIAddr from `localhost:3435` to `:3435`.
- added comparing step on adding hypervisors keys on `--hypervisor-pks <visor-pk>` flag, that if key equal to visor PK, then change this visor to hypervisor.
- added `--vpn-server-enable` flag to make vpn-server app autostart attribute *true*
- improve `--package` flag to distinguished between hypervisor and visor, and change its config output file name
- add a checking `--output` or `-o` flag, and if set was used then override `-p` and `-s` flag

How to test this PR:
- build at this PR by `make build`
- generate config with `./skywire-cli config gen --public-rpc`
- check for CLIAddr value, should be `:3435`
- add visor pk as hypervisor pks by `skywire-cli config gen --hypervisor-pks <visor-pk> -ro skywire-config.json`
- check config file that hypervisors should be empty `[]`, and hypervisor config part should added to config file.
- generate/update config with `./skywire-cli config gen -r --vpn-server-enable`
- check config file to confirm that vpn-server autostart is true
- generate config twice with `-pi` and `-p` to check their different name
- generate config with `-pir` or `-sr` with `-o config.json` flag to see override functionality


0pcom:
cli_addr should actually be ":3545"

I think it's equivalent to 0.0.0.0 for the ip address to be blank but the port is listed

jdknives (approved):

0pcom:
We need consistency in the default behavior for some things.

Is enable-auth to be considered default? If so, can you make sure it is evenly implemented, it was changed to not enabled by default in the -p flag but the default when you give no arguments is enabled

and can you put a flag for the config generation to `--disable-auth` also, if we go with the aforementioned default?

mrpalide:
As had conversation with Moses on matrix, we want these changes:
- [x] disable OS detection in runtime, and set flag instead
  -  as default set to `linux`, and can set by `--os windows` or `--os macos` flag.
- [x] disable auth in hypervisor as default, and be enabled by `-p` flag, or set OS to `windows` or `macos`
- [x] set `enable-auth` and `disable-auth` flag separately
- [x] remove `-s` flag completely
- [x] all apps be available as default, and remove each by flags
  - you should separate apps by `,`, same as `--disable-apps skysocks,vpn-server`

jdknives (approved):
Looks alright. Only need to double check we are not using the deprecated flag in Skybian still.

jdknives (approved):

PR #1060, opened 2022-01-15
Windows installer

### LINKED ISSUE #745: Create Skywire Windows Installer
**Feature description**

We will need to create a Skywire Windows installer to facilitate users setting up Skywire. It should be similar to the .deb package and MacOS installer in that it only contains:

- skywire-visor
- vpn-client
- skywire-cli

It should also make use of the Systray application built in #651.

We need to test whether there are any substantial features that do not run on Windows at this point and fix them.

0pcom:
no

Fixes #745 	

 Changes:	
- adds windows installer related files

How to test this PR:

#### build:

Follow the instructions in scripts/win_installer/README.md

#### installer:

* Download the [skywire windows installer](https://github.com/the-skycoin-project/skywire/releases/download/winstaller/skywire.msi)
* Install skywire
* Open cmd window and run `skywire`, or run `Skywire` directly by start menu shortcut

#### Current Shortcomings:
- ~~the executables are not by default in the PATH~~
- ~~the shortcut / launcher is blocked by the firewall~~
- ~~#651 is not integrated into this yet~~
- ~~user must generate their own configuration (no auto-config currently)~~
- ~~#1061~~


internal/gui/gui.go
```diff
@@ -192,27 +190,6 @@ func handleVPNLinkButton(conf *visorconfig.V1) {
 	}
 }
 
-// GetAvailPublicVPNServers gets all available public VPN server from service discovery URL
-func GetAvailPublicVPNServers(conf *visorconfig.V1) []string {
```
jdknives: Was supposed to be used for a quick connect functionality in the Systray application. So you could use the Systray directly without having to open the browser. Can likely be kept and used when we improve the Systray application.

jdknives (approved):
Looks good to me apart from one comment.

internal/gui/gui.go
mrpalide: Oh, I didn't know about it. So I'll added it back. 

Update:
Done. #e264ec5

ersonp (approved):

PR #1062, opened 2022-01-18
automigrate from direct url to dmsghttp in local visors

mrpalide:

Fixes skywire-services [#525](https://github.com/SkycoinPro/skywire-services/issues/525)

 Changes:	
- add MigrateLocalVisor process during starting visor process
- add new field `connection_type` on config file to distinguish between `direct` (normal URL) and `dmsghttp` connection type

How to test this PR:
- generate a config without `-d` flag, check connection_type
- generate a config / update current config with `-d` flag and check connection_type again


jdknives:
Will close as we should move that to the Debian package together with other migration logic we currently have.

PR #1076, opened 2022-01-31
Migrate update logic to Debian package

### LINKED ISSUE #1070: Move migration code into Debian scripts
**Feature description**

We currently handle config migrations and other upgrading procedures in the Skywire Visor binary as we previously did not have a pacakging system available to do these checks with. That is less than optimal and should be moved out of the binary and into the Debian scripts as it is easier to handle and should be less error prone hopefully.

mrpalide:

Fixes #1070 

 Changes:	
- Remove old logic, based on downloading compressed version from github release page
- Add new logic to update from Debian package repository	

How to test this PR:
- Burn fresh skybian image on SD card
- Clone this PR, and build binaries by `GOOS=linux GOARCH=arm64 GOARM=6 make build BUILDTAG=linux_arm64`. (I use `arm64` because of my board RPi4, you should use your own board architect).
- Replace SD binaries files with new built binaries. Following files should replaced in `/usr/bin`:
  - `skywire-cli`
  - `skywire-visor`
  - `setup-node`
  - `apps/*`
- Start board
- Open hypervisor UI and Click on update
- After complete and restarting service, check version of visor, should be 0.5.1.


pkg/util/updater/updater.go
```diff
@@ -207,321 +185,53 @@ func (u *Updater) getVersion(updateConfig UpdateConfig) (string, error) {
...
-	downloadedBinaryPath := filepath.Join(downloadedBinariesPath, binary)
-	if _, err := os.Stat(downloadedBinaryPath); os.IsNotExist(err) {
-		downloadedBinaryPath = filepath.Join(downloadedBinariesPath, appsSubfolder, binary)
-	}
-
-	if _, err := os.Stat(downloadedBinaryPath); os.IsNotExist(err) {
-		u.log.Warnf("%v is not found in update, skipping", binary)
-		return nil
+func (u *Updater) addRepo() error {
+	if err := exec.Command("sudo", "add-apt-repository", "deb http://176.9.28.105 sid main").Run(); err != nil {
```
jdknives: Can we use the domain here? It is deb.skywire.skycoin.com

pkg/util/updater/updater.go
```diff
@@ -207,321 +185,53 @@ func (u *Updater) getVersion(updateConfig UpdateConfig) (string, error) {
...
-		if err := capturer.Release(); err != nil {
-			u.log.Errorf("Failed to release output: %w", err)
-		}
-	}()
-
-	out := io.MultiWriter(f, u.progressBar(stdoutWriter, resp.ContentLength, filename))
-
-	if _, err := io.Copy(out, resp.Body); err != nil {
-		return "", err
+func (u *Updater) restartBoard() {
```
jdknives: Is a restart needed?

jdknives (changes requested):
I would also like to move the config migration code to the debian package.

pkg/util/updater/updater.go
mrpalide: Mh, I used URLs in wiki page that Moses created. I'll change it.

pkg/util/updater/updater.go
mrpalide: Yes, I thought, based on logic of Moses Debian package. I'll check it with him again.

0pcom:
> I would also like to move the config migration code to the debian package.

@jdknives  you are referring to issue [#1070](https://github.com/skycoin/skywire/issues/1070)?
Is there something I am missing with the current skywire-autoconfig script? The configs are generated or regenerated on every update.

@mrpalide why are we doing repository configuration here? Configuring the repository initially is probably not desirable for users if it was not already configured. They probably downloaded it directly.

the `apt` command itself will tell you that it is not stable to use in this way..
```
    WARNING : apt does not have a stable CLI interface. Use with caution in scripts.
```

In the second place
here:
https://github.com/mrpalide/skywire/blob/88fd12196f3b45aa0fea98c7a2a73cf746da167d/pkg/util/updater/updater.go#L132

nothing else will run after that. The function won't return anything, the process will be killed. A new instance _may_ start but there is no way to know and not a good way to test.

https://github.com/mrpalide/skywire/blob/88fd12196f3b45aa0fea98c7a2a73cf746da167d/pkg/util/updater/updater.go#L125

you are uninstalling the program that is currently executing this command. the process will be killed.

pkg/util/updater/updater.go
mrpalide: I change it to restart service, not the board at all.

pkg/util/updater/updater.go
mrpalide: Done.

ersonp (commented):
The code looks good. Can't test it as I have misplaced my RPi.

PR #1087, opened 2022-02-08
Update documentation & other small fixes

0pcom:
Changes:	
* documents the current state of the command-line functionality for skywire-cli and visor binaries + example config files
* fix `make run-source`
* match the default config (path) used by the visor with the default config path generated by the cli
* fix the behavior of skywire-cli config-gen -p to match the config names used by scripts in the package

### DIFF
--- a/cmd/skywire-cli/README.md
+++ b/cmd/skywire-cli/README.md
@@ -54,6 +54,7 @@ Usage:
   skywire-cli [command]
 
 Available Commands:
+  completion  Generate completion script
   config      Contains sub-commands that interact with the config of local skywire-visor
   help        Help about any command
   mdisc       Contains sub-commands that interact with a remote DMSG Discovery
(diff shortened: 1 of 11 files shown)


cmd/skywire-cli/README.md
```diff
@@ -166,7 +168,9 @@ Available Commands:
   set-app-autostart Sets the autostart flag for an app of given name
   start-app         Starts an app of given name
   stop-app          Stops an app of given name
+  summary           Obtains summary of visor information
   tp                Returns summary of given transport by id
+  update            Obtains summary of visor information
   version           Obtains version and build info of the node
```
mrpalide: I think we should change `update` description.

mrpalide (changes requested):

mrpalide (approved):

PR #1092, opened 2022-02-08
migrate to dmsghttp on binaries

### LINKED ISSUE #1070: Move migration code into Debian scripts

mrpalide:

Fixes #1070 

 Changes:	
- Add migration to dmsghttp by checking local location in skywire-visor running

How to test this PR:
1. Open `https://ipinfo.io/country`
2. Change line 44 in `/internal/netutil/netutil.go` to result of step 1
3. Build binaries
4. Run `./skywire-cli config gen`
5. Run `./skywire-visor -c skywire-config.json`
6. Check `skywire-config.json` and see changes on addresses.

### DIFF
--- a/internal/netutil/netutil.go
+++ b/internal/netutil/netutil.go
@@ -1,6 +1,10 @@
 package netutil
 
-import "net"
+import (
+	"io"
+	"net"
+	"net/http"
+)
 
 // LocalAddresses returns a list of all local addresses
 func LocalAddresses() ([]string, error) {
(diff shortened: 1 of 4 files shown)


internal/netutil/netutil.go
```diff
@@ -26,3 +30,19 @@ func LocalAddresses() ([]string, error) {
 
 	return result, nil
 }
+
+// LocalProtocol check a condition to use dmsghttp or direct url
+func LocalProtocol() bool {
+	resp, err := http.Get("https://ipinfo.io/country")
```
ersonp, line 36: As this `ip.skycoin.com` is used everywhere for location data, I think we should use it here too.

ersonp:
Migration is working, we need to deploy `dmsghttp` enabled services and add their PK's to `dmsghttp-config.json` as currently it has dummy values `dmsg://pk:port`.

internal/netutil/netutil.go
mrpalide: I didn't check our service, but I checked `ipinfo.io` by some guys in our mainnet telegram group and work well there.
We can change ours if can confirm that work fine there.

ersonp: I forgot about that issue. I think it's better to just keep it `ipinfo.io`.

jdknives (approved):

PR #1108, opened 2022-02-22
Fix logic of `-o` flag in generate config

mrpalide:

Fixes the problem with using `p` and `o` flag together.

 Changes:	
- Change condition logic of `o` flag

How to test this PR:
- `make build`
- `./skywire-cli config gen -pio skywire-config.json`
- checking locations in `skywrie-config.json` file


0pcom:
looks good, nice work

PR #1114, opened 2022-02-25
fix dmsghttp-config.json file path in skywire config

mrpalide:

 Changes:	
- set fix path for `dmsghttp-config.json` on `skywire-config.json`.

How to test this PR:
- `make build`
- `./skywire-cli config gen -d` to check `dmsghttp_path` field on generated configuration


ersonp (approved):

ersonp:
Just a small question; are we using the path from the config anywhere? and if yes then just add a comment [here](https://github.com/skycoin/skywire/blob/34c8ed7e49123b19bbea275baad1ea36bd9549e2/pkg/visor/visorconfig/v1.go#L39).

jdknives (approved):

PR #1151, opened 2022-04-11
various small fixes

0pcom:
Changes:	
-	fix hidden flags nil pointer error on mac & windows ( --all flag error)
-	optimized `run-source` & `run-vpnsrv` makefile directives and related subdirectives
-	fix visor read config from stdin for systray

On that last one, runVisor now can accept a config passed to it as input (or nil). Running with systray was trying to read the config a second time, calling for initConfig . When the config comes via stdin there is no second read possible.

Additions:
-	`prepare` and `prepare-systray` makefile directives
-	`run-systray` makefile directive
-	`run-systray-test` makefile directive

How to test this PR:
```
make run-systray-test
```


0pcom:
In tests I discovered that the systray app is not appropriate to run as root on linux and does not function as intended. It possibly could but it cannot directly launch the browser as root. I have disabled systray if the root user is detected on macOS or linux.

privilege elevation for the vpn client / server is still yet to be addressed.

local/.gitignore
```diff
@@ -0,0 +1,4 @@
+# Ignore everything in this directory
```
ersonp: We are already ignoring the contents [here](https://github.com/the-skycoin-project/skywire/blob/b158522464cfabdf176898b8454628c5b5a8a34d/.gitignore#L16) and I  also do not see the reason why we need to push an empty folder to the repo. Since after starting the visor with sudo we need to delete the folder again anyway to start a visor without sudo.

ersonp:
> In tests I discovered that the systray app is not appropriate to run as root on linux and does not function as intended. It possibly could but it cannot directly launch the browser as root. I have disabled systray if the root user is detected on macOS or linux.
> 
> privilege elevation for the vpn client / server is still yet to be addressed.

we need to set `cap_net_admin` by `sudo setcap 'cap_net_admin+p' ./apps/vpn-client` and `sudo setcap 'cap_net_admin+p' ./apps/vpn-server`

andi think we should provide a `WARN` log in case if the systray app is started as root.

cmd/apps/skysocks-client/skysocks-client.go
```diff
@@ -28,9 +30,78 @@ const (
...
+func init() {
+	rootCmd.Flags().SortFlags = false
+
+	rootCmd.Flags().StringVarP(&addr, "addr", "a", skyenv.SkysocksClientAddr, "Client address to listen on")
+	rootCmd.Flags().StringVarP(&serverPK, "srv", "q", "", "PubKey of the server to connect to")
+}
+
+var rootCmd = &cobra.Command{
+	Use:   "skysocks",
+	Short: "Skywire SOCKS5 Proxy Server",
```
mrpalide: Seems short description is wrong here

cmd/apps/skychat/skychat.go
```diff
@@ -34,62 +35,95 @@ const (
...
+func init() {
+	rootCmd.Flags().SortFlags = false
+	rootCmd.Flags().StringVarP(&addr, "addr", "a", ":8001", "address to bind")
+}
 
-//go:embed static
-var embededFiles embed.FS
+var rootCmd = &cobra.Command{
+	Use:   "skysocks",
+	Short: "Skywire SOCKS5 Proxy Server",
```
mrpalide: Should be skychat, not skysocks.

cmd/apps/vpn-client/vpn-client.go
```diff
@@ -19,142 +21,193 @@ import (
...
+	thisUser, err := user.Current()
+	if err != nil {
+		panic(err)
 	}
-
-	serverPK := cipher.PubKey{}
-	if err := serverPK.UnmarshalText([]byte(*serverPKStr)); err != nil {
-		fmt.Printf("Invalid VPN server pub key: %v\n", err)
+	if (thisUser.Username != "root") && ((skyenv.OS == "linux") || (skyenv.OS == "mac")) {
+		fmt.Println("vpn client must be run as root")
```
mrpalide: No, not root.
We handle administrator privilege in Windows Installer, and in Mac or Linux, need set `net_cap_admin` value for `vpn-client` binary.

cmd/apps/vpn-client/vpn-client.go
```diff
@@ -19,142 +21,193 @@ import (
...
+	Short: "Skywire VPN Client",
+	Long: `
+	┬  ┬┌─┐┌┐┌   ┌─┐┬  ┬┌─┐┌┐┌┌┬┐
+	└┐┌┘├─┘│││───│  │  │├┤ │││ │
+	 └┘ ┴  ┘└┘   └─┘┴─┘┴└─┘┘└┘ ┴ `,
+	Run: func(_ *cobra.Command, _ []string) {
+		if serverPKStr == "" {
+			// TODO(darkrengarius): fix args passage for Windows
+			//*serverPKStr = "03e9019b3caa021dbee1c23e6295c6034ab4623aec50802fcfdd19764568e2958d"
+			fmt.Println("VPN server pub key is missing")
```
mrpalide: Actually never read key from config and told `VPN server pub key is missing`, even when I set it.
```
[2022-04-12T12:58:01+04:30] INFO (STDOUT) [proc:vpn-client:db6f0128f6da48b49f298d6b4d0df125]: VPN server pub key is missing
```

cmd/apps/skychat/skychat.go
```diff
@@ -34,62 +35,95 @@ const (
...
+		defer appC.Close()
+
+		if _, err := buildinfo.Get().WriteTo(os.Stdout); err != nil {
+			fmt.Printf("Failed to output build info: %v", err)
+		}
 
-func main() {
-	appC = app.NewClient(nil)
-	defer appC.Close()
+		fmt.Print("Successfully started skychat.")
```
mrpalide: Skychat never run, and get this confusing log
```
[2022-04-12T12:53:23+04:30] INFO (STDOUT) [proc:skychat:140eb2010c7d43128ce42e3e7be6af34]: Successfully started skychat.Serving HTTP onddrlisten tcp: address ddr: missing port in address
```
Seems cannot read from config, same as vpn-client.

mrpalide (changes requested):

ersonp:
One small change is needed for skywire windows. In windows `config gen` the only app that needs to be included is `vpn-client` as all other apps are not supported.

local/.gitignore
0pcom: This was something that @jdknives mentioned doing last we spoke. To address the permissions errors between runs as root and not. The issue still remains, with all the files in the folder though not the folder itself. so perhaps that is why it wasn't done. This is an issue I am attempting to address more thoroughly. Doesn't work as intended - will remove.

0pcom:
On the systray app I changed the behavior from being disabled as root to only hiding the buttons that launch the browser when the process is run as root.

 on the config gen @ersonp I changed it to detect if the applications are at the bin path, and if not they are added to the list of apps to disabled
 
 exceptions to this behavior are in two instances;

* the config is being written to stdout (`-n` flag)
* the cli command was started with `go run`

most of the other stuff mentioned has been cleaned up

local/.gitignore
0pcom: removed

cmd/apps/skysocks-client/skysocks-client.go
0pcom: changed

cmd/apps/skychat/skychat.go
0pcom: changed

cmd/apps/vpn-client/vpn-client.go
0pcom: reverted to original state

cmd/apps/vpn-client/vpn-client.go
0pcom: check it again, I wasn't able to reproduce this after my changes

cmd/apps/skychat/skychat.go
0pcom: this was related to the short command being wrong, per your previous comment

0pcom:
I've prevented writing to non-root owned path as root and vice versa. With skywire-cli config gen and with skywire-visor

@ersonp I had to revert the `-o` flag for config gen to the way I had it before your recent PR.

At issue is flags.Changed not working right.. my workaround was to have the output be actually unset and set the value after it's detected that it contains an empty string, as well as to record that it was unset into a boolean to be used for overriding the output of the `-p` flag default config path

I was otherwise unable to get the `-o` flag to override the `-p` flag for config gen.

Tomorrow I need to add a flag which will generate one default config at a set location in the user space, which references the users.db and local folder along the same path as the config.

cmd/apps/vpn-client/vpn-client.go
0pcom: i've reverted the apps back to how they are on develop, which should address this. I did reproduce it at one point but seems ok now.

ersonp:
Can we just limit the scope of this PR to flags and reading from STDIN because the changes with the path is not actually solving the problem but just shifting it. I think it would be better if we fixed it via the read and write functions of all the files in the `/local` folder. If I remember correctly this issue was solved previously and only reappeared a little while ago (as far as I can tell). So I think it would be better if we took a deeper look at it in a different PR.

0pcom:
@ersonp I've changed it to warn about writing with elevated permissions here:

image

I want to maintain the failure in the reverse instance as not failing means it runs but doesn't work, and just repeats "waiting for RPC get ready"

image

cmd/skywire-cli/commands/config/gen.go
```diff
@@ -261,6 +335,32 @@ var genConfigCmd = &cobra.Command{
 		}
 		//don't write file with stdout
 		if !stdout {
+			thisUser, err := user.Current()
```
jdknives: Something like `userLvl` or `owner` etc sounds better.

jdknives: Something like `userLvl` or `owner` etc sounds better.

cmd/skywire-cli/commands/config/gen.go
```diff
@@ -261,6 +335,32 @@ var genConfigCmd = &cobra.Command{
...
+			if err != nil {
+				panic(err)
```
jdknives: What errors could this return? Are they clear enough in themselves? Should we exit cleanly and log something here instead?

jdknives: What errors could this return? Are they clear enough in themselves? Should we exit cleanly and log something here instead?

ersonp (approved):

internal/gui/gui.go
```diff
@@ -64,24 +64,36 @@ var (
...
-		systray.SetTemplateIcon(icon, icon)
-		systray.SetTooltip("Skywire")
-
-		initOpenVPNLinkBtn(conf)
-		initAdvancedButton(conf)
-		initVpnClientBtn(conf, httpC, logger)
-		initQuitBtn()
-
-		go handleUserInteraction(conf, doneCh)
+	if checkRoot() {
```
jdknives: `isRoot` maybe clearer?

internal/gui/gui.go
```diff
@@ -417,6 +438,24 @@ func handleUserInteraction(conf *visorconfig.V1, doneCh chan<- bool) {
 	}
 }
 
+func handleRootInteraction(conf *visorconfig.V1, doneCh chan<- bool) {
+	for {
+		select {
+		//		case <-mOpenHypervisor.ClickedCh:
```
jdknives, line 444: What are these commented cases doing here?

internal/gui/gui_unix.go
```diff
@@ -3,8 +3,23 @@
...
+
+	"github.com/skycoin/skywire/pkg/util/osutil"
+)
 
 func platformExecUninstall() error {
 	return osutil.Run("/bin/bash", "-c", deinstallerPath)
 }
+
+func checkRoot() bool {
+	thisUser, err := user.Current()
```
jdknives: Same suggestion as above. Using `this` sounds odd to me.

internal/gui/gui_unix.go
```diff
@@ -3,8 +3,23 @@
...
+	if err != nil {
+		panic(err)
```
jdknives: I would prefer to not panic here and log a fatal err with some context.

pkg/skyenv/values.go
```diff
@@ -169,3 +169,9 @@ func Version() string {
 	}
 	return v
 }
+
+// HomePath gets the current user's home folder
+func HomePath() string {
```
jdknives, line 174: What if this returns an error? Will the visor start? If so, where will it start?

pkg/skyenv/values_darwin.go
```diff
@@ -22,3 +19,13 @@ func PackageConfig() PkgConfig {
 	pkgconfig.Hypervisor.EnableAuth = true
 	return pkgconfig
 }
+
+// UserConfig contains installation paths (for mac)
+func UserConfig() PkgConfig {
+	var usrconfig PkgConfig
+	usrconfig.Launcher.BinPath = "/Library/Application Support/Skywire/apps"
+	usrconfig.LocalPath = HomePath() + "/.skywire/local"
+	usrconfig.Hypervisor.DbPath = HomePath() + "/.skywire/users.db" //permissions errors if the process is not run as root.
```
jdknives: What are you trying to say with the comment? It seems it was copied from previous implementation for `opt`. Does it indeed fail if not run as root in HOME?

pkg/skyenv/values_windows.go
```diff
@@ -21,3 +19,13 @@ func PackageConfig() PkgConfig {
 	pkgconfig.Hypervisor.EnableAuth = true
 	return pkgconfig
 }
+
+// UserConfig contains installation paths (for windows)
+func UserConfig() PkgConfig {
+	var usrconfig PkgConfig
+	usrconfig.Launcher.BinPath = "C:/Program Files/Skywire/apps"
+	usrconfig.LocalPath = HomePath() + "/.skywire/local"
+	usrconfig.Hypervisor.DbPath = HomePath() + "/.skywire/users.db" //permissions errors if the process is not run as root.
```
jdknives: Same question about the validity of the comment as above.

jdknives (approved):

cmd/skywire-cli/commands/config/gen.go
0pcom: no need to really error here & Ive changed it to fail transparently. This is only checking for root and it will just not change the default from false to true.

internal/gui/gui.go
0pcom: done

internal/gui/gui_unix.go
0pcom: done

internal/gui/gui_unix.go
0pcom: the logging overhead is unnecessary and it doesn't really need to fail, only to return true if root is detected.

this check is for limiting the available  buttons in the systray to ones that are appropriate to launch by the current user. The browser won't be allowed to run as root in the userspace by the system, typically.

pkg/skyenv/values.go
0pcom: You are correct, this could cause an odd configuration to be produced if HomePath returns an empty string.

I've now implemented a check on the config gen command to not offer the `-u --user` flag if HomePath returns an empty string. Similarly to how the visor determines if the -u flag should be an option based on the presence of the config or not.

if HomePath returned empty before these changes it _would have been_ obvious from the config gen help menu as it's displayed in the flag description.

pkg/skyenv/values_darwin.go
0pcom: accidental inclusion, and the original comment issue is now basically resolved

pkg/skyenv/values_windows.go
0pcom: accidental inclusion, and the original comment issue is now basically resolved

mrpalide (approved):

PR #1155, opened 2022-04-17
Change `-p --pkg` flags for config gen and visor

0pcom:
A single config filename is now produced by:

```
skywire-cli config gen -p
```

the same config file can be used by the visor with the same flag
```
skywire-visor -p
```
the config name can be set independently for the different operating systems, similarly to the paths.

It's not my intention to force the use of skywire.json as the config file name which is used when skywire is run with the -p flag as root. 

The Configjson constant in skyenv values_linux.go is currently used in the linux packages. Maintainers of the batch files / scripts for windows and macOS can use a different name in values_windows and values_darwin.

This PR also encompasses a bit of cleanup to the logging and moving the generation of the config update string on incompatible versions and the initial loggng to their own functions

How to test this PR:

```
skywire-cli config gen --help
sudo skywire-cli config gen -p
sudo skywire-visor --help
sudo skywire-visor -p
```


mrpalide:
I have problem here, actually maybe I missing something but...
This is my `/opt/skywire` folder that is empty
image
when I use `skywire-cli config gen -p` get `Config file already exists` fatal, and when I add `-r` flag, get `directory not own by root` fatal error.
image
but my `/opt/skywire` owner is root
image
I missing something or not?

pkg/skyenv/values_darwin.go
```diff
@@ -6,13 +6,14 @@ package skyenv
 const (
 	//OS detection at runtime
 	OS = "mac"
-
 	// SkywirePath is the path to the installation folder.
 	SkywirePath = "/Library/Application Support/Skywire"
+	// Ptext is the description to use for the -p flag for config gen and for the visor
+	Ptext = "use mac installation path: " + SkywirePath
+	// Configjson is the config name generated by the script included with the installation on mac
+	Configjson = "skywire.json"
```
jdknives, line 14: What is the point of us introducing another name for the config file? We had `skywire-config.json` so far as default path.

pkg/skyenv/values_darwin.go
```diff
@@ -6,13 +6,14 @@ package skyenv
 const (
 	//OS detection at runtime
 	OS = "mac"
-
 	// SkywirePath is the path to the installation folder.
 	SkywirePath = "/Library/Application Support/Skywire"
+	// Ptext is the description to use for the -p flag for config gen and for the visor
+	Ptext = "use mac installation path: " + SkywirePath
```
jdknives, line 12: Seems redundant to me. Can we just use the SkwirePath var we already export and the OS we can get from the stdlib?

jdknives (changes requested):

pkg/skyenv/values_darwin.go
0pcom: it was intentionally set like that to differentiate the config for the package from the default

0pcom: this was not introduced, it's been like this in the linux packages for over a year.

While it seems like a trivial thing to just use the default config name, it would require at this point another layer of config migration logic, and I have enough to do at this point for the packages and images to change that currently.

ersonp:
The `-p` flag in skywire-visor is not working for me.
image

skywire-conf worked
image

mrpalide:
My last problem with `-p` flag remain yet :(

image

0pcom:
here I've corrected the behavior @mrpalide 
image

this is an appropriate failure when run without sudo
image

0pcom:
@ersonp I've fixed the error you cited here
image

cmd/skywire-cli/commands/config/gen.go
```diff
@@ -149,39 +181,70 @@ var genConfigCmd = &cobra.Command{
...
+				}
+				owner, err := script.Exec(`stat -c '%U' ` + confPath1).String()
+				if err != nil {
+					logger.Error("cannot stat: " + confPath1)
+				}
+				rootOwner, err := script.Exec(`stat -c '%U' /root`).String()
+				if err != nil {
+					logger.Error("cannot stat: /root")
+				}
+				if (owner != rootOwner) && root {
```
jdknives: We are leaving a warning here only. Comment above seems to indicate we would fail, right?

cmd/skywire-cli/commands/config/gen.go
```diff
@@ -149,39 +181,70 @@ var genConfigCmd = &cobra.Command{
...
+			if _, err = exec.LookPath("stat"); err == nil {
+				confPath1, _ := filepath.Split(confPath)
+				if confPath1 == "" {
+					confPath1 = "./"
+				}
+				owner, err := script.Exec(`stat -c '%U' ` + confPath1).String()
+				if err != nil {
+					logger.Error("cannot stat: " + confPath1)
+				}
+				rootOwner, err := script.Exec(`stat -c '%U' /root`).String()
```
jdknives: This will not work on MacOS or Windows, right? Not sure about their directory structure.

cmd/skywire-visor/commands/root.go
```diff
@@ -146,24 +170,24 @@ var rootCmd = &cobra.Command{
 			os.Exit(1)
 		}
 		//log for initial checks
-		log := initLogger(tag, syslogAddr)
-		_, hook := logstore.MakeStore(runtimeLogMaxEntries)
-		log.AddHook(hook)
+		mLog := initLogger(tag, syslogAddr)
+		log := mLog.PackageLogger("pre-run")
+
+		//_, hook := logstore.MakeStore(runtimeLogMaxEntries)
```
jdknives: Dead code?

internal/gui/gui.go
```diff
@@ -64,24 +64,36 @@ var (
 )
 
 // GetOnGUIReady creates func to run on GUI startup.
-func GetOnGUIReady(icon []byte, conf *visorconfig.V1) func() {
+func GetOnGUIReady(icon []byte, conf *visorconfig.V1) (ret func()) {
```
jdknives: I dont see why we are having a named return. Makes code more complicated and doesnt improve readability imo.

internal/gui/gui.go
```diff
@@ -417,6 +438,24 @@ func handleUserInteraction(conf *visorconfig.V1, doneCh chan<- bool) {
 	}
 }
 
+func handleRootInteraction(conf *visorconfig.V1, doneCh chan<- bool) {
+	for {
+		select {
+		//		case <-mOpenHypervisor.ClickedCh:
```
jdknives: Dead code?

jdknives (changes requested):

cmd/skywire-cli/commands/config/gen.go
0pcom: i assume you mean the next line
```
logger.Warn("writing config as root to directory not owned by root")
```

it will only fail with insufficient permissions. At some point It was suggested to change this to a warning, it's implemented identically as it is in the visor except for a different logging message

cmd/skywire-cli/commands/config/gen.go
0pcom: I have not tested on those, i think mac will use the same name for root user (root) but it may not have /root as a directory

I will change this to only happen on linux

cmd/skywire-visor/commands/root.go
0pcom: commented out while testing and then forgotten.

removed with next commit

internal/gui/gui.go
0pcom: changed in [4de574f](https://github.com/skycoin/skywire/pull/1155/commits/4de574fc6930e038a1b9759635667b512b1596f1)

internal/gui/gui.go
0pcom: changed in [4de574f](https://github.com/skycoin/skywire/pull/1155/commits/4de574fc6930e038a1b9759635667b512b1596f1)

cmd/skywire-cli/commands/config/gen.go
```diff
@@ -221,6 +285,46 @@ var genConfigCmd = &cobra.Command{
 				}
 			}
 		}
+		skywire := os.Args[0]
+		match := strings.Contains("/tmp/", skywire)
+		if (!stdout) || (!match) {
```
mrpalide, line 251: Could we exclude windows here? means add `selecteOS != "windows"` to conditions? Because we need `.exe` at file name there, otherwise cannot run (specific for vpn-client that use there currently).

internal/gui/gui_unix.go
```diff
@@ -3,8 +3,22 @@
...
+	"os/user"
+
+	"github.com/skycoin/skywire/pkg/util/osutil"
+)
 
 func platformExecUninstall() error {
 	return osutil.Run("/bin/bash", "-c", deinstallerPath)
 }
+
+func isRoot() bool {
```
mrpalide, line 16: Could you add same function in `gui_windows.go` file? Something like:
```
func isRoot() bool {
    return true
}
```

mrpalide (approved):
Functionality and codes approved, just a little change there is need to handle.

cmd/skywire-cli/commands/config/gen.go
0pcom: I fixed the behavior for windows. Have not directly tested but implemented as follows:

```
//binaries have .exe extension on windows
var exe string
if skyenv.OS == "windows" {
exe = ".exe"
}
// Disable apps not found at bin_path with above exceptions for go run and stdout
if _, err := os.Stat(conf.Launcher.BinPath + "/" + "skychat" + exe); err != nil {
```

so the string variable exe will just be an empty string unless the OS is windows

mrpalide: I think is sufficient and OK. I'll test it when switching to windows after my current Erson redis layout testing,

internal/gui/gui.go
mrpalide: This make problem in windows installer actually. We run visor with administrator privilege because of vpn-client, and need these buttons (open hypervisor UI / open VPN UI) there.

internal/gui/gui.go
```diff
@@ -64,13 +64,13 @@ var (
...
 // GetOnGUIReady creates func to run on GUI startup.
-func GetOnGUIReady(icon []byte, conf *visorconfig.V1) (ret func()) {
+func GetOnGUIReady(icon []byte, conf *visorconfig.V1) func() {
 	doneCh := make(chan bool, 1)
 	logger := logging.NewMasterLogger()
 	logger.SetLevel(logrus.InfoLevel)
 
 	httpC := getHTTPClient(conf, context.Background(), logger)
-
+	var ret func()
```
jdknives: No need to declare the variable here. We can just use `return func(){...}` directly inside the function, right?

jdknives (changes requested):

internal/gui/gui.go
0pcom: rectified in [99e3dcc](https://github.com/skycoin/skywire/pull/1155/commits/99e3dcc4982b68229c90378ff5e55c10f8793828)

mrpalide (approved):

PR #1157, opened 2022-04-19
Fix version check

### LINKED ISSUE #1154: Visor should not fail on path or minor version mismatch
**Describe the bug**

Currently the visor fails if the config version is behind the visor version even if the mismatch is only a patch or minor version. This should be dropped in favor of only checking for major version mismatch. Patch and minor changes are not supposed to introduce any breaking changes so it is unnecessary and annoying during development to fail on mismatch.

0pcom:
Fixes #1154


mrpalide:
It's OK for different patch version, but on minor different get error yet.
image

cmd/skywire-visor/commands/root.go
```diff
@@ -173,22 +170,20 @@ var rootCmd = &cobra.Command{
 			os.Exit(1)
 		}
 		//log for initial checks
-		log := initLogger(tag, syslogAddr)
-		_, hook := logstore.MakeStore(runtimeLogMaxEntries)
-		log.AddHook(hook)
+		mLog := initLogger(tag, syslogAddr)
+		log := mLog.PackageLogger("pre-run")
+
+		//_, hook := logstore.MakeStore(runtimeLogMaxEntries)
```
jdknives: Please remove commented code if you want to remove it.

cmd/skywire-visor/commands/root.go
```diff
@@ -299,7 +215,8 @@ var rootCmd = &cobra.Command{
 
 func runVisor(conf *visorconfig.V1) {
 	var ok bool
-	log := initLogger(tag, syslogAddr)
+	mLog := initLogger(tag, syslogAddr)
```
jdknives: Why are we initializing this and then assinging it to `log`?

cmd/skywire-visor/commands/root.go
```diff
@@ -581,3 +446,144 @@ func checkHvIsRunning(addr string, retries int) bool {
 	}
```
jdknives, line 438: `isHvRunning` should be sufficient as a name. Also a HEAD request here should be enough since we are not going to use the resp.Body it seems. Lastly, is there a elegant way to just wait for hypervisor being started up before executing this?

cmd/skywire-visor/commands/root.go
```diff
@@ -581,3 +446,144 @@ func checkHvIsRunning(addr string, retries int) bool {
...
 	return false
 }
+
+func updateConfig(mLog *logging.MasterLogger, conf *visorconfig.V1) {
```
jdknives: I am wondering if it would be better to ask the user to run the update CLI command. For most users we will perform config upgrades via package scripts. For others running from source, they could/should be able to run CLI I would think. Would save us the trouble from reimplementing the CLI code here.

cmd/skywire-visor/commands/root.go
```diff
@@ -581,3 +446,144 @@ func checkHvIsRunning(addr string, retries int) bool {
...
+	skywire = os.Args[0]
+	//indicates where skywire was started
+	path, err := os.Getwd()
+	if err != nil {
+		log.WithError(err).Fatal()
+	}
+	workDir = path
+	//retrieve build info
+	visorBuildInfo = buildinfo.Get()
+	if visorBuildInfo.Version == "unknown" {
```
jdknives: About the following blocks of code: 

What is the value of checking for all these detailed build infos? If we really do need them, should it be moved to the buildinfo pkg?

pkg/visor/visorconfig/parse.go
```diff
@@ -28,7 +28,16 @@ func Parse(log *logging.Logger, r io.Reader, confPath string, visorBuildInfo *bu
 
 	// we check if the version of the visor and config are the same
 	if (conf.Version != "unknown") && (visorBuildInfo.Version != "unknown") {
-		compat = strings.Contains(strings.Split(visorBuildInfo.Version, "-")[0], strings.Split(conf.Version, "-")[0])
+		cver := strings.Split(visorBuildInfo.Version, "-")[0] //v0.6.0
```
jdknives: Can we use this package to compare versions? Only need to compare major version I think https://pkg.go.dev/golang.org/x/mod/semver

0pcom:
> It's OK for different patch version, but on minor different get error yet. image

is this not the desired behavior?

The config needs to be updated if there is such a drastic change to the version, I thought?

@jdknives didn't object so I guess my assumption is wrong

I will implement the semver package as you suggest

cmd/skywire-visor/commands/root.go
0pcom: I personally never use config update.

Config gen is very solidly implemented, whereas it isn't clear that config update will act as desired and was otherwise untested for updating possibly between versions. Its intended use was only to change the values that cannot be directly set with config gen.

cmd/skywire-visor/commands/root.go
0pcom: this is not a function I have touched, I'm guessing @mrpalide made that one. I've changed the name and will see what i can do on the rest

cmd/skywire-visor/commands/root.go
0pcom: apparently a mistake, I've changed it without issue

cmd/skywire-visor/commands/root.go
0pcom: done

pkg/visor/visorconfig/parse.go
```diff
@@ -24,11 +25,37 @@ func Parse(log *logging.Logger, r io.Reader, confPath string, visorBuildInfo *bu
...
 	}
-	log.WithField("config version: ", conf.Version).Info()
-
+	if conf.Version != "unknown" {
+		log.WithField("config version: ", conf.Version).Info()
+	}
 	// we check if the version of the visor and config are the same
 	if (conf.Version != "unknown") && (visorBuildInfo.Version != "unknown") {
-		compat = strings.Contains(strings.Split(visorBuildInfo.Version, "-")[0], strings.Split(conf.Version, "-")[0])
+		v1, err := semver.Make(strings.TrimPrefix(conf.Version, "v"))
```
ersonp: We can name this `cVer` and the one below `vVer`.

pkg/visor/visorconfig/parse.go
```diff
@@ -24,11 +25,37 @@ func Parse(log *logging.Logger, r io.Reader, confPath string, visorBuildInfo *bu
...
+			return conf, compat, err
+		}
+		if v1.Major == v2.Major {
+			compat = true
+		}
+	} else {
+		compat = true
 	}
+	/*
+			cver := strings.Split(visorBuildInfo.Version, "-")[0] //v0.6.0
```
ersonp: Seems like this was left in by mistake.

cmd/skywire-visor/commands/root.go
mrpalide: actually is older than me on skywire team :smile:, bdw isHvRunning is sufficient. :+1:

mrpalide (commented):
Based on its issue #1154, no need to check minor update, and if I missing something and need to check then PR is OK.

pkg/visor/visorconfig/parse.go
0pcom: removed

pkg/visor/visorconfig/parse.go
0pcom: changed

jdknives:
@0pcom correct. Only need to check on major increments. There are a few remaining issues: 

1. 

```
[2022-04-26T12:09:19+02:00] INFO [buildinfo]: version="v1.0.0"
[2022-04-26T12:09:19+02:00] INFO [buildinfo]: built on="2022-04-26T10:09:09Z"
[2022-04-26T12:09:19+02:00] INFO [buildinfo]: against commit="c33e3c8fdf0db9443c863ba86b3beafb8da63bc8"
[2022-04-26T12:09:19+02:00] INFO [visor:config]: Reading config from file.
[2022-04-26T12:09:19+02:00] INFO [visor:config]: filepath="skywire-config.json"
[2022-04-26T12:09:19+02:00] INFO [visor:config]: config version: ="v0.6.0-241-g20aa58f7"
[2022-04-26T12:09:19+02:00] ERROR [visor:update-config]: config version does not match visor version
[2022-04-26T12:09:19+02:00] ERROR [visor:update-config]: skywire version: ="v1.0.0"
[2022-04-26T12:09:19+02:00] INFO [visor:update-config]: match:true
[2022-04-26T12:09:19+02:00] INFO [visor:update-config]: updstr:go run cmd/skywire-cli/skywire-cli.go config gen -b
[2022-04-26T12:09:19+02:00] INFO [visor:update-config]: please update your config with the following command:
```

Logging is too verbose. Most of the info logging should be at debug level. Some of the logs are duplicate such as the last three ones.

cmd/skywire-visor/commands/root.go
```diff
@@ -581,3 +440,70 @@ func checkHvIsRunning(addr string, retries int) bool {
...
+	updstr = "\n		" + updstr + "\n"
+	log.Info("please update your config with the following command:\n", updstr)
+	log.Fatal("failed to start skywire")
+
+}
+
+func logBuildInfo(mLog *logging.MasterLogger) {
+	log := mLog.PackageLogger("buildinfo")
+	visorBuildInfo = buildinfo.Get()
+	if visorBuildInfo.Version != "unknown" {
```
jdknives, line 445: Can we put these logs on one line?

pkg/skyenv/values_darwin.go
```diff
@@ -8,6 +8,10 @@ const (
 	OS = "mac"
 	// SkywirePath is the path to the installation folder.
 	SkywirePath = "/Library/Application Support/Skywire"
+	// Ptext is the description to use for the -p flag for config gen and for the visor
+	Ptext = "use mac installation path: " + SkywirePath
```
jdknives: I think you misunderstood my request. I would like to get rid of the additional global variable. We can just construct this message inside the flag declaration instead of exposing a global variable.

pkg/skyenv/values_darwin.go
```diff
@@ -8,6 +8,10 @@ const (
...
+	// Configjson is the config name generated by the script included with the installation on mac
+	Configjson = "skywire.json"
```
jdknives: We talked about making this congruent with the names for generic skywire configs by using `skywire-config.json`. Did that not work out or did you miss it?

jdknives (changes requested):

pkg/skyenv/values_darwin.go
0pcom: I need to implement the systray and config gen at the user level for the packages first

when that is working I will go back and figure out migration of the configs in /opt/skywire from skywire.json to skywire-config.json

It's just too much to do at the same time. I'm trying to avoid unintended consequences.

pkg/skyenv/values_darwin.go
0pcom: done with [f672c82](https://github.com/skycoin/skywire/pull/1157/commits/f672c824991c377eca6180c4d63158c7ea9dd1b1)

cmd/skywire-visor/commands/root.go
0pcom: done with [0e6b703](https://github.com/skycoin/skywire/pull/1157/commits/0e6b703b761c59028e17765c9c7d9308b775e561)

0pcom:
@jdknives i've removed the generation of the update command string per our conversation here [0e6b703](https://github.com/skycoin/skywire/pull/1157/commits/0e6b703b761c59028e17765c9c7d9308b775e561)

cmd/skywire-visor/commands/root.go
```diff
@@ -380,7 +378,7 @@ func initConfig(mLog *logging.MasterLogger, confPath string) *visorconfig.V1 { /
 		log.WithError(err).Fatal("Failed to read in config.")
 	}
 	if !compat {
-		updateConfig(mLog, conf)
+		updateConfig(mLog)
```
jdknives: This can be one line of just using log.Fatalf() for instance. Seems overkill to me to assign a new package logger and then log error and fatal separately.

jdknives (changes requested):

cmd/skywire-visor/commands/root.go
0pcom: you are right, separate func was vestigial - fixed in [cd88cfe](https://github.com/skycoin/skywire/pull/1157/commits/cd88cfecedfd287b1ca7ae221b4b02b7a359ad0c)
image

PR #1176, opened 2022-05-03
fix visor uses default config with no arguments

0pcom:
yes

visor uses skywire-config.json with no arguments by default

 Changes:	
-	visor should use the default config if present without needing to specify `-c skywire-config.json`

How to test this PR:
```
git clone --single-branch -b fix-visor-default-config https://github.com/the-skycoin-project/skywire
cd skywire
make build
./skywire-cli config gen -r
./skywire-visor

```


jdknives (changes requested):
Needs mention in changelog.

PR #1183, opened 2022-05-09
Fix Systray on Linux

### LINKED ISSUE #1177: Systray app appears but does nothing when clicked (archlinux)
I've updated the skywire-autoconfig script and git.PKGBUILD to work with the recent changes on develop.

The binaries compile successfully with the systray tag, but the systray app itself does nothing besides appear correctly.

To test on archlinux / arch-based distros:
```
yay -S skywire
```

Select cleanbuild if given the option, then ctrl+c before the build begins
```
cd ~/.cache/yay/skywire
makepkg -fip git.PKGBUILD
```

install the package by following the prompts
the skywire-autoconfig script will run as root from the postinstall process
run it again as your user (not root)
```
skywire-autoconfig
```

the config for the user space (home config) will be generated and skywire started. The systray icon should appear

Click on the systray icon. Nothing happens.

mrpalide:

Fixes #1177 

 Changes:	
- change some buttons on systray menu
- get/use last version of systray

How to test this PR:
- make build-systray
- run a visor (without sudo and root permission)
- check systray

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -359,6 +359,12 @@ mac-installer-help: ## Show installer creation help
 win-installer:
 	@powershell '.\scripts\win_installer\script.ps1'
 
+set-cap-vpn-client:
+	sudo setcap 'cap_net_admin+p' ./apps/vpn-client
+
+set-cap-vpn-server:
+	sudo setcap 'cap_net_admin+p' ./apps/vpn-server
+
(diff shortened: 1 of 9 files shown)


0pcom (approved):
looks good, tested working on linux

PR #1184, opened 2022-05-10
Fix/vpn stats

### LINKED ISSUE #946: VPN statistics are incorrect
**Describe the bug**

The statistics the VPN server returns currently are very imprecise for some reason. Both the latency as well as the up and download statistics are off by a substantial amount. We need to investigate whats the cause of the imprecision.

ersonp:
yes

Fixes #946 

 Changes:	
- Fixed bandwidth 	

How to test this PR:
1. Start a vpn-server
2. Start a vpn-client and connect to the VPN server
3. Compare the stats in the UI with any networking tool like `nload` on `wtun4` interface


ersonp:
https://user-images.githubusercontent.com/38801159/167630074-d47def90-4ac2-4bf4-9705-804f3e9eeef8.mp4

The upload and download speeds are now almost the same as shown in the tool `nload` for the tun interface. We send the data in API in bytes which I suppose is converted to `bits` for the upload and download speed in the UI. We need to specify it a bit more clearly that the speed is in terms of `bits` as compared to the total that is in terms of `bytes`.
```
[
   {
      "is_alive":true,
      "latency":196,
      "upload_speed":691,
      "download_speed":561,
      "bandwidth_sent":624761,
      "bandwidth_received":665736,
      "error":"",
      "connection_duration":89
   }
]
```
The latency is still a bit inaccurate (working on it so see if anything more can be done) and shoots up whenever there is a big download (tested this with meter.net) and the ping spikes there too while a download is going on so the behaviour seems normal.
image

The ping to the server the `vpn-server` constantly gives a ping of `240 ms`
image

mrpalide:
I test it, and can confirm that **Speed** and **Bandwidth** are OK.

https://user-images.githubusercontent.com/79150699/167779555-ef4d8de2-a353-4281-9a0a-65ba5631417c.mov

But **ping** is a few different from what showing in UI and results from teminal.

image

jdknives (approved):

PR #1208, opened 2022-05-17
Fix vpn status

### LINKED ISSUE #1205: Inconsistent states after starting the vpn client
**Describe the bug**
Using the VPN UI, if you press the connect button, the UI will show that it is starting, then will appear like it is not connecting and then connected.

**Environment information:**

- OS: Linux

**Steps to Reproduce**

1. Open the VPN UI and make sure you have a valid server selected.
2. Press the start button. The UI will show the data transmission graphs and will say that the VPN is disconnected.
3. Wait a moment. The data transmission graphs will be removed and the connect button will be show again, like when the vpn client app is totally stopped. This happens because the API returns 0 as the status of the vpn-client app.
4. Wait a moment. The data transmission graphs will be shown again and the app will appear as connecting or connected. This happens because the API returns 1 or 3 as the status of the vpn-client app.

**Actual behavior**
If you make an API call to start the vpn-client app and, just after receiving the response, you make another call to check the status of the app, the API will return 0 as the status, which means that the app is stopped. Only after an unknown delay the API will start returning another status code.

**Expected behavior**
The API should return a non 0 status code, unless the app was stopped super fast due to an error.

ersonp:
yes

Fixes #1205 

 Changes:	
- Added `App` in API interface
- Updated `AppState` and `AppStates` methods of Launcher
- Fixed `putApp` method of Hypervisor 

How to test this PR:
1. Start visor `./skywire-visor -c skywire-config.json`
2. Open postman and send a `PUT` request to `http://localhost:8000/api/visors/<visor-pk>/apps/vpn-client` with body
    ```
    {"status":1}
    ```
3. Check response

### DIFF
--- a/pkg/app/launcher/launcher.go
+++ b/pkg/app/launcher/launcher.go
@@ -174,39 +174,30 @@ func (l *Launcher) AppState(name string) (*AppState, bool) {
 		return nil, false
 	}
 	state := &AppState{AppConfig: ac, Status: AppStatusStopped}
-	if _, ok := l.procM.ErrorByName(ac.Name); ok { //nolint:errcheck
-		state.Status = AppStatusErrored
+	if err, ok := l.procM.ErrorByName(ac.Name); ok { //nolint:errcheck
+		if err != "" {
+			state.DetailedStatus = err
+			state.Status = AppStatusErrored
+		}
 	}
-	if _, ok := l.procM.ProcByName(ac.Name); ok { //nolint:errcheck
-		state.Status = AppStatusRunning
+	if proc, ok := l.procM.ProcByName(ac.Name); ok { //nolint:errcheck
+		state.DetailedStatus = proc.DetailedStatus()
+		connSummary := proc.ConnectionsSummary()
+		if connSummary != nil {
+			state.Status = AppStatusRunning
+		}
+		if state.DetailedStatus == vpn.ClientStatusConnecting {
+			state.Status = AppVPNClientConnecting
+		}
 	}
 	return state, true
 }
 
 // AppStates returns list of AppStates for all registered apps.
 func (l *Launcher) AppStates() []*AppState {
-	l.mx.Lock()
-	defer l.mx.Unlock()
-
 	var states []*AppState
 	for _, app := range l.apps {
-		state := &AppState{AppConfig: app, Status: AppStatusStopped}
-		if err, ok := l.procM.ErrorByName(app.Name); ok {
-			if err != "" {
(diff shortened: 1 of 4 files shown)


mrpalide (approved):

pkg/app/launcher/launcher.go
```diff
@@ -174,39 +174,30 @@ func (l *Launcher) AppState(name string) (*AppState, bool) {
 		return nil, false
 	}
 	state := &AppState{AppConfig: ac, Status: AppStatusStopped}
-	if _, ok := l.procM.ErrorByName(ac.Name); ok { //nolint:errcheck
-		state.Status = AppStatusErrored
+	if err, ok := l.procM.ErrorByName(ac.Name); ok { //nolint:errcheck
```
jdknives, line 177: Not implemented by you, but would it be more elegant to actually return an `error` here instead of a `string` and `bool`?

jdknives: Do we need the `//nolint:errcheck` here still?

pkg/visor/hypervisor.go
```diff
@@ -710,7 +716,18 @@ func (hv *Hypervisor) putApp() http.HandlerFunc {
 			}
 		}
 
-		httputil.WriteJSON(w, r, http.StatusOK, ctx.App)
+		if err := ctx.API.SetAppDetailedStatus(ctx.App.Name, vpn.ClientStatusConnecting); err != nil {
```
jdknives, line 719: Why are we setting the status for the app from the hypervisor? Sounds counterintuitive to me that the frontend would be able to manipulate the state of the application thats running directly. It should be able to call certain actions and then based on whether or not they succeed, the app changes its state, which the frontend then can poll.

pkg/app/launcher/launcher.go
ersonp: `//nolint:errcheck` can be taken out. `l.procM.ErrorByName(ac.Name)` was actually created by me, and I think there was some trouble sending the error through RPC, I will check it once again if it's possible.

pkg/visor/hypervisor.go
ersonp: There is a small delay when we send back the response of the API `putApp` and the app starting and changing the the `detailed_status` by itself. That used to cause the vpn-ui to show disconnecting first and then connecting after we started it.
Another way would be to send a `starting` detailed status regardless of the apps state and it will update to the correct status after a while. But that makes the other apps appear slow so I went with this approach.

PR #1241, opened 2022-06-01
Fix updater [WIP]

0pcom:
yes

The initial work adds a much abbreviated update process in place of the existing one, without changing much.

Needed / Proposed Changes:

- Remove version checks completely
- prompt if repo configuration is needed
- otherwise, update without prompting or checking the version, via `apt`

#### Version Checks

`apt update && apt install skywire-bin` will update skywire to the latest version or install it if it is not installed. Packages are not re-installed, by default, with `apt install` (correct me if I am wrong).

The version checking is entirely unnecessary, all of it. Yet they occupy a substantial part of the code that I was not able to remove it without breaking things.

#### Prompt for repository configuration

If the repository used for updating the .deb packages is not already configured in the software sources, it is highly abnormal to preform this configuration without the explicit consent of the user.

Secondly, the way this configuration was previously done was no longer congruent with the skybian images (on my fork / in my open PR), and actually conflicting with that (wouldn't detect the existing config and then would redundantly set it in a different file).

This would be the open question - the repository configuration in my skybian fork is provided by the skybian package, which is architecture-specific to arm architectures. This repository configuration could be made a seprate .deb package (and could include the repository signing key) and would then be able to work on amd64. There is no real reason not to do this, from my perspective, and the advantages for testing should be obvious.

So there needs to be a prompt to preform this configuration ; a .deb package would be downloaded from the archive of the apt repo server, and then upon it's installation, skywire-bin would be available to install via `apt`

#### otherwise update without prompting

Clicking the update button should either not prompt the user for confirmation if the repository is set, or it should  only give a generic "are you sure you want to install / update?" confirmation dialog.

The visor should restart at the end of the update

#### Testing

requires:

* visor is run as root
* .deb based distro

```
git clone https://github.com/the-skycoin-project/skywire
git checkout fix-updater
make run-source-sudo
```

* open [http://127.0.0.1:8000](http://127.0.0.1:8000)
* go to the visor page
* click update from the drop-down menu
* ui prompt for adding the repository configuration (installing the package described earlier)
* click yes, software sources are configured as they would be in skybian
* skywire is installed

So in theory, when this can take place, the update button _will also function as the installer_ or a wrapper for the process which installs it, and _that should be the easiest thing to test_

#### Still needed

In order for the above outlines process to be robust in terms of using the key that the visor was already running with, when it is being installed by the updater button the key should be passed via environmental variable to skkywire-autoconfig, which will further pass it to the config gen command.

Additionally, the autoconfig script assumes skywire was already running as a systemd service, so there should be some thought given to the transition between the visor running from the source code to the visor running from the installation.

### DIFF
--- a/pkg/util/updater/updater.go
+++ b/pkg/util/updater/updater.go
@@ -6,17 +6,17 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"sync/atomic"
 	"os/user"
-	"github.com/skycoin/skywire-utilities/pkg/buildinfo"
-	"github.com/google/go-github/github"
+	"sync/atomic"
+
 	"github.com/bitfield/script"
+	"github.com/google/go-github/github"
 	"github.com/skycoin/skycoin/src/util/logging"
 	"github.com/zcalusic/sysinfo"
 
+	"github.com/skycoin/skywire-utilities/pkg/buildinfo"
 	"github.com/skycoin/skywire/pkg/restart"
 )
-
 
 const (
 	owner          = "skycoin"
@@ -111,7 +111,7 @@ func (u *Updater) Update(updateConfig UpdateConfig) (updated bool, err error) {
 		return false, errors.New("operating system not supported")
 	}
 
-	if (si.OS.Vendor == "debian") {
+	if si.OS.Vendor == "debian" {
 		u.status.Set("Checking for available package")
 		available, err := script.Exec(`apt-cache search skywire-bin`).String()
 		if err != nil {
@@ -140,7 +140,7 @@ func (u *Updater) Update(updateConfig UpdateConfig) (updated bool, err error) {
 		//this may not return before the process is restarted
 		return true, nil
 	}
-	if (si.OS.Vendor == "arch") {
+	if si.OS.Vendor == "arch" {
 		u.status.Set("not yet implemented")
 		u.log.Error("not yet implemented")
 		return false, errors.New("not yet implemented")
@@ -174,28 +174,6 @@ func (u *Updater) UpdateAvailable(channel Channel) (*Version, error) {
 
 	return latestVersion, nil
 }
-
-func (u *Updater) getVersion(updateConfig UpdateConfig) (string, error) {
-	version := updateConfig.Version
-	if version == "" {
-		latestVersion, err := u.UpdateAvailable(updateConfig.Channel)
-		if err != nil {
-			return "", fmt.Errorf("failed to get last Skywire version: %w", err)
-		}
-
-		// No update is available.
-		if latestVersion == nil {
-			return "", nil
-		}
-
-		version = latestVersion.String()
-	}
-
-	u.log.Infof("Update found, version: %q", version)
-
-	return version, nil
-}
-
 
 func needUpdate(last *Version) bool {
 	current, err := currentVersion()
--- a/pkg/util/updater/updater.go
+++ b/pkg/util/updater/updater.go
@@ -6,15 +6,17 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"os/exec"
 	"sync/atomic"
-
-	"github.com/google/go-github/github"
-	"github.com/skycoin/skycoin/src/util/logging"
-
+	"os/user"
 	"github.com/skycoin/skywire-utilities/pkg/buildinfo"
+	"github.com/google/go-github/github"
+	"github.com/bitfield/script"
+	"github.com/skycoin/skycoin/src/util/logging"
+	"github.com/zcalusic/sysinfo"
+
 	"github.com/skycoin/skywire/pkg/restart"
 )
+
 
 const (
 	owner          = "skycoin"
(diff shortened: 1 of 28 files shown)


pkg/util/updater/updater.go
```diff
@@ -86,60 +88,64 @@ const (
 // NOTE: Update may call os.Exit.
 func (u *Updater) Update(updateConfig UpdateConfig) (updated bool, err error) {
 	if !atomic.CompareAndSwapInt32(&u.updating, 0, 1) {
-		return false, ErrAlreadyStarted
+		return false, errors.New("updating already started")
```
jdknives, line 91: Why are we replacing this with an inline error?

pkg/util/updater/updater.go
```diff
@@ -86,60 +88,64 @@ const (
...
-	version, err := u.getVersion(updateConfig)
+	current, err := user.Current()
 	if err != nil {
-		return false, err
-	}
-
-	// No update is available.
-	if version == "" {
-		return false, nil
+		u.status.Set("failed checking permissions")
```
jdknives, line 97: Is this status we set logged later down the line by the calling function?

pkg/util/updater/updater.go
```diff
@@ -86,60 +88,64 @@ const (
...
 		}
-		u.log.Info("Uninstalling old version compeleted.")
+		//this may not return before the process is restarted
+		return true, nil
 	}
-
-	u.status.Set(fmt.Sprintf("Installing Skywire %q", version))
-	if err := u.aptInstall(); err != nil {
-		return false, err
+	if si.OS.Vendor == "arch" {
```
jdknives, line 143: We are not going to implicitly support arch here in the binaries. Please take this out.

pkg/util/updater/updater.go
```diff
@@ -86,60 +88,64 @@ const (
...
-		return false, err
-	}
-	if currentVersion.Minor == 4 && currentVersion.Patch <= 2 {
-		u.status.Set("Uninstall current skywire version") // if needed
-		if err := u.aptRemove(); err != nil {
+	if si.OS.Vendor == "debian" {
+		u.status.Set("Checking for available package")
+		available, err := script.Exec(`apt-cache search skywire-bin`).String()
+		if err != nil {
+			return false, err
```
jdknives, line 118: I would maybe use `fmt.Errorf` here so we can add some additional context to the error. Same goes for below errors that stem from executing `apt` commands.

pkg/util/updater/updater.go
```diff
@@ -86,60 +88,64 @@ const (
...
+		}
+		u.log.Info("synced package database")
+		u.log.Info("installing skywire-bin")
+		if _, err := script.Exec(`sudo NOAUTOCONFIG=true apt install skywire-bin -y`).String(); err != nil {
+			u.log.Error("error installing skywire-bin")
+			return false, err
+		}
+		u.log.Info("installed skywire-bin")
+		u.log.Info("updating config and restarting visor")
+		if _, err := script.Exec(`DMSGPTYTERM=true skywire-autoconfig`).String(); err != nil {
```
jdknives, line 136: Will this run if we run apt-update before successfully?

jdknives (approved):

jdknives:
An additional comment from my side. Instead of checking for the architecture, I would like to just check for the presence of the Skybian build tag. If it is set, we update. Otherwise we do not. Anything that is not a Skybian system can and should be handled differently, namely through the user running `apt` or `pacman` commands for Linux and just getting the latest installer for MacOS and Windows. This should be cut back to be an updating convenience feature specifically for the boards we support and know of. In all other scenarios, the updating should not be handled by the visor, since the much better solution is to let the user handle updating manually.

pkg/util/updater/updater.go
0pcom: This i will need to implement for testing. None of the code i added for updating .deb was tested yet because i am writing it on arch.

I can remove it before this gets merged or we can discuss more of an appropriate context for this, but if i am able to implement this, it would let me much more easily test the updater on the system it is being developed on.

all of the .deb packages, everything i have done, was first done on arch and tested there and then re-implemented for debian.

it may not be able to be implemented, because of how permissions need to be handled for invoking `yay` on archlinux, but i need to try.

pkg/util/updater/updater.go
0pcom: I had taken out most of the code in that file and several others, until the point it wouldnt work, so i copied the one function i modified and then git reset.

most of the errors already declared will be taken out, based on my proposal.

i will change it to non-inline errors before all is said and done.

pkg/util/updater/updater.go
0pcom: sure

pkg/util/updater/updater.go
0pcom: it should

0pcom:
> An additional comment from my side. Instead of checking for the architecture, I would like to just check for the presence of the Skybian build tag. If it is set, we update. Otherwise we do not. Anything that is not a Skybian system can and should be handled differently, namely through the user running `apt` or `pacman` commands for Linux and just getting the latest installer for MacOS and Windows. This should be cut back to be an updating convenience feature specifically for the boards we support and know of. In all other scenarios, the updating should not be handled by the visor, since the much better solution is to let the user handle updating manually.

there is no architecture check that i know of. I am checking the linux distro, which should be at least additionally checked along with the build tag.

In the second place, the same binaries are in the same packages used by the images in the same repo used by anyone who installs skywire-bin. The arm binaries could have the skybian build tag, but this wont differentiate a skybian installation from the same package installed outside the skybian image.

its possible to switch on the skybian env instead of the build tag.

in the second place, if the updater is right, it should be available as part of every package equally (no reason not to) or otherwise there needs to be a carve-out that it will work on amd64 for testing.

if its not equally implemented and available for updating all the architecture .deb packages which are available, then we need to take out the updater completely and move it to skybian.

it should be separate anyways, i think, but its about to work correctly with a few more changes.

no objections to removing all the version checks or otherwise the outline of the update procedure?

0pcom:
closing this PR in favor of #1244 after discussion with @jdknives

PR #1267, opened 2022-06-22
fix systray xfce

mrpalide:

Fixes #	

 Changes:	
- add a sed command to change dbus vendor `conn.go` file, both for `Makefile` (build locally) and `.goreleaser-linux.yml`

How to test this PR:
- `make build-systray` and run on xfce desktop
- also can download my built from [here](https://freeshell.de/~mrpalide/skywire.tar.xz), extract it and test

Screencast:

https://user-images.githubusercontent.com/79150699/175022020-3db0604e-5a73-4bd2-b91e-ab4b706e5a65.mov

### DIFF
--- a/.goreleaser-linux.yml
+++ b/.goreleaser-linux.yml
@@ -14,6 +14,7 @@ release:
 before:
   hooks:
     - go mod tidy
+    - sed -i '/go conn.handleCall(msg)/c\conn.handleCall(msg)' ./vendor/github.com/godbus/dbus/v5/conn.go
 builds:
 
   - id: skywire-visor-amd64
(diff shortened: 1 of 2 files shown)


0pcom:
this is using the binaries you compiled / the provided archive:

image

still a very inconsistent behavior.

However, I think there is a chance this is not to do with the systray, but instead with a delay caused by the rpc server.

```
[2022-06-25T14:30:04-05:00] WARN [systray:rpc_client]: RPC connection failed. Try again in 2 seconds.
[2022-06-25T14:30:08-05:00] INFO [systray:rpc_client]: RPC Connection established
```

if i click on the systray app very soon after starting the visor, it behaves strangely.
It takes time to initialize after clicking on it, and does not initialize the buttons / menu items when the visor has started.

I have the same issue with my rework of the systray.

I think we need a baseline measurement here, to eliminate the rpc connection delay as a possible cause of such failures.

PR #1271, opened 2022-06-27
New update procedure for the UI

Senyoret1:

 Changes:	
- Now, the option for updating all the visors opens this modal window:
p
Visors with Mac and Windows tags are added to the non-updatable list.

- When trying to update a single node, a confirmation modal window is shown, for opening the terminal.

How to test this PR:
Use the update options in the visor list and in the visor details page.


Senyoret1:
- Now the update option opens the terminal with the ?commands=update argument, to make the terminal start the update procedure automatically.
- The texts were updated, to indicate that the update procedure will start automatically.
- Other small fixes.

Note: the ?commands=update argument will only work after https://github.com/skycoin/dmsg/pull/178 is implemented.

PR #1305, opened 2022-07-12
hypervisor UI using go templating [WIP]

0pcom:
a demonstration of a basic hypervisor UI using go templating, and hugo as the templating engine

Changes of note to the go code are in pkg/visor/hypervisor.go

the hugo sources are at static/hypervisor-template/

__A hugo template was chosen randomly for the basic demontration - this is not intended for production use__

Testing:
```
git checkout hv-go-template
make run-source
```

or otherwise start a visor from the `hv-go-template` branch

navigate to the [hypervisor UI](127.0.0.1:8000)

The public key of the hypervisor is rendered twice.

Once as a variable which has been passed into the page, and once from a template function. The template function is called from the template when it is rendered, and returns the key.

For our purposes it may be desirable to pass a limited set of information into the template as a struct, and possibly using those variables as arguments to certain template functions calls


0pcom:
closing for now ; will revise and present this again in the future

PR #1309, opened 2022-07-13
auto-peering visors to the hypervisor (skybian)

### LINKED ISSUE #1312: autopeering visors to a hypervisor
The autopeering of visors to a hypervisor is accomplished in skybian once on the first boot via an http endpoint which provides the public key of the hypervisor.

The system I propose will integrate into the package scripts and systemd services to replace [this](https://github.com/skycoin/skybian/blob/master/script/skymanager.sh#L33-L41)...

```
#query remote node for pk
_pubkey=$(curl ${_ip}:7998)
#rough errorcheck
if [[ ("${_pubkey}" == *"FATAL"*) || ("${_pubkey}" == *"Failed"*) ]] ; then
_pubkey="0"
fi
fi

#configure skywire
skywire-autoconfig ${_pubkey}
```

...in a much better way inside the skywire source code using somewhat arbitrary functionality which I will describe.

From the perspective of the packager, it's highly advantageous for certain things like getting a remote hypervisor public key via some script logic to be moved into the source.

It is possible to override defaults of a systemd unit by creating a drop-in file (i.e. /etc/systemd/system/skywire.service.d/skywire.conf) 

This drop-in file could be provided by the skybian package to use different flags to start the visor:
```
ExecStart=/usr/bin/skywire-visor -pl 'skywire-cli hv pk`
```

In the above example, the `skywire-visor -l --autopeer` flag is a slight modification of the `-j` flag which explicitly sets a remote hypervisor at runtime; the -l flag should allow an arbitrary command to be specified which is assumed to return a public key or list of public keys. This gets around the fact that _subshells will not work in the systemd file_, in any case we don't want to evaluate that command only immediately but also when the visor cannot connect to the hypervisor.

The visor should only try intermittently to re-peer. Errors returned by the command should be printed in the visor's logging, preferably. This should not need to interrupt the running visor process.

Allowing for an arbitrary command to be specified grants ultimate flexibility to this system and allows any scheme to be implemented.

The second command, `skywire-cli hv pk` should by default (but changeable with flags) query the http endpoint provided by the srvpk.service.. at the xxx.xxx.xxx.2 ip address of the current subnet with default port 7998
It should return the public key or the error from the other skywire-cli command.

It was suggested by @jdknives to keep the srvpk.service how it is.

The major change here, what I am suggesting actually takes the skywire-autoconfig script out of the picture and also the hypervisor key is not actually set in the config.

It is possible to set it in the config.

My suggestion for that is a `skywire-cli config gen` flag which is similarly able to execute an arbitrary command which is assumed to return a public key or list of public keys. The output of config gen could then be filtered (preferably internally with another flag) to only print out the hypervisor public key.

While it may be worthwhile to implement this for / with `config gen` I think it's a little bit overkill to think about trying to update the config with the hypervisor. We definitely should not do this in certain instances.

In the instance that someone has set their desktop as the remote hypervisor for their visors, there must be some way to disable this from causing errant peering attempts. The user would need to set some environmental variable on all the boards is the easiest way I can think which won't need to be repeated, still a sub-optimal solution in that regard

Assuming that the automation here does all the hard work of the initial peering, is it too much to ask that the user do some additional config after the fact?

My only other solution would be to not set the hypervisor key at all in the config by default and the user must explicitly set their public key (via skywire-autoconfig) in the visors after they appear in the hypervisor UI with the following command exactly:

```
skywire-autoconfig $(skywire-cli visor hvpk)
```

so the `skywire-visor -l` flag should be ignored if remote hypervisors are set in the config file.

0pcom:
attempt to make visors re-peer to a new hypervisor when a new hypervisor is present at a predetermined static ip

Fixes #1312

this PR ports the existing auto-peering functionality from skybian to the skywire source code / binaries and attempts to extend this system to re-peer visors to the hypervisor when a connection is lost

testing requires a hypervisor running the following:
```
skywire-cli visor pk -w
```
The srvpk endpoint may be used instead of the above command

run the visor from source as follows:
```
go run cmd/skywire-cli/skywire-cli.go config gen -brn | sudo go run cmd/skywire-visor/skywire-visor.go -l "go run cmd/skywire-cli/skywire-cli.go hv pk -i 192.168.0.100" -mn
```

replace the ip address in the above command example with that of the running machine.


0pcom:
to test without using the default http endpoint for the public key (or any)
one can create a file that contains the public key of a running hypervisor for which re-peering will be attempted

```
echo -n "024034dd09cbff03787db740124496ce8fdbe0cac2a51e20366cb59be10b665506" | tee skywiretest.txt
```

then invoke the visor as follows:

```
 sudo go run cmd/skywire-visor/skywire-visor.go -pml "cat '/home/user/skywiretest.txt'"
```

as you run this visor in the foreground, turn the (remote) hypervisor off and on to observe that the command string provided as an argument to the -l flag is executed on redial attempts to the hypervisor.

To further test, one can edit the file to contain a different public key while this is happening, save the file, and it should then attempt to redial with the new public key.

this needs to be made versatile enough to handle a list of remote hypervisor public keys. Ideally, the http endpoint would output the keys in the same format that they are read, to require minimal parsing.

cmd/skywire-cli/commands/config/gen.go
```diff
@@ -222,6 +226,12 @@ var genConfigCmd = &cobra.Command{
 		if testEnv {
 			serviceConfURL = testconf
 		}
+		if autopeer {
+			ip, err := script.Exec(`ip route show | grep -i 'default via'| awk '{print $3 }'`).String()
```
jdknives: Is this even required? The hv for Skybian will always be at `192.0.0.2`, right?

cmd/skywire-cli/commands/hv/hvpk.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"fmt"
+
+	"github.com/bitfield/script"
+	"github.com/spf13/cobra"
+)
+
+func init() {
+	ipaddr, err := script.Exec(`bash -c "_gateway=$(ip route show | grep -i 'default via'| awk '{print $3 }'); _ip=${_gateway%.*}.2; printf ${_ip}"`).String()
```
jdknives: Either camelcase or just use `ip`.

cmd/skywire-cli/commands/hv/hvpk.go
```diff
@@ -0,0 +1,31 @@
...
+	if err != nil {
+		err.Error()
```
jdknives: What is this doing?

cmd/skywire-cli/commands/hv/hvpk.go
jdknives: Also, is there a need to actually get the IP? Skybian has preconfigured Ips.

cmd/skywire-cli/commands/hv/hvpk.go
```diff
@@ -0,0 +1,31 @@
...
+	RootCmd.AddCommand(pkCmd)
+	pkCmd.Flags().IntVarP(&port, "port", "p", 7998, "port to query")
+	pkCmd.Flags().StringVarP(&ipadd, "ip", "i", ipaddr, "ip address to query")
+}
+
+var pkCmd = &cobra.Command{
+	Use:   "pk",
+	Short: "Fetch Hypervisor Public Key",
+	Run: func(_ *cobra.Command, _ []string) {
+		hvpk, err := script.Exec(`bash -c "curl ` + fmt.Sprintf("%s:%d", ipadd, port) + `"`).String()
```
jdknives: I know it may be more convenient for you to just use the script libary in all places, but I would prefer to use Go whereever possible without unreasonable drawbacks.

cmd/skywire-cli/commands/hv/hvpk.go
```diff
@@ -0,0 +1,31 @@
...
+		if err != nil {
+			err.Error()
```
jdknives: Same issue as above.

cmd/skywire-cli/commands/hv/root.go
```diff
@@ -17,6 +17,8 @@ var (
 	pk      string
 	url     string
 	pkg     bool
+	port    int
+	ipadd   string
```
jdknives: Please try to use somewhat canonical names at least in your own part of the codebase. So something like `ip` or `ipAddr`.

cmd/skywire-cli/commands/hv/hvpk.go
```diff
@@ -0,0 +1,31 @@
...
+		}
+		fmt.Printf("%s", hvpk)
```
jdknives: A `Println` may be better here? What is the formatting doing? Its also missing a newline I would assume.

cmd/skywire-cli/commands/visor/info.go
```diff
@@ -14,16 +14,18 @@ import (
 var path string
 var pkg bool
 var web bool
+var webport string
```
jdknives: Camel case.

jdknives: Also, what do you think about just serving an endpoint from the hv UI returning only its public key instead of running this via CLI?

cmd/skywire-visor/commands/root.go
```diff
@@ -55,6 +55,8 @@ var (
 	hypervisorUI         bool
 	remoteHypervisorPKs  string
 	disableHypervisorPKs bool
+	autopeer             bool
```
jdknives: Camel case.

cmd/skywire-visor/commands/root.go
```diff
@@ -262,6 +268,33 @@ func runVisor(conf *visorconfig.V1) {
 			conf.Hypervisors = append(conf.Hypervisors, pubkey)
 		}
 	}
+	//autopeering should only happen when there is no local or remote hypervisor set in the config.
+	if conf.Hypervisors != nil {
```
jdknives, line 293: Can be one line

jdknives: Also what is this check even doing? Its false by default and if the user sets it to true, we will overwrite what we do here anyways?

cmd/skywire-visor/commands/root.go
```diff
@@ -262,6 +268,33 @@ func runVisor(conf *visorconfig.V1) {
...
+		if conf.Hypervisor != nil {
+			autopeer = false
+		}
+	}
+	if autopeer {
+		log.Infof("autopeering...")
+		var hvkey string
```
jdknives: No need to declare this explicitly here.

cmd/skywire-visor/commands/root.go
```diff
@@ -262,6 +268,33 @@ func runVisor(conf *visorconfig.V1) {
...
+		hvkey, err := script.Exec(peercmd).String() //TODO: mrpalide this needs to run again on loss of connection to the hypervisor within the same constraints
+		if err != nil {
+			log.Error("error autopeering")
```
jdknives: Please use a full sentence here or follow the same pattern you use elsewhere.

cmd/skywire-visor/commands/root.go
```diff
@@ -262,6 +268,33 @@ func runVisor(conf *visorconfig.V1) {
...
+		} else {
+			hvkey = strings.TrimSuffix(hvkey, "\n")
+			hypervisorPKsSlice := strings.Split(hvkey, ",")
```
jdknives, line 304: I think string.Trimspace would be sufficient here.

cmd/skywire-visor/commands/root.go
```diff
@@ -262,6 +268,33 @@ func runVisor(conf *visorconfig.V1) {
...
+			for _, pubkeyString := range hypervisorPKsSlice {
+				if err := pubkey.Set(pubkeyString); err != nil {
+					log.Warnf("Cannot add %s PK as remote hypervisor PK due to: %s", pubkeyString, err)
+					continue
+				}
+				log.Infof("%s PK added as remote hypervisor PK", pubkeyString)
+				conf.Hypervisors = append(conf.Hypervisors, pubkey)
+				skyenv.AutoPeer = true
```
jdknives: What is this doing?

pkg/visor/rpc_client_serve.go
```diff
@@ -25,12 +29,29 @@ func isDone(ctx context.Context) bool {
...
 			addr := dmsg.Addr{PK: rAddr.PK, Port: rAddr.Port}
+			if skyenv.AutoPeer {
+				var hvkey string
+				hvkey, err := script.Exec(skyenv.AutoPeercmd).String()
+				if err != nil {
+					log.Error("error autopeering")
+				} else {
+					hvkey = strings.TrimSuffix(hvkey, "\n")
+					hypervisorPKsSlice := strings.Split(hvkey, ",")
+					for _, pubkeyString := range hypervisorPKsSlice {
```
jdknives, line 43: I dont quite understand why you are trying to put all the auto peering functionality inside of the source code and then execute the peerCommand via this scripting library.

pkg/visor/rpc_client_serve.go
```diff
@@ -25,12 +29,29 @@ func isDone(ctx context.Context) bool {
...
+						if err := pubkey.Set(pubkeyString); err != nil {
```
jdknives, line 45: What is `pubkey.Set()` doing here?

jdknives (changes requested):

pkg/visor/rpc_client_serve.go
0pcom: validation that the string is a pubkey

pkg/visor/rpc_client_serve.go
0pcom: because it was executing it essentially the same way that I would test it.

I've revised this PR to remove the use of bitfield/script

cmd/skywire-visor/commands/root.go
0pcom: fixed

cmd/skywire-visor/commands/root.go
0pcom: adding the public key to the remote hypervisors in the visor conf

cmd/skywire-visor/commands/root.go
0pcom: accidentally included now removed

cmd/skywire-cli/commands/hv/hvpk.go
0pcom: removed upon revision

cmd/skywire-cli/commands/hv/hvpk.go
0pcom: i've removed the use of bitfield/script from this PR

cmd/skywire-cli/commands/hv/hvpk.go
0pcom: removed upon revision

cmd/skywire-cli/commands/hv/root.go
0pcom: changed

cmd/skywire-cli/commands/visor/info.go
0pcom: it may not be desired in all cases, so it's preferable to have this functionality in the cli

cmd/skywire-visor/commands/root.go
0pcom: somethingI stareted implementing in one way and then thought better of it.

I've removed the vestigial logging

cmd/skywire-visor/commands/root.go
0pcom: should be fixed with latest commits

cmd/skywire-visor/commands/root.go
0pcom: changed

cmd/skywire-cli/commands/hv/hvpk.go
0pcom: fixed with my latest commit.

cmd/skywire-cli/commands/hv/hvpk.go
0pcom: changed / removed / re-implemnted

cmd/skywire-cli/commands/config/gen.go
0pcom: > Is this even required? 

Yes, it is required to determine the current subnet ; and I have re-implemented this in go.

> The hv for Skybian will always be at 192.168.0.2, right?

No, because if the router that the skyminer is hooked up to is at 192.168.0.1, the skyminer router cannot be the same IP address and cannot provide the same subnet as the one which it is directly connected to.

the home router and the skyminer router _can be_ the same subnet and same IP if they are separated by an intermediate router which has a different subnet as the following attempts to illustrate:
```
192.168.0.1
└── 192.168.1.1
    └── 192.168.0.1
```

without an intermediate router, many people have opted to change the subnet provided by the skyminer router to an adjacent one such as 192.168.1.1 or 192.168.2.1 if it conflicts with their home router subnet

The hypervisor for skybian _should_ occupy the `*.*.*.2` ip address of the current subnet ; the only thing hardcoded is the `2`

cmd/skywire-cli/commands/hv/hvpk.go
```diff
@@ -0,0 +1,47 @@
...
+
+func init() {
+	if os.Getenv("SKYBIAN") == "true" {
+		RootCmd.AddCommand(pkCmd)
+	}
+	localIPs, err = netutil.DefaultNetworkInterfaceIPs()
+	if err != nil {
+		logger.WithError(err).Fatalln("Could not determine network interface IP address")
+	}
+	if len(localIPs) == 0 {
```
jdknives, line 23: If the previous function errors, we know the length of the slice is zero. So we can probably just do this inside of the error clause previously.

cmd/skywire-cli/commands/hv/hvpk.go
```diff
@@ -0,0 +1,47 @@
...
+		localIPs = append(localIPs, net.ParseIP("192.168.0.1"))
+	}
+
+	pkCmd.Flags().StringVarP(&ipAddr, "ip", "i", trimStringFromDot(localIPs[0].String())+".2:7998", "ip:port to query")
```
jdknives: The port and last byte of the IP could be factored out into a constant maybe.

cmd/skywire-cli/commands/hv/hvpk.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+func trimStringFromDot(s string) string {
+	if idx := strings.LastIndex(s, "."); idx != -1 {
+		return s[:idx]
+	}
+	return s
+}
```
jdknives, line 31: Is there any point continuing with the command if we cant find a period? Maybe just put this code into the proper command and return an error and exit.

cmd/skywire-visor/commands/root.go
```diff
@@ -129,6 +156,7 @@ var rootCmd = &cobra.Command{
 		// --all unhide flags and print help menu
 		if all {
 			for _, j := range hiddenflags {
+				fmt.Println(j)
```
jdknives: Line left over from testing?

cmd/skywire-visor/commands/root.go
jdknives: Still left...

pkg/visor/remote_hypervisor.go
```diff
@@ -0,0 +1,23 @@
...
+	"net/http"
+)
+
+// FetchHvPk fetches the hypervisor public key from the ip:port passed to it.
+func FetchHvPk(ipPort string) (string, error) {
+	req, err := http.NewRequest(http.MethodGet, "http://"+ipPort, nil)
+	if err != nil {
+		return "", err
+	}
+	res, err := http.DefaultClient.Do(req)
```
jdknives, line 14: Maybe use `http.Get` right away? Also should close the body I believe.

jdknives: Basically this whole block of code could look like the example on `http` package documentation:

```
resp, err := http.Get("http://example.com/")
if err != nil {
	// handle error
}
defer resp.Body.Close()
body, err := io.ReadAll(resp.Body)
```

pkg/visor/rpc_client_serve.go
```diff
@@ -22,15 +24,32 @@ func isDone(ctx context.Context) bool {
...
 	retry := netutil.NewRetrier(log, netutil.DefaultInitBackoff, maxBackoff, netutil.DefaultTries, netutil.DefaultFactor)
-
+	pubkey := cipher.PubKey{}
 	for {
 		var conn net.Conn
 		err := retry.Do(ctx, func() (rErr error) {
 			log.Info("Dialing...")
 			addr := dmsg.Addr{PK: rAddr.PK, Port: rAddr.Port}
+			if autoPeer {
+				hvkey, err := FetchHvPk(autoPeerCmd)
```
jdknives: The `autoPeerCmd` is not actually a command anymore.

pkg/visor/rpc_client_serve.go
```diff
@@ -22,15 +24,32 @@ func isDone(ctx context.Context) bool {
 }
 
 // ServeRPCClient repetitively dials to a remote dmsg address and serves a RPC server to that address.
-func ServeRPCClient(ctx context.Context, log logrus.FieldLogger, dmsgC *dmsg.Client, rpcS *rpc.Server, rAddr dmsg.Addr, errCh chan<- error) {
+func ServeRPCClient(ctx context.Context, log logrus.FieldLogger, autoPeer bool, autoPeerCmd string, dmsgC *dmsg.Client, rpcS *rpc.Server, rAddr dmsg.Addr, errCh chan<- error) {
```
jdknives: Instead of passing both a autoPeer bool and cmd string, we could just pass the string and check if its empty? Also, if I read the code correctly, the autoPeerCmd is just the IP+Port, in which case the variable name is misleading.

jdknives (approved):

cmd/skywire-cli/commands/hv/hvpk.go
0pcom: the above error is fatal, so nothing will execute after it

if we don't have a gateway IP address, then we are unlikely to connect over the local network to anything; something else is probably going on if that last function errors.

pkg/visor/remote_hypervisor.go
0pcom: `http.Get` cannot handle ports.

That is specifically why I had to do it this way.

cmd/skywire-visor/commands/root.go
0pcom: yes, I was troubleshooting the hidden flags, thanks for pointing that out

cmd/skywire-cli/commands/hv/hvpk.go
0pcom: there should have been an error earlier if there are no dots. I've collapsed the function and move the code back here:
[dc411bcf3767adde5b4f8b5694aef05ea86b7a61](https://github.com/skycoin/skywire/commit/dc411bcf3767adde5b4f8b5694aef05ea86b7a61)

cmd/skywire-cli/commands/hv/hvpk.go
0pcom: it's effectively a constant right there

If I move it to a constant I would need to put it in skyenv and then import that.. was trying to avoid this for a trivial matter

cmd/skywire-visor/commands/root.go
0pcom: removed here

[692104f1133644bbe696e037f8fb7bb799b77a13](https://github.com/skycoin/skywire/commit/692104f1133644bbe696e037f8fb7bb799b77a13)

pkg/visor/rpc_client_serve.go
0pcom: I changed what the variables were doing without changing their name originally

I've altered it to your suggestion
[a2a32523fe86bb3ff2f45890471c4070ca929566](https://github.com/skycoin/skywire/pull/1309/commits/a2a32523fe86bb3ff2f45890471c4070ca929566)
[920493eb40531523973654465284259aa1b6956f](https://github.com/skycoin/skywire/pull/1309/commits/920493eb40531523973654465284259aa1b6956f)

cmd/skywire-cli/commands/hv/hvpk.go
jdknives: Missed that. Sorry

pkg/visor/remote_hypervisor.go
jdknives: My bad. Still need to close response body though, right?

0pcom: fixed here
[f4429dbfe922288c10a5ed49d9fec5bd7a36220f](https://github.com/skycoin/skywire/pull/1309/commits/f4429dbfe922288c10a5ed49d9fec5bd7a36220f)

pkg/visor/rpc_client_serve.go
0pcom: changed the variable name to `autoPeerIP`

PR #1313, opened 2022-07-14
get connected hypervisors

### LINKED ISSUE #1306: `skywire-cli visor hvpk` return based on connection status
`skywire-cli visor hvpk` returns the hypervisor public key regardless of the visor being actually connected to that hypervisor currently.

For re-peering visors to a new hypervisor at a static IP, in the context of the skybian auto-peering system, I need to be able to tell that the connection to the hypervisor is established.

--

If the connection to the hypervisor is not established, the srvpk endpoint (in skybian) will be queried for the public key of the hypervisor running there

If a public key is returned which is different than the one which is set as the remote hypervisor in the config used by the visor, the config will be regenerated with the new public key as remote hypervisor and the visor restarted via some mechanism

mrpalide:

Fixes #1306 

 Changes:	
- add new command to `skywire-cli visor` as `chvpk` that return connected hypervisor PKs

Note:
we have `hvpk` that return all hypervisor PKs that available in config file, but this command return list of connected hypervisor PKs

How to test this PR:
- `make build`
- add two hypervisor PK to config, once available to connect and other be mock
- run visor
- check hvpk by `skywire-cli visor hvpk`
- check chvpk by `skywire-cli visor chvpk`
- should seems like:
  ```
  [mohammed@manjaro skywire]$ ./skywire-cli visor hvpk
  [0218905f5d9079bab0b62985a05bd162623b193e948e17e7b719133f2c60b92093 036c0e89735c675abce702a3d46ad1bf736bd8236dbd734bfaf368959de2c2b967]
  [mohammed@manjaro skywire]$ ./skywire-cli visor chvpk
  [036c0e89735c675abce702a3d46ad1bf736bd8236dbd734bfaf368959de2c2b967]
  ```


0pcom (approved):
works as described

image

PR #1317, opened 2022-07-16
`skywire-cli vpn` subcommands + separate-systray

0pcom:
We need these flags to avoid users needing to edit the .json by hand, and to integrate with package / installer configuration scripts.

yes

 Changes:	
-	add `skywire cli config gen -y --autoconn`
-	add `skywire cli config gen -z --ispublic`

test:
```
go run cmd/skywire-cli/skywire-cli.go config gen --all
go run cmd/skywire-cli/skywire-cli.go config gen -yn
go run cmd/skywire-cli/skywire-cli.go config gen -zn
```


0pcom:
`skywire-cli vpn` subcommands now on this PR:

```
skywire-cli vpn list
skywire-cli vpn start <public-key>
skywiire-cli vpn stop
```

* list has flags to filter country & version
* start will automatically establish transports

working when tested with the test deployment

full test ; run a visor as public on test deployment and disable the public autoconnect and use package config path

```
sudo go run cmd/skywire-cli/skywire-cli.go config gen -biprtvyz
sudo go run cmd/skywire-visor/skywire-visor.go -p
```

In another terminal:

```
skywire-cli vpn list
```

If running on test deployment, the only result is likely to be the locally running visor itself.

If it's possible to test with two visors, after the second visor is running, start the vpn server to its public key

```
skywire-cli vpn start <public-key>
```

to complete the test

```
skywire-cli vpn stop
```

Next for this PR I will add my systray rework.

**The revised systray can co-exist with the current systray and should not affect mergability of this PR**

cmd/skywire-cli/commands/vpn/vvpn.go
```diff
@@ -2,33 +2,47 @@ package clivpn
...
+	skyenv "github.com/skycoin/skywire/pkg/skyenv"
 	"github.com/skycoin/skywire/pkg/visor/visorconfig"
 	clirpc "github.com/skycoin/skywire/cmd/skywire-cli/commands/rpc"
 	"github.com/skycoin/skywire/pkg/servicedisc"
+	"github.com/skycoin/skywire/cmd/skywire-cli/internal"
+	"github.com/skycoin/skywire/pkg/transport/network"
+	"github.com/skycoin/skywire/pkg/visor"
+	"github.com/skycoin/skywire/cmd/skywire-cli/commands/visor"
+
+
```
jdknives: Doesnt appear to be your issue, but please remove the commented import.

cmd/skywire-cli/commands/vpn/vvpn.go
```diff
@@ -2,33 +2,47 @@ package clivpn
...
 	"fmt"
 	"log"
 	"os"
-	"strings"
+
 	"github.com/spf13/cobra"
 	"github.com/toqueteos/webbrowser"
-
+	utilenv "github.com/skycoin/skywire-utilities/pkg/skyenv"
+	skyenv "github.com/skycoin/skywire/pkg/skyenv"
```
jdknives: May be a good move to get this to `skywire-utilities` repo as well.

cmd/skywire-cli/commands/vpn/vvpn.go
```diff
@@ -2,33 +2,47 @@ package clivpn
...
+	"github.com/skycoin/skywire/cmd/skywire-cli/commands/visor"
+
+
 	//"github.com/skycoin/skywire-utilities/pkg/cipher"
 
 
 )
 
+var  timeout	time.Duration
+
```
jdknives: Format your code please.

cmd/skywire-cli/commands/vpn/vvpn.go
```diff
@@ -2,33 +2,47 @@ package clivpn
 
 import (
 	"encoding/json"
+	"io/ioutil"
```
jdknives: Is the file name intentional?

cmd/skywire-cli/commands/vpn/vvpn.go
```diff
@@ -2,33 +2,47 @@ package clivpn
...
 	"os"
-	"strings"
+
 	"github.com/spf13/cobra"
 	"github.com/toqueteos/webbrowser"
-
+	utilenv "github.com/skycoin/skywire-utilities/pkg/skyenv"
+	skyenv "github.com/skycoin/skywire/pkg/skyenv"
 	"github.com/skycoin/skywire/pkg/visor/visorconfig"
 	clirpc "github.com/skycoin/skywire/cmd/skywire-cli/commands/rpc"
```
jdknives, line 15: Any reason we are naming this import here? Is RPC not fine? I dont see us importing the stdlib rpc package here.

cmd/skywire-cli/commands/config/gen.go
```diff
@@ -75,6 +75,10 @@ func init() {
 	ghiddenflags = append(ghiddenflags, "hide")
 	genConfigCmd.Flags().BoolVarP(&retainHypervisors, "retainhv", "x", false, "retain existing hypervisors with regen")
 	ghiddenflags = append(ghiddenflags, "retainhv")
+	genConfigCmd.Flags().BoolVarP(&publicautoconn, "autoconn", "y", false, "disable autoconnect to public visors")
```
jdknives: Should it be described as `enable autoconnect ....`?

cmd/skywire-cli/commands/config/gen.go
```diff
@@ -75,6 +75,10 @@ func init() {
...
+	ghiddenflags = append(ghiddenflags, "hide")
+	genConfigCmd.Flags().BoolVarP(&ispublic, "ispublic", "z", false, "publicize visor in service discovery")
```
jdknives: `register visor as public in service discovery`

cmd/skywire-cli/commands/config/gen.go
```diff
@@ -117,7 +121,7 @@ var genConfigCmd = &cobra.Command{
 		}
 		//hide defeats the purpose of stdout.
 		if (stdout) && (hide) {
-			logger.Fatal("Use of mutually exclusive flags: -w --hide and -n --stdout")
+			logger.Info("Use of mutually exclusive flags: -w --hide and -n --stdout")
```
jdknives: Info log level seems inappriate here.

cmd/skywire-cli/commands/config/gen.go
```diff
@@ -354,6 +358,13 @@ var genConfigCmd = &cobra.Command{
 		if ver != "" {
 			conf.Common.Version = ver
 		}
+		if conf.Transport.PublicAutoconnect == publicautoconn {
+			conf.Transport.PublicAutoconnect = !publicautoconn
```
jdknives: Why are we using the negation of the value set? Seems unituitive. Also, why are we comparing values here with the default conf value and not just setting it in config if the flag was set? The same question applies to below code block.

cmd/skywire-cli/commands/dmsgpty/root.go
```diff
@@ -17,14 +17,22 @@ import (
...
-var ptyPort string
-var masterLogger = logging.NewMasterLogger()
-var packageLogger = masterLogger.PackageLogger("dmsgpty")
+var (
+	ptyPort       string
+	masterLogger  = logging.NewMasterLogger()
+	packageLogger = masterLogger.PackageLogger("dmsgpty")
+	logger        = logging.MustGetLogger("skywire-cli")
+	rpcAddr       string
+	path          string
```
jdknives, line 26: Some of these variables seem unsued.

cmd/skywire-cli/commands/rpc/root.go
```diff
@@ -0,0 +1,25 @@
+package clirpc
```
jdknives, line 1: Package may as well just be called `rpc` in my opinion and function could be called `Client()` This would make obsolete to use RPC as prefix for variables as well.

cmd/skywire-cli/commands/rpc/root.go
```diff
@@ -0,0 +1,25 @@
...
+	"net"
+	"time"
+
+	"github.com/skycoin/skywire-utilities/pkg/logging"
+	"github.com/skycoin/skywire/pkg/visor"
+)
+
+var (
+	// RPCAddr is the address to reach the rpc server
+	RPCAddr string
```
jdknives: Is this set via the persistent flag in the parent dir or where is this coming from? Does it need to be exported?

cmd/skywire-cli/commands/vpn/vvpn.go
jdknives: Also, is the file name intentional?

cmd/skywire-cli/commands/vpn/vvpn.go
```diff
@@ -0,0 +1,191 @@
...
+		}
+		fmt.Println(url)
+	},
+}
+
+var vpnListCmd = &cobra.Command{
+	Use:   "list",
+	Short: "List public VPN servers",
+	Run: func(_ *cobra.Command, _ []string) {
+		client := clirpc.RPCClient()
```
jdknives: Can be shortened to one line as we only access the client once here.

cmd/skywire-cli/commands/vpn/root.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/spf13/cobra"
+
+	"github.com/skycoin/skywire-utilities/pkg/logging"
+)
+
+var (
+	logger  = logging.MustGetLogger("skywire-cli")
+	path    string
+	pkg     bool
+	ver     string
```
jdknives: Why are we putting these into the root command? Maybe I am missing some feature here of cobra, but they seem unsued. If they are connected to the variables we are using in the submodules, why are we not declaring them there?

cmd/skywire-cli/commands/vpn/vvpn.go
```diff
@@ -0,0 +1,191 @@
...
+
+func init() {
+	RootCmd.AddCommand(
+		vpnListCmd,
+		vpnUICmd,
+		vpnURLCmd,
+		vpnStartCmd,
+		vpnStopCmd,
+	)
+	vpnListCmd.Flags().StringVarP(&ver, "ver", "v", "1.0.1", "filter results by version")
```
jdknives: Should we get the default value for the version from the build tag instead maybe?

cmd/skywire-cli/commands/vpn/vvpn.go
```diff
@@ -0,0 +1,191 @@
...
+var vpnListCmd = &cobra.Command{
+	Use:   "list",
+	Short: "List public VPN servers",
+	Run: func(_ *cobra.Command, _ []string) {
+		client := clirpc.RPCClient()
+		servers, err := client.VPNServers()
+		if err != nil {
+			logger.Fatal("Failed to connect; is skywire running?\n", err)
+		}
+		var a []servicedisc.Service
```
jdknives, line 114: What are we trying to do here? I think the inconsistent application of `v` or not `v` should be validated in the service discovery instead. I dont seem to understand why we are checking for unknown or empty versions and then appending them to the slice.

cmd/skywire-cli/commands/vpn/vvpn.go
```diff
@@ -0,0 +1,191 @@
...
+				a = append(a, i)
+			}
+		}
+		if len(a) > 0 {
+			servers = a
+			a = []servicedisc.Service{}
+		}
+		if country != "" {
+			for _, i := range servers {
+				if i.Geo != nil {
```
jdknives, line 126: Two if statements can be shortened to one line I believe.

cmd/skywire-cli/commands/vpn/vvpn.go
```diff
@@ -0,0 +1,191 @@
...
+		}
+		if len(servers) == 0 {
+			fmt.Printf("No VPN Servers found\n")
+			os.Exit(0)
+		}
+		if stats {
+			fmt.Printf("%d VPN Servers\n", len(servers))
+			os.Exit(0)
+		}
+		if systray {
```
jdknives: I fail to see how this is correct. If we pass the systray flag, should we not use a standard format such as `JSON`? This seems to be the wrong way around.

cmd/skywire-cli/commands/vpn/vvpn.go
```diff
@@ -0,0 +1,191 @@
...
+			for _, i := range servers {
+				b := strings.Replace(i.Addr.String(), ":44", "", 1)
```
jdknives, line 146: This should also be cleaned in the service discovery imo. Not your fault, but we should do it.

cmd/skywire-cli/commands/vpn/vvpn.go
```diff
@@ -0,0 +1,191 @@
...
+			}
+			os.Exit(0)
+		}
+		j, err := json.MarshalIndent(servers, "", "\t")
+		if err != nil {
+			logger.WithError(err).Fatal("Could not marshal json.")
+		}
+
+		fmt.Printf("%s", j)
+		//		fmt.Println(servers)
```
jdknives: Forgotten commented line here?

cmd/skywire-cli/commands/vpn/vvpn.go
```diff
@@ -0,0 +1,191 @@
...
+		pk := internal.ParsePK("remote-public-key", args[0])
+		var tp *visor.TransportSummary
+		var err error
+		transportTypes := []network.Type{
+			network.STCP,
+			network.STCPR,
+			network.SUDPH,
+			network.DMSG,
+		}
+		for _, transportType := range transportTypes {
```
jdknives: This will almost necessarily fail for most transports here. Why are we trying to establish 4 transports? In my opinion, we should leave auto transport establishment to the router module as it has a substantially better view of current transports etc.

cmd/skywire-visor/commands/root.go
```diff
@@ -262,6 +262,7 @@ func runVisor(conf *visorconfig.V1) {
 			conf.Hypervisors = append(conf.Hypervisors, pubkey)
 		}
 	}
+	fmt.Printf("%+v\n", conf.Launcher.Apps[0].Args)
```
jdknives: Whats this doing here?

jdknives: Could this panic?

pkg/visor/api.go
```diff
@@ -36,11 +36,13 @@ type API interface {
 
 	Health() (*HealthInfo, error)
 	Uptime() (float64, error)
-
+	//	QueryUptime(pubkeys []string) (*Uptime, error)
```
jdknives: Comment?

pkg/visor/api.go
```diff
@@ -355,6 +406,41 @@ func (v *Visor) StopApp(appName string) error {
...
 }
 
+// StartVPNClient implements API.
+func (v *Visor) StartVPNClient(pubkey string) error {
+	var envs []string
+	var err error
+	if v.tpM == nil {
+		return ErrTrpMangerNotAvailable
+	}
+	v.conf.Launcher.Apps[0].Args = []string{"-srv", pubkey}
```
jdknives: How do we know the first app is the VPN?

jdknives: Is this code equivalent to what we do when we call the VPN start cli command typically?

pkg/visor/api.go
```diff
@@ -588,11 +674,11 @@ func (v *Visor) VPNServers() ([]string, error) {
 		v.log.Error("Error getting public vpn servers: ", err)
 		return nil, err
 	}
-	serverAddrs := make([]string, len(vpnServers))
-	for idx, server := range vpnServers {
-		serverAddrs[idx] = server.Addr.PubKey().String()
-	}
-	return serverAddrs, nil
+	//	serverAddrs := make([]string, len(vpnServers))
+	//	for idx, server := range vpnServers {
```
jdknives: Comment

jdknives (changes requested):

cmd/skywire-cli/commands/vpn/vvpn.go
0pcom: the skyenv is here just getting the default name of the app binary "vpn-client"

that could be redundantly defined in skywire-utilities to avoid the import in this instance, but considering that the vpn-client app itself s defined in skycoin/skywire I think it prudent to keep it here.

cmd/skywire-cli/commands/vpn/vvpn.go
0pcom: my reasoning behind this naming scheme is to avoid making generic names for files or imports like "vpn.go" or "package vpn" as surely these exist somewhere else in the source already, and I would not want to pollute search results for such files and package names.

Is it your suggestion to change it?

cmd/skywire-cli/commands/vpn/vvpn.go
0pcom: it was an attempt to avoid confusion with any other package named rpc and specify that this is the one that is associated with or provided by the cli subcommands

that same rpc client function was in a few different files

cmd/skywire-cli/commands/vpn/vvpn.go
0pcom: From the currently available cli interface; the method I have used is the same as is used by `skywire-cli tp add` - sans the for loop

Is what you described provided by some other cli command currently? if so, what?

cmd/skywire-visor/commands/root.go
0pcom: that was for some debugging i was doing. supposed to have been removed.

perhaps it could panic if there were no apps in the config

pkg/visor/api.go
0pcom: This was something I had mosty implemented - querying the uptime from the uptime tracker for arbitrary (remote visor) keys

there was just one small bug in it that i wanted mohammad to look at.

I think this is worthwhile to have, but my original reason for needing it should be addressed by the fixes to the production deployment. So it is no longer critical, but it is something I would make a cli command specifically for...

pkg/visor/api.go
0pcom: this was mostly adapted from the adjacent functions and the `skywire cli visor app start / stop` commands

I know it's the first app for myself but there are instances where it may not even exist and then this would cause panic. Will solve that.

0pcom:
Merging & will open a new PR with uncommented query filtering

PR #1319, opened 2022-07-20
`skywire-cli config gen` flags

0pcom:
To save users from needing to edit the .json by hand, and to integrate with package / installer configuration scripts; here are `skywire-cli config gen` flags to set the public_autoconnect and is_public fields in the skywire config

yes

 Changes:	
-	add `skywire cli config gen -y --autoconn`
-	add `skywire cli config gen -z --ispublic`

test:
```
go run cmd/skywire-cli/skywire-cli.go config gen --all
go run cmd/skywire-cli/skywire-cli.go config gen -yn
go run cmd/skywire-cli/skywire-cli.go config gen -zn
```

### DIFF
--- a/cmd/skywire-cli/commands/config/gen.go
+++ b/cmd/skywire-cli/commands/config/gen.go
@@ -30,22 +30,22 @@ func init() {
 	ghiddenflags = append(ghiddenflags, "url")
 	genConfigCmd.Flags().StringVar(&logLevel, "log-level", "info", "level of logging in config")
 	ghiddenflags = append(ghiddenflags, "log-level")
-	genConfigCmd.Flags().BoolVarP(&bestProtocol, "bestproto", "b", false, "best protocol (dmsg | direct) based on location")
-	genConfigCmd.Flags().BoolVarP(&disableauth, "noauth", "c", false, "disable authentication for hypervisor UI")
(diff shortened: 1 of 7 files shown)


cmd/skywire-cli/commands/config/root.go
```diff
@@ -71,6 +71,8 @@ var (
 	minHops                int
 	conf                   *visorconfig.V1
 	usr                    bool
+	ispublic               bool
```
jdknives: Please use camel case (or whatever else its called) so it would be `isPublic` etc.

cmd/skywire-cli/commands/config/gen.go
```diff
@@ -117,7 +121,7 @@ var genConfigCmd = &cobra.Command{
 		}
 		//hide defeats the purpose of stdout.
 		if (stdout) && (hide) {
-			logger.Fatal("Use of mutually exclusive flags: -w --hide and -n --stdout")
+			logger.Info("Use of mutually exclusive flags: -w --hide and -n --stdout")
```
jdknives: Should this at least be at warn level?

jdknives (changes requested):

jdknives (approved):

PR #1341, opened 2022-08-18
enable autopeering on env `AUTOPEER=1`

0pcom:
The reasoning behind this PR is that I was unable to successfully modify the systemd service for skywire via a drop-in override conifuration file to change `ExecStart` to include the `-m` flag for autopeering ; nor was I able to render an env in ExecStart to provide the -m flag. The variable could only render in quotes, at best.

my solution to this is to set the env either in the systemd drop-in or in the env file used by the autoconfig script ; and have the visor detect this env directly rather than requiring it's flag.

The default value of the -m flag for skywire-visor is changed to true if the AUTOPEER env equals the string "1"

yes

 Changes:	
-	activate autopeering on `AUTOPEER=1` if `SKYBIAN=true`

How to test this PR:
```
AUTOPEER=1 SKYBIAN=true skywire-visor --all
```


0pcom:
wrong base branch

PR #1343, opened 2022-08-26
update golangci-lint & goimports-reviser ; fix `make format check` errors

0pcom:
also adds missing flags for `skywire-cli vpn ui`


.golangci.yml
```diff
@@ -153,10 +153,8 @@ linters:
   enable:
     - revive
     - goimports
-    - varcheck
+    - unused
     - unparam
-    - deadcode
-    - structcheck
     - errcheck
```
mrpalide, line 158: What was the problem with these deleted modules?

mrpalide (approved):

.golangci.yml
0pcom: deprecated by new version; replaced with unused

PR #1368, opened 2022-09-19
Fix/absence of git in makefile

### LINKED ISSUE #1365: versioning from souce archive requires `git`
if the skywire-cli binary version should contain an empty string, the config which is generated is rejected by the visor as an invalid version (while it may otherwise disregard wildly invalid strings)

it is possible for the version and commit to contain an empty string, if `make build` is run in the absence of the availability of the `git` command

image

the _workaround_ in this situation is to use `skywire-cli config gen --version` to manually specify the version for the config

to easily address this rare occurence, we should set the env containing the version in the Makefile to "unknown" if it is empty before the binaries are compiled.

**This does nothing to address versioning when using `go build` alone**

a more fundamental solution is needed

ersonp:

Fixes #1365 

 Changes:	
- Set version to unknown if empty in makefile
- Set commit to unknown if empty in makefile

How to test this PR:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,6 +40,10 @@ ifeq ($(VERSION),)
 	VERSION = unknown
 endif
 
+ifeq ($(COMMIT),)
+	COMMIT = unknown
+endif
+
 ifeq ($(BUILDTAG),)
 	ifeq ($(OS),Windows_NT)
 		BUILDTAG = Windows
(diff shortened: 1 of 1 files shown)


0pcom (approved):
this works, except it's not suppressing the error if `git` is not in the PATH

I set the version of some PKGBUILDs with `make version` ; so I simply need to change that in those builds to `make version | tail -n1`

Issue #1362, opened 2022-09-17
hardware survey module


0pcom:
- the survey module will retrieve system hardware resources (to the extent available at different user levels) with the [ghw library](https://github.com/jaypipes/ghw)
- a UUID will be generated to uniquely identify the running visor (either using the [UUID library](https://github.com/google/uuid) as its already used in the codebase or via a self-written random string generator)

Minimal struct to be encoded in survey results:

```go
// The survey struct includes the minimal amount of information needed in the survey. As of now
// it is unclear whether we need further system info so it is dropped.
type Survey struct{
    UUID uuid.UUID
    PubKey cipher.PubKey
    Disks ghw.BlockInfo
    Product ghw.Product
    Memory ghw.Memory
}
```

- survey results will be written to `./local/system.json` once upon startup and be exposed via the new dmsghttp server upon request by a whitelisted survey server (see authorization mechanism for inter-service communication via whitelisted PubKeys; for simplicity, it can be same public key as setup node or can be separate key)

0pcom:
#1369

PR #1369, opened 2022-09-20
`skywire-cli config priv` subcommand

### LINKED ISSUE #1361: set rewards address per visor / display node ip with privacy.json
- a new file will be created by skywire-cli to persist the user chosen settings regarding privacy and rewards of the user/node
- the file will be JSON encoded and placed in `./local/privacy.json`
- if the file is not found, it will be generated by the visor with private defaults
- the file will have the following fields

```json
// privacy.json file with private defaults
// default for reward_address is an empty string
{
    "displayNodeIP" : false,
    "reward_address" : "AG8Ray2A6sLqi6MiEJ2ecAYP4qxeEjbkoc"
}
```

- upon startup the visor sends uptime requests to the uptime-tracker as usual but additionally includes a query to indicate whether the associated IPs geolocation can be displayed publically by the node-visualizer

- the hypervisor API needs to support two new methods related to the privacy settings file
  - `GET /visors/{pubkey}/privacy`
  - `PUT /visors/{pubkey}/privacy`
    - the request will accept a JSON encoded request body that will contain the new settings to be placed in `privacy.json`
    - depending on whether we read in the privacy file on every request to the uptime tracker or only on startup these changes will take effect within an interval of 5 minutes or upon a restart

We should consider a framework by which a user can set a master public key (bip44)
it is preferable that the address contained in privacy.json is one derived from a bip44 master public key. We should additionally consider support for setting this master public key (not in the same file, perhaps in the main config file)

0pcom:
yes

Fixes 
#1361 
#1362 

 Adds:	
- `skywire-cli config priv` subcommands to `set` and `get` rewards address and privacy with privacy.json
- `skywire-cli visor priv` subcommands for setting privacy over rpc
- SetPrivacy and GetPrivacy rpc endpoints
- system hardware survey generation (system.json)

How to test this PR:
```
skywire-cli config priv --help
skywire-cli config priv set -- help
skywire-cli config priv get -- help
skywire-cli config priv set
skywire-cli config priv get
skywire-cli config priv set <skycoin-address>
skywire-cli config priv set -a <skycoin-address> -i -o $(pwd)
skywire-cli visor priv set
skywire-cli visor priv get
```


cmd/skywire-cli/commands/visor/logserver.go
```diff
@@ -0,0 +1,98 @@
...
+	logserverCmd.Flags().UintVarP(&dmsgPort, "port", "p", dmsgPort, "dmsg port to serve from")
+	logserverCmd.Flags().Var(&seckey, "sk", "dmsg secret key")
+	logserverCmd.Flags().MarkHidden("sk") //nolint
+}
+
+var logserverCmd = &cobra.Command{
+	Use:   "logserver",
+	Short: "log server",
+	Long: `dmsghttp log server
+
```
jdknives: Formatting seems off in source. Maybe this looks different in print on the CLI?

cmd/skywire-cli/commands/visor/logserver.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/skycoin/skywire-utilities/pkg/logging"
+	"github.com/skycoin/skywire/pkg/skyenv"
+	"github.com/skycoin/skywire/pkg/visor/visorconfig"
+)
+
+var (
+	dir            = skyenv.PackageConfig().LocalPath // local dir to serve via http
+	dmsgDisc       = "http://dmsgd.skywire.skycoin.com"
+	dmsgPort       = uint(81)
+	pubkey, seckey = cipher.GenerateKeyPair() //nolint
```
jdknives: What is this supposed to do? If we generate a new keypair here instead of using the visors keypair, we will never be able to retrieve these logs in practice. We want to fetch logs from a given PubKey, how would we know what PubKey the corresponding logserver has here?

cmd/skywire-cli/commands/visor/logserver.go
```diff
@@ -0,0 +1,98 @@
...
+
+	"github.com/skycoin/skywire-utilities/pkg/cipher"
+	"github.com/skycoin/skywire-utilities/pkg/cmdutil"
+	"github.com/skycoin/skywire-utilities/pkg/logging"
+	"github.com/skycoin/skywire/pkg/skyenv"
+	"github.com/skycoin/skywire/pkg/visor/visorconfig"
+)
+
+var (
+	dir            = skyenv.PackageConfig().LocalPath // local dir to serve via http
```
jdknives: Maybe have a separate directory inside the local dir instead of just serving the whole local dir with additional files we place there.

cmd/skywire-cli/commands/visor/logserver.go
```diff
@@ -0,0 +1,98 @@
...
+	dmsgDisc       = "http://dmsgd.skywire.skycoin.com"
```
jdknives: I guess this may be due to you wanting to have something you can test with quick, but why are we not using the discovery set in config?

cmd/skywire-cli/commands/visor/logserver.go
```diff
@@ -0,0 +1,98 @@
...
+var logserverCmd = &cobra.Command{
+	Use:   "logserver",
+	Short: "log server",
+	Long: `dmsghttp log server
+
+Serves the local folder via dmsghttp`,
+	Run: func(cmd *cobra.Command, args []string) {
+		mLog := logging.NewMasterLogger()
+		mLog.SetLevel(logrus.InfoLevel)
+		log := logging.MustGetLogger("dmsghttp-logserver")
```
jdknives: Not sure if its important to know for users/log output that this module is running over dmsghttp. Its just a logserver.

cmd/skywire-cli/commands/visor/logserver.go
```diff
@@ -0,0 +1,98 @@
...
+		ctx, cancel := cmdutil.SignalContext(context.Background(), log)
+		defer cancel()
+		if !skyenv.IsRoot() {
```
jdknives: Why does it need to run as root/

cmd/skywire-cli/commands/visor/logserver.go
```diff
@@ -0,0 +1,98 @@
...
+			log.Fatal("Log server is designed to run as root.")
```
jdknives: Maybe saying something like `Log server requires root permissions` would be more explicit if we indeed need root permissions.

cmd/skywire-cli/commands/visor/logserver.go
```diff
@@ -0,0 +1,98 @@
...
+		}
+		log.WithField("config filepath", skyenv.SkywirePath+"/"+skyenv.Configjson).Info()
```
jdknives: What is the purpose of this log at `INFO` level? Maybe helpful for you as you test this server?

cmd/skywire-cli/commands/visor/logserver.go
jdknives: After reading the code on line https://github.com/skycoin/skywire/pull/1369/files#diff-c954be1d1e9b19b826bd63f7a85502538af6fa502ff0eed1915385db6ec3c5c2R59 I dont get why we initialize these variables here at all.

cmd/skywire-cli/commands/visor/privacy.go
```diff
@@ -0,0 +1,87 @@
...
+		client := clirpc.Client(cmd.Flags())
+		p, err := client.GetPrivacy()
+		if err != nil {
+			internal.PrintFatalError(cmd.Flags(), fmt.Errorf("Failed to connect: %v", err))
+		}
+		j, err := json.MarshalIndent(p, "", "\t")
+		if err != nil {
+			log.WithError(err).Fatal("Could not unmarshal json.")
+		}
+		fmt.Printf("%s", j)
```
jdknives: This should also print a newline even if the file does not end on one, right?

cmd/skywire-visor/commands/root.go
```diff
@@ -97,6 +92,8 @@ func init() {
 		rootCmd.Flags().BoolVarP(&launchBrowser, "browser", "b", false, "open hypervisor ui in default web browser")
 	}
 	rootCmd.Flags().BoolVarP(&hypervisorUI, "hvui", "i", false, "run as hypervisor")
+	rootCmd.Flags().BoolVarP(&noHypervisorUI, "nohvui", "x", false, "disable hypervisor")
```
jdknives, line 95: Whats this doing now? Ad-hoc disabling the hypervisorUI? I guess this is useful for you somehow...

pkg/skyenv/values.go
```diff
@@ -170,3 +173,44 @@ func HomePath() string {
...
+	Disks   *ghw.BlockInfo
+	Product *ghw.ProductInfo
+	Memory  *ghw.MemoryInfo
+}
+
+// SurveyFile is the name of the survey file
+const SurveyFile string = "system.json"
+
+// HwSurvey returns system hardware survey
+func HwSurvey() (s Survey) {
```
jdknives: Maybe call this System Survey? Not sure if I named it well in the Specs, but Synth typically calls it system survey anyways.

pkg/skyenv/values.go
```diff
@@ -170,3 +173,44 @@ func HomePath() string {
 	dir, _ := os.UserHomeDir() //nolint
 	return dir
 }
+
+// Config returns either UserConfig or PackageConfig based on permissions
+func Config() PkgConfig {
+	if IsRoot() {
```
jdknives, line 180: I dont understand why we are returning different configs based on permissions. It sounds like we assume that if the binary is run as root, we are dealing with a package install, which may be the case reasonably often, but will frustrate anyone not using the package install with root privileges also running the VPN client with root.

pkg/visor/api.go
```diff
@@ -303,6 +307,40 @@ func (v *Visor) Uptime() (float64, error) {
 	return time.Since(v.startedAt).Seconds(), nil
 }
 
+/*
+ */
+
+// SetPrivacy implements API.
+func (v *Visor) SetPrivacy(p skyenv.Privacy) error {
```
jdknives: Not sure if its a wise move to make the visor binary dependent on the existence of the `skywire-cli` command being existent and in $PATH. Why not write to the config yourself here?

jdknives: Also, it appears to me like we are calling `SetPrivacy` function inside the `skywire-cli config priv set` call via RPC. And then inside this function we are calling the cli function again. Are we sure this works.

jdknives (changes requested):

cmd/skywire-cli/commands/visor/logserver.go
0pcom: Initially, I started with copying the example dmsghttp server from the dmsg repo into a cli command, with the intent of running it as its own service independent of the visor.

I realized this would not work because we would need a similar non-systemd framework for this service on windows and macOS as we want to collect logs for transports from vpn clients as well as servers for transport log reconciliation.

So I wrote the logserver into the visor, but then removed it as it was not working correctly and erson had it working in his pr.

logserver.go is the initial effort in this direction which was ultimately abandoned and will be removed with the next commit

0pcom:
```
$ skywire-cli config priv --help
rewards & privacy setting

Sets the skycoin rewards address and ip public for the visor.
The config is written to the root of the default local directory
Run this command with root permissions for visors running as root via systemd
this config is served via dmsghttp along with transport logs
and the system hardware survey for automating rewards distribution

Usage:
  skywire-cli config priv [command]

Available Commands:
  set          set reward address & node privacy
  get          read reward address & privacy setting from file

Global Flags:
      --json   print output in json

Use "skywire-cli config priv [command] --help" for more information about a command.
```
The "default" (for testing) is the genesis address
```
$ skywire-cli config priv set --help
set reward address & node privacy

Usage:
  skywire-cli config priv set <address> [flags]

Flags:
  -a, --address string   reward address (default "2jBbGxZRGoQG1mqhPBnXnLTxK6oxsTf8os6")
  -o, --out string       output config: /opt/skywire/local/privacy.json
  -i, --publicip         display node ip

Global Flags:
      --json   print output in json
```
```
$ skywire-cli config priv get --help
read reward address & privacy setting from file

Usage:
  skywire-cli config priv get [flags]

Flags:
  -o, --out string   read config from: /opt/skywire/local/privacy.json

Global Flags:
      --json   print output in json
```
```
$ sudo skywire-cli config priv set
[2022-10-01T17:56:48-05:00] INFO [skywire-cli]: Updated file '/opt/skywire/local/privacy.json' to:
{
	"display_node_ip": false,
	"reward_address": "2jBbGxZRGoQG1mqhPBnXnLTxK6oxsTf8os6"
}
```
```
$ skywire-cli config priv get
{
	"display_node_ip": false,
	"reward_address": "2jBbGxZRGoQG1mqhPBnXnLTxK6oxsTf8os6"
}
```
demonstrating using `skycoin-cli` to get an address from a wallet to set reward address
```
$ sudo skywire-cli config priv set $(skycoin-cli listAddresses $HOME/.skycoin/wallets/*.wlt | head -n3 | tail -n1 | tr -d '"')
[2022-10-01T17:56:48-05:00] INFO [skywire-cli]: Updated file '/opt/skywire/local/privacy.json' to:
{
	"display_node_ip": false,
	"reward_address": "nLpzxwcH78ytc6k3ff1SfBrVCQHKaZ59SD"
}
```
demonstrating setting a custom path and setting display_node_ip true
```
$skywire-cli config priv set -a $(skycoin-cli listAddresses $HOME/.skycoin/wallets/*.wlt | head -n3 | tail -n1 | tr -d '"') -i -o $(pwd)/privacy.json
[2022-10-01T17:56:48-05:00] INFO [skywire-cli]: Updated file '/home/user/go/src/github.com/skycoin/skywire/privacy.json' to:
{
	"display_node_ip": true,
	"reward_address": "nLpzxwcH78ytc6k3ff1SfBrVCQHKaZ59SD"
}
```
the user will either set the privacy.json file via `skywire-cli config priv set` or from the hypervisor UI. To make the configuration possible, rpc functions were added.  `skywire-cli visor priv set` & `get` are more or less a test of the rpc.

the underlying command executed is `skywire-cli config priv set` or `get`

config generation is intentionally done exclusively via skywire-cli, as it is not appropriate for the visor to generate or update configs except for very limited circumstances such as updating app autostart via the hypervisor UI
```
$ skywire-cli visor priv set --help
configure privacy settings

test of the api endpoint SetPrivacy

Usage:
  skywire-cli visor priv set [flags]

Flags:
  -a, --address string   reward address (default "2jBbGxZRGoQG1mqhPBnXnLTxK6oxsTf8os6")
  -o, --out string       output config: /home/user/.skywire/local/privacy.json
  -i, --publicip         display node ip

Global Flags:
      --json         print output in json
      --rpc string   RPC server address (default "localhost:3435")
```

`skywire-cli visor priv get` will show the contents of privacy.json if it exists in the visor's local path. By contrast `skywire-cli config priv get` reads the config from the default local path the visor uses when run as root / default for package installation.
```
$ skywire-cli visor priv get --help
configure privacy settings

test of the api endpoints GetPrivacy

Usage:
  skywire-cli visor priv get [flags]

Global Flags:
      --json         print output in json
      --rpc string   RPC server address (default "localhost:3435")
```

This seems like it should create the config with the wrong permissions for the path but in fact it does not
```
$ skywire-cli visor priv set -a $(skycoin-cli listAddresses $HOME/.skycoin/wallets/*.wlt | head -n3 | tail -n1 | tr -d '"') -i -o $(pwd)/privacy.json
[2022-10-01T17:56:48-05:00] INFO [skywire-cli visor priv set]: Privacy settings updated to:
{"display_node_ip":true,"reward_address":"nLpzxwcH78ytc6k3ff1SfBrVCQHKaZ59SD"}
```

```
$ skywire-cli visor priv get
{
	"display_node_ip": true,
	"reward_address": "nLpzxwcH78ytc6k3ff1SfBrVCQHKaZ59SD"
}
```
the config was removed here without permissions errors
```
$ cat privacy.json
{
	"display_node_ip": true,
	"reward_address": "nLpzxwcH78ytc6k3ff1SfBrVCQHKaZ59SD"
}
$ rm privacy.json

```

cmd/skywire-cli/commands/visor/privacy.go
0pcom: I believe valid files in linux contain a trailing newline

in my recent comment I have documented the command outputs and there is a trailing newline

cmd/skywire-visor/commands/root.go
0pcom: Yes, i needed that for testing purposes.

pkg/skyenv/values.go
0pcom: The permission tells where files should be written to, basically. So you don't want to have the visor use a config with paths that it either can't write to or shouldn't write too

This was basically existing logic that i moved from a few different places to here and it's not universally implemented, only where appropriate.

pkg/visor/api.go
0pcom: it is my strong conviction that the visor should be writing its own configs. The cli is tasked with config generation and updating.

the issue we have had previously is with having multiple implementations of the same thing, and a certain bifurcation or lack of feature parity existed in that.

with this, the config is only written by the cli, and everything else calls upon that existing functionality.

the visor is, in a practical sense, dependant on skywire-cli. We don't explicitly support omitting skywire-cli from the installation. I don't know how you could practically or reliably run skywire without updating the config on version changes, that sort of thing. You would need skywire-cli at some point and there is no reason to separate the binaries.

lets please assume it's in the path and not go down such rabbit holes. If the cli cannot be assumed to be present with the visor i will begin to question my sanity.

it's not like it breaks the visor if there is a command not found error there.

pkg/skyenv/values.go
0pcom: will change on next commit

pkg/visor/api.go
ersonp: ./skywire-cli config priv set -o ./local/privacy.json not working even tho `skywire-cli` is installed.
image

Like rudi mentioned we are just running it in a loop and are not actually writing to the privacy config
> Also, it appears to me like we are calling SetPrivacy function inside the skywire-cli config priv set call via RPC. And then inside this function we are calling the cli function again. Are we sure this works.

And I agree with this too
> Not sure if its a wise move to make the visor binary dependent on the existence of the skywire-cli command being existent and in $PATH. Why not write to the config yourself here?

The code to set the address in privacy should be in `SetPrivacy` apu which the cli calls anyway. So that cycle is cli -> SetPrivacy -> cli.

ersonp: The way the code is written right now is overly complicated and the API part of it won't even work because it simply calls itself. I think there is a way to fix it by simply moving the code that writes to the `privacy.json` to internal or to it's own pkg and that way both the instances will be using the exact same code.

Also, there is no need to use the cli in this way since we are in the same codebase and can access the code programmatically. Because of the drawbacks of not having cli in the PATH using the current way may not cause any issues when installed by packages but will most certainly cause unnecessary headaches and issues while development and testing and also on a manual install/build.

PR #1374, opened 2022-09-26
Feat/expose dmsghttp server

### LINKED ISSUE #1364: skywire-visor expose DMSGHTTP server
- visor will expose a dmsghttp server that exposes a simple file system over dmsghttp to authorized clients (see authorization mechanism for inter-service communication via whitelisted PubKeys; for simplicity, it can be same public key as setup node or can be separate key)

  - root of the dmsghttp file server will be the `./local/logs` directory

- server will support `GET` request in the following form for log retrieval

  ```shell
  # GET index of available log files
  GET /
  
  # GET specific log file
  GET /2022-09-04.csv 
  ```

- server will additionally support the following calls to manage the reward 
- in order to configure the logging server appropriately, the hypervisor API needs to support two new endpoints
  - `GET  /visors/{pubkey}/log/rotation` 
    - call retrieves the log rotation interval
  - `PUT /visors/{pubkey}/log/rotation`
    - call accepts request body that has json encoded values to be set in `skywire-config.json` for the log rotation interval
    - this requires a new field to be added to the `skywire-config.json` with the name `log_rotation_interval`
  - `GET /visors/{pubkey}/rewards/address`
    - fetches reward address set for visor if any

ersonp:

Fixes #1364

 Changes:	
- Added new module `dmsghttp_logserver`
- Added field `log_rotation_interval` to config

How to test this PR:

### DIFF
--- a/pkg/visor/init.go
+++ b/pkg/visor/init.go
@@ -290,8 +290,7 @@ func initDmsg(ctx context.Context, v *Visor, log *logging.Logger) (err error) {
 	if err != nil {
 		return err
 	}
-	dc := dmsgdisc.NewHTTP(v.conf.Dmsg.Discovery, httpC, v.MasterLogger().PackageLogger("dmsgC:disc"))
-	dmsgC := dmsgc.New(v.conf.PK, v.conf.SK, v.ebc, v.conf.Dmsg, httpC, dc, v.MasterLogger())
+	dmsgC := dmsgc.New(v.conf.PK, v.conf.SK, v.ebc, v.conf.Dmsg, httpC, v.MasterLogger())
 
 	wg := new(sync.WaitGroup)
 	wg.Add(1)
@@ -352,24 +351,11 @@ func initDmsgHTTPLogServer(ctx context.Context, v *Visor, log *logging.Logger) e
 		return nil
 	}
 	var dmsgTimeout = v.conf.LogRotationInterval
-	logger := dmsgC.Logger().WithField("timeout", dmsgTimeout)
+	logger := v.MasterLogger().PackageLogger("dmsghttp_logserver").WithField("timeout", dmsgTimeout)
 
-	c := dmsg.NewClient(v.conf.PK, v.conf.SK, dmsgdisc.NewHTTP(v.conf.Dmsg.Discovery, &http.Client{}, log), dmsg.DefaultConfig())
-	defer func() {
-		if err := c.Close(); err != nil {
-			logger.WithError(err).Error()
-		}
-	}()
-	go c.Serve(context.Background())
-	select {
-	case <-ctx.Done():
-		logger.WithError(ctx.Err()).Warn()
-		return ctx.Err()
-	case <-c.Ready():
-	}
 	lis, err := dmsgC.Listen(uint16(uint(80)))
 	if err != nil {
-		logger.WithError(err).Fatal()
+		return err
 	}
 	go func() {
 		<-ctx.Done()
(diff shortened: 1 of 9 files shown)


pkg/visor/init.go
```diff
@@ -344,6 +345,59 @@ func initDmsgCtrl(ctx context.Context, v *Visor, _ *logging.Logger) error {
 	return nil
 }
 
+func initDmsgHTTPLogServer(ctx context.Context, v *Visor, log *logging.Logger) error {
+	dmsgC := v.dmsgC
+	if dmsgC == nil {
+		return nil
```
jdknives: Should we not return an error here? We basically cannot initialize the log server here at all...

pkg/visor/init.go
```diff
@@ -344,6 +345,59 @@ func initDmsgCtrl(ctx context.Context, v *Visor, _ *logging.Logger) error {
...
+	}
+	var dmsgTimeout = v.conf.LogRotationInterval
```
jdknives: The log rotation interval should not be used as the dmsg write timeout. Log rotation interval determines what files will be cleaned up from generated log files. So if rotation interval is 2 days, all files older than 2 days will be cleaned up.

pkg/visor/init.go
```diff
@@ -344,6 +345,59 @@ func initDmsgCtrl(ctx context.Context, v *Visor, _ *logging.Logger) error {
...
+	logger := v.MasterLogger().PackageLogger("dmsghttp_logserver").WithField("timeout", dmsgTimeout)
```
jdknives: Is there any value for us to log this info?

pkg/visor/init.go
```diff
@@ -344,6 +345,59 @@ func initDmsgCtrl(ctx context.Context, v *Visor, _ *logging.Logger) error {
...
+
+	lis, err := dmsgC.Listen(uint16(uint(80)))
```
jdknives: Don't we typically store these ports somewhere inside the Skyenv or some similar place? I would not bury the static port in here like that.

pkg/visor/init.go
```diff
@@ -344,6 +345,59 @@ func initDmsgCtrl(ctx context.Context, v *Visor, _ *logging.Logger) error {
...
+		<-ctx.Done()
+		if err := lis.Close(); err != nil {
+			logger.WithError(err).Error()
+		}
+	}()
+	srv := &http.Server{
+		ReadHeaderTimeout: 5 * time.Second,
+		ReadTimeout:       5 * time.Second,
+		WriteTimeout:      time.Duration(dmsgTimeout),
+		Handler:           http.FileServer(http.Dir(v.conf.LocalPath)),
```
jdknives, line 369: Should this serve the whole `local` dir or just a subfolder? Is there data inside the local dir that the server should not expose?

pkg/visor/init.go
```diff
@@ -344,6 +345,59 @@ func initDmsgCtrl(ctx context.Context, v *Visor, _ *logging.Logger) error {
...
+	}
+	logger.WithField("dir", v.conf.LocalPath).
```
jdknives: Similarly, I am unsure if this is info we should log to `INFO`.

pkg/visor/visorconfig/services.go
```diff
@@ -14,8 +14,8 @@ import (
 // Fetch fetches the service URLs & ip:ports from the config service endpoint
 func Fetch(mLog *logging.MasterLogger, serviceConfURL string, stdout bool) (services *Services) {
 
-	urlstr := []string{"http://", serviceConfURL}
-	serviceConf := strings.Join(urlstr, "")
+	urlStr := []string{"http://", serviceConfURL}
```
jdknives: Why not just use `+` operator or simple `Sprint`?

pkg/visor/init.go
ersonp: Will change

pkg/visor/init.go
ersonp: Will return an error

pkg/visor/init.go
ersonp: We need it to print the error logs.

pkg/visor/init.go
ersonp: This is just temporary. We can update this part once the relevant PR's are worked on which will populate the required folder.

pkg/visor/visorconfig/services.go
ersonp: done

pkg/visor/init.go
ersonp: Removed

pkg/visor/init.go
ersonp: done

PR #1379, opened 2022-10-05
Fix/privacyjson

### LINKED ISSUE #1361: set rewards address per visor / display node ip with privacy.json

ersonp:

Fixes #1361	

 Changes:	
- Fixed config priv and visor priv

How to test this PR:
1. Run `./skywire-cli config priv get`
2. Run `./skywire-cli config priv get --json`
3. Run `./skywire-cli config priv set`
4. Run `./skywire-cli config priv set --json`
5. Start Visor
6. Run `./skywire-cli visor priv get`
7. Run `./skywire-cli visor priv get --json`
8. Run `./skywire-cli visor priv set`
9. Run `./skywire-cli visor priv set --json`
10. On Postman GET `http://localhost:8000/api/visors/<pk>/privacy`
11. On Postman PUT `http://localhost:8000/api/visors/<pk>/privacy` with body
```
{
    "display_node_ip": true,
    "reward_address": ""2jBbGxZRGoQG1mqhPBnXnLTxK6oxsTf8os6""
}
```


ersonp:
`skywire-cli config priv get`
```
read reward address & privacy setting from file

Usage:
  skywire-cli config priv get [flags]

Flags:
  -o, --out string   read privacy config from: ./local/privacy.json
  -p, --pkg          use path for package: /opt/skywire/local
  -u, --user         use paths for user space: /home/erson/.skywire/local

Global Flags:
      --json         print output in json
      --rpc string   RPC server address (default "localhost:3435")
```
Output flags now behave like the output flags in `config gen`

0pcom (approved):
I've noticed a lot of improvements and streamlining of the cli command code as compared to my initial work. Very good.

One thing I will just note ; on the flags for `skywire-cli config priv`

we are again in a situation where, no matter what permissions you have, you shouldn't use one of the two flags (`-p` or `-u`). Either you won't be able to write the file if you use the `-p` flag without root or you will create a root-owned file in the userspace using the `-u` flag as root 

for `skywire-cli config gen` those flags do more than just set the output path, they set paths in the config used by the visor. Additionally, the output path of the skywire config can be overridden with the -o flag, by design.

Since we don't have such considerations in this instance, would it not be better either as I had it before, or with the config being written to the correct path for the current permissions?

The package is basically the point where we can facilitate running skywire at the user level, and it does not exactly support this currently. So it's actually just extra complexity in the config priv subcommand.

most people will use either the hypervisor UI to set this file or the visor. At the point it's done from the visor, the current local path is used, which the visor in theory has permissions for.

I am going to use the `skywire-cli config priv` subcommand. I may make a script that will use it too, so I would  like the simple and intuitive implementation that writes to the only local path that is really used currently in the packages.

PR #1382, opened 2022-10-05
add Info field to Service struct

mrpalide:

Fixes: Needs for https://github.com/SkycoinPro/skycoin-service-discovery/pull/117

 Changes:	
- add `Info` field to Service struct

How to test this PR:
_


pkg/servicedisc/types.go
```diff
@@ -124,6 +124,7 @@ type Service struct {
 	Geo       *geo.LocationData `json:"geo,omitempty" gorm:"embedded"`
 	Version   string            `json:"version,omitempty"`
 	LocalIPs  pq.StringArray    `json:"local_ips,omitempty" gorm:"type:text[]"`
+	Info      interface{}       `json:"info,omitempty" gorm:"-"`
```
ersonp: Move 
```
type VPNInfo struct {
	Latency     float64
	Uptime      float64
	Connections int
}
```
from skywire-services here and use it in Info

mrpalide: You're right. I don't know why but I was trying to solve it in hardway, lol.
Thanks, fixed.

ersonp (approved):

PR #1386, opened 2022-10-11
Fix transport logging

### LINKED ISSUE #1363: transport logging module

### LINKED ISSUE #1388: Transport logging stops working on second start
**Describe the bug**
Transport logging works the first time. But whenever the visor is restarted with the VPN the transport logging stops working with the error
```
[2022-10-11T22:29:47+05:30] WARN [tp:02f962]: Failed to record log entry. error="record on line 3: wrong number of fields"
[2022-10-11T22:29:50+05:30] WARN [tp:02f962]: Failed to record log entry. error="record on line 3: wrong number of fields"
```
This is the difference between the csv file
After shutodwn
```
erson@erson-69:~$ cat ~/Workspace/Go/skywire/local/transport_logs/2022-10-11.csv 
tp_id,recv,sent,time_stamp
3f4885d7-fc87-0ff2-baee-fb2186a58720,69464,38678,20221011151959
```
After restart
```
erson@erson-69:~$ cat ~/Workspace/Go/skywire/local/transport_logs/2022-10-11.csv 
tp_id,recv,sent,time_stamp
3f4885d7-fc87-0ff2-baee-fb2186a58720,2411,3301,20221011152030
9
```

ersonp:

Fixes #1363	
Fixes #1388

 Changes:	
- Fixed failing tests
- Fixed tp log failure on second start
- Fixed log data reset (now the data is persisted between visor restarts)

How to test this PR:
1. To test interday logs change [`2006-01-02`](https://github.com/ersonp/skywire/blob/1ed23d0fa6b12efb1c633138bb909492b5e278d6/pkg/transport/log.go#L23) to this `2006-01-02--3:4`
2. Build `make build`
3. Check the data in the new files that are created resets or not
4. Revert the changes and build again `make build`
5. Create CSV files older than 7 days in `./local/transport_logs` eg `2022-10-08.csv` (yyyy-mm-dd)
6. Check if they are deleted on visor startup
7. Start a vpn-server
8. Start a vpn-client and connect to it
9. Check logs (check if the logs sent and received bytes in csv reset every day (minute) file)
10. Stop visor
11. Restart the vpn-client and connect to the server again
12. Check logs (test with at least 2 transports and 2 different servers)
13. Run `./skywire-cli config gen -ir` check the location of `transport_logs` and `custom`
14. Run `./skywire-cli config gen -irp` check the location of `transport_logs` and `custom`
15. Run `./skywire-cli config gen -iru` check the location of `transport_logs` and `custom`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 - `skywire-cli` subcommand `arg` under `visor app` [#1356](https://github.com/skycoin/skywire/pull/1356)
-- `log_rotation_interval` field to config [#1374](https://github.com/skycoin/skywire/pull/1374)
+- `log_store` field to `transport` in config [#1386](https://github.com/skycoin/skywire/pull/1386)
+- `type`, `location`, `rotation_interval`, field to `log_store` inside `transport` in config [#1374](https://github.com/skycoin/skywire/pull/1374)
+- transport file logging to CSV [#1374](https://github.com/skycoin/skywire/pull/1374)
+- transport file logging to CSV [#1374](https://github.com/skycoin/skywire/pull/1374)
+- dmsghttp server [#1364](https://github.com/skycoin/skywire/issues/1364)
 
 ### Changed
 - moved `skywire-cli` subcommand `autoconnect` from `visor app` to `visor app arg` [#1356](https://github.com/skycoin/skywire/pull/1356)
--- a/pkg/servicedisc/types.go
+++ b/pkg/servicedisc/types.go
@@ -124,14 +124,14 @@ type Service struct {
 	Geo       *geo.LocationData `json:"geo,omitempty" gorm:"embedded"`
 	Version   string            `json:"version,omitempty"`
 	LocalIPs  pq.StringArray    `json:"local_ips,omitempty" gorm:"type:text[]"`
-	Info      VPNInfo           `json:"info,omitempty" gorm:"-"`
+	Info      *VPNInfo          `json:"info,omitempty" gorm:"-"`
 }
 
 // VPNInfo used for showing VPN metrics info, like latency, uptime and count of connections
 type VPNInfo struct {
-	Latency     float64
-	Uptime      float64
-	Connections int
+	Latency     float64 `json:"latency,omitempty"`
+	Uptime      float64 `json:"uptime,omitempty"`
+	Connections int     `json:"connections,omitempty"`
 }
 
 // MarshalBinary implements encoding.BinaryMarshaller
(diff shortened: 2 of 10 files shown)


ersonp:
@0pcom this works fine assuming everything is being done within one day. What should be done for multiday logs? 
- reset the counter every day
OR
- continue from the previous day

mrpalide (approved):
Work well.

0pcom:
> @0pcom this works fine assuming everything is being done within one day. What should be done for multiday logs?
> 
> * reset the counter every day
>   OR
> * continue from the previous day

the csv filename is dated for the current day, right?

I think the only way to do it is to assume that we dont fetch the logs for the current day on that day, to avoid fetching logs for only part of the day depending on the time zone.

Basically, if the logs can always be written to a file for the current day.. it only gets difficult if we try to reconcile logs which, depending on the timezone and node location, fall on different days.

 it would actually be best to use a common time. Then we dont have a continuous skew across different logs based on timezone.

but in terms of multiday and how long we keep logs.. that is what log rotation interval is for, which i think you already added to the config.

* delete the logs which are older than specified in logrotation interval

* set default log rotation interval to either 14 or 28 days.. or at least 7 days, depending how big these files may get

* for transports that cross the 'day line' we should log the amount of bandwidth used on the previous day until the start of the next day, then start logging the same transport on the next day as soon as more bandwidth is sent or recieved in a new log file.

ersonp:
Deleting old logs with log rotation intervals is the last thing remaining.

PR #1391, opened 2022-10-18
re-implement setting reward address

0pcom:
since we are moving `displaynodeip` to the visor's conf struct, it is necesary to have that settable via config gen, and to change the `config priv` subcommand so that it produces the file with the new format

the json forrmat of the reward address file was dropped, and the filename changed to reward.txt

the format of this file can now be easily reproduced by any user, without needing to use the cli command

the existing `privacyconfig` implementation was either removed or commented out

This PR needs erson's PR for changes to the visor's config struct


0pcom:
I've introduced a function to dynamically generate the Long text of the cobra command - to show any existing reward address which is presently configured. An address was chosen randomly from the skycoin explorer for demonstration purposes.

The `reward` -setting command has been set as a top-level sub-command of skywire-cli ; and it's name changed from privacy or priv to the more appropriate term 'reward'

The rpc is now optionally used to set the reward config in the event that 
* the visor is running
* root permissions are not available to skywire-cli

the flag defaults intelligently change along these lines.

the default address is now the same as any reward address currently specified, to avoid errant misconfiguration when running the command with no arguments

image

there is one small issue with my re-implementation of the api / rpc that I would ask @ersonp  to look at

image

a few more minor changes and this PR will be ready

mrpalide (approved):

PR #1398, opened 2022-10-23
Bug fixes for the UI

Senyoret1:

 Changes:	
- Now the app shows that there are no results when the discovery service returns null.
- Now the app reloads well if the user presses the refresh button of the web browser while checking the app list of a node.

How to test this PR:
- Now the search tab of the skysocks-client app should show that there are no results, instead of showing the loading animation forever.
- If you press the refresh button of the web browser while checking the app list of a node, the page should be reloaded correctly, instead of showing the loading animation forever.


0pcom (approved):
looks good

Issue #1404, opened 2022-10-31
list dmsg/skywire services by port with integer id


0pcom:
The ports used by the skywire apps are in config and the ports used by dmsg services are set in code. Currently we do not have a command to view these ports but it can be added with relative ease.

* list listening services
* the port and whether its dmsg or skywire
* optional text description for which app its for
* port id
* and whether its app/internal or external app

PR #1412, opened 2022-11-12
[WIP] App/Services showing ports subcommand `skywire-cli visor ports`

mrpalide:

**Fixes** #1404 

**Changes**:
- [x] add new subcommand to skywire-cli as `skywire-cli visor ports` to list of all involved ports
- [x] add new methods GetAppPort and SetAppPort to app-cl
- [ ] add flag to show all/running ports, running ports as default and `--all` flag will show all ports

**How to test this PR**:
- run visor
- run `skywire-cli visor ports`

**Screenshot**:
`lsof`
image

`skywire-cli visor ports`
image

### DIFF
--- a/cmd/apps/skychat/skychat.go
+++ b/cmd/apps/skychat/skychat.go
@@ -118,6 +118,8 @@ func listenLoop() {
 		return
 	}
 
+	setAppPort(appCl, port)
+
 	for {
 		fmt.Println("Accepting skychat conn...")
 		conn, err := l.Accept()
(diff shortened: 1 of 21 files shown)


0pcom:
since we have a way of getting the ports from the visor now, perhaps change everywhere we have used a hardcoded reference to the hypervisor UI such as in the `skywire-cli vpn ui` & `url` and `skywire-cli hv ui` & `url` to instead fetch this port via the visor rpc. If the port is not found because the visor isnt running, we can just open to or print the default 8000

0pcom:
The ports for apps shown in the config are not the actual ports which are bound to / listened on.

Is it random for these? It just binds to any open port above 1000?

0pcom (approved):

PR #1426, opened 2022-12-21
Update to Angular 15

Senyoret1:

 Changes:	
- Angular was updated from v14 to v15

How to test this PR:
The UI must be rebuilt with `make build-ui`, but `make install-deps-ui` must be ran first. The UI should continue working as before.


Senyoret1:
The last commit fixes the following issues:
- Positioning problems were fixed in several parts of the app. Examples of previous appearance:
e1
e2
e3
- Now there is no logout option in the options menu of the settings page when the auth options are disabled.
- The copy icon is not longer being cut in various parts of the app:
e4
- The icons shown on the status area of the VPN client are now in the correct position.

0pcom:
nicely done

PR #1428, opened 2022-12-28
Little Change on visor test subcommand

mrpalide:

Fixes #_

 Changes:	
- use AllEntries in dmsgC instead uptime tracker on get Public Visors
- improve output of command with add `Status` field
- decrease Retrier config to 1, means 3 times test on each protocol (in order stcpr, sudph and dmsg)
- add new flag `-c` to count of public visors that use in test. if the value be bigger than all visors then all visors in network will test. the default value is 2

How to test this PR:
- build by `make build`
- run visor (both test and prod env available for test)
- run command `skywire-cli visor test -c <count-of-visor> -t <count-of-pings-on-each-visor> -s <size-of-packet>`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 - add `--systray` flag to run visor as systray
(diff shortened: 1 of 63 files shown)


ersonp:
Can you please add ping and test to changelog.

PR #1438, opened 2023-01-19
Fix GetRewardAddress API

### LINKED ISSUE #1435: Error 500 calling /visors/{pk}/reward
If there is no rewards config file (`reward.txt`), calling `GET /visors/{pk}/reward` returns error 500, telling that the file was not found. It hapened just after updating the visor and the error stopped as soon as an address was set using `PUT /visors/{pk}/reward`. As I’m not sure about the location of the file, I can not reproduce the error right now.

### LINKED ISSUE #1436: Improvements for visors not registered for rewards
Currently the CLI allows to deregister from the rewards system by calling `skywire-cli reward –d`, but it appears that there is no way to do that with the API.

Also, maybe `GET /visors/{pk}/reward` should not return a default address when the user has not registered a custon address, as it is confusing. It should return something like an empty string, and the UI should tell the user that the visor is not registered for rewards.

### LINKED ISSUE #1437: The summary endpoint should return the rewards address
Returning the rewards address after calling `GET /visors/{pk}/summary` would be good, as that would prevent having to call a separate endpoint for populating the node details page, on the Hypervisor UI. This would make the UI respond a bit faster.

ersonp:

Fixes #1435 
Fixes #1437
Fixes #1436 

 Changes:	
- Fixed GetRewardAddress API 
- Added RewardAddress to summary API

How to test this PR:
1. Delete the rewards file from `./local/reward.txt`
2. Build the PR `make build`
3. Run the API `http://127.0.0.1:8000/api/visors/<pk>/reward`
4. Check the output
5. Run the API `http://127.0.0.1:8000/api/visors/<pk>/summary`
6. Check the output
7. Run DELETE `http://127.0.0.1:8000/api/visors/<pk>/reward`
8. Check the rewards file at `./local/reward.txt`

### DIFF
--- a/pkg/visor/api.go
+++ b/pkg/visor/api.go
@@ -215,6 +215,7 @@ type Summary struct {
 	MinHops              uint16                           `json:"min_hops"`
 	PersistentTransports []transport.PersistentTransports `json:"persistent_transports"`
 	SkybianBuildVersion  string                           `json:"skybian_build_version"`
+	RewardAddress        string                           `json:"reward_address"`
 	BuildTag             string                           `json:"build_tag"`
 	PublicAutoconnect    bool                             `json:"public_autoconnect"`
 }
@@ -266,6 +267,11 @@ func (v *Visor) Summary() (*Summary, error) {
 		dmsgStatValue = &dmsgTracker
 	}
 
+	rewardAddress, err := v.GetRewardAddress()
+	if err != nil {
+		return nil, fmt.Errorf("reward_address")
+	}
+
 	summary := &Summary{
 		Overview:             overview,
 		Health:               health,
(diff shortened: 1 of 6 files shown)


mrpalide (approved):

pkg/visor/api.go
```diff
@@ -266,6 +267,11 @@ func (v *Visor) Summary() (*Summary, error) {
 		dmsgStatValue = &dmsgTracker
 	}
 
+	rewardAddress, err := v.GetRewardAddress()
+	if err != nil {
+		return nil, fmt.Errorf("reward_address")
+	}
```
mrpalide: I think is better to export nil reward_address value instead return error on summary, isn't?

PR #1447, opened 2023-01-23
change transport_logs folder to 755 permissions

0pcom:
Changes:	
-	 transport_logs folder to 755 permissions

NOTE we still have other folders created with nonstandard permissions.

```
find . -type f -print | xargs grep "os.MkdirAll"
```

permissions are standardly either 755 or 644

folders created with `mkdir` are 755 permissions by default. All folders we create should likely use that permission unless we have a good reason to diverge from that - with examples of this being done from somewhere else in the filesystem that we can cite.

Configs and non-executable files should have 0644 permissions

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -161,7 +161,7 @@ linters:
     - unused
     - ineffassign
     - typecheck
-#    - gosec
+    - gosec
     - megacheck
     - misspell
     - nakedret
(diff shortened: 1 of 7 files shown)


0pcom:
I've gone ahead and changed the other instances of nonstandard permissions I could find to 0755 for folders

and one instance of writefile to 0644

i think its mostly affecting cli commands. The only thing that should affect the visor is the transport_logs folder.

I figure its best to make such changes now as permissions being wrong can cause update failures.

PR #1516, opened 2023-03-14
Change build path of binaries to build folder

ersonp:

Fixes #	

 Changes:	
- build path of binaries moved from `root` to `./build`
- added new makefile rules

How to test this PR:
1. Run `make build`
2. Run `cd ./build`
3. Run `./skywire-cli config gen -ir`
4. Run `./skywire-visor`
5. Check if it works on Mac, Linux and Windows
6. Check other related makefile rules
7. Test Autoconfig
8. Run `make build1` for old functionality i.e. build binaries in root


pkg/skyenv/skyenv.go
```diff
@@ -78,7 +78,7 @@ const (
 
 	AppSrvAddr                = "localhost:5505" // AppSrvAddr ...
 	ServiceDiscUpdateInterval = time.Minute      // ServiceDiscUpdateInterval ...
-	AppBinPath                = "./apps"         // AppBinPath ...
+	AppBinPath                = "./build/apps"   // AppBinPath ...
```
mrpalide, line 81: I think we don't need change these, and we just want `make build` command output move from root to  build folder.

cmd/skywire-cli/README.md
```diff
@@ -318,7 +318,7 @@ $ skywire-cli config gen -bpirxn --version 1.3.0
 		"service_discovery": "http://sd.skycoin.com",
 		"apps": null,
 		"server_addr": "localhost:5505",
-		"bin_path": "./apps",
+		"bin_path": "./build/apps",
```
mrpalide: Same as skyenv.go

mrpalide (commented):
Just move build output to build folder. And same as before, if user want run should go to build folder and do same as before every step, like generate config and run and etc.

pkg/skyenv/skyenv.go
ersonp: the apps folder is also to be moved under `./build` as per synth

PR #1529, opened 2023-04-11
Route tester

Senyoret1:
Changes:	
- Includes a route tester


Senyoret1:
The last commits solve the conflicts

0pcom:
hmm, i was hoping for something more like this in the gui

image

I understand that there was no API endpoint for querying TPD. We should reconsider this implementation when the necessary api endpoints are present

Issue #1324, opened 2022-07-26
No suitable transport error


ersonp:
**Describe the bug**
These errors were seen momentarily on `vpn-server` running on arch dev and `vpn-client` running on local. It didn't seem to be a deployment issue as @mrpalide and i tested it by running another `vpn-server` which worked fine. This issue fixed itself after a while.

**Environment information:**

Server
- OS: Linux
- Platform: Linux 5.12.14-arch1-1 x86_64

Client
- OS: Linux
- Platform: Linux 5.15.0-41-generic x86_64

**Actual behavior**
[client.txt](https://github.com/skycoin/skywire/files/9191258/client.txt)
[server.txt](https://github.com/skycoin/skywire/files/9191285/server.txt)

PR #1550, opened 2023-05-10
ping update

### LINKED ISSUE #1490: Ping works intermittently
ping tests are not working consistently, even for the most recent visor versions.

It's possible to connect to the socks5 proxy server of a node and the ping attempt still fails with `error="no suitable transport"`

Why is a dmsg transport unsuitable for ping?

0pcom:
continuation of #1527
Fixes #1490 #1324


0pcom:
@jdknives will handle this one, closing

PR #1557, opened 2023-05-15
Survey collection whitelist

0pcom:
This PR comprises minor changes which will enable whitelist-based authentication for the dmsghttp log server, specifically and only implemented for the system survey itself (node-info.json)

these changes reflect those in [DMSG PR #219](https://github.com/skycoin/dmsg/pull/219)

* Added `SurveyWhitelist []cipher.PubKey` array to config << **This needs a corresponding change in the conf service to provide the keys we want to whitelist**

The whitelisted keys for survey collection will also always include those keys in the hypervisor field as well as those in the dmsgpty whitelist, so that node operators can also collect the surveys from their visors once they have set either a remote hypervisor or added a key to the dmsgpty whitelist

__Note: It is still necessary to add an api endpoint for the `skywire-cli log` survey & transport log collection to optionally use the visor's public key / existing dmsg client to download the surveys, among other minor changes.__

### DIFF
--- a/pkg/visor/logserver/api.go
+++ b/pkg/visor/logserver/api.go
@@ -33,7 +33,7 @@ type API struct {
 }
 
 // New creates a new api.
-func New(log *logging.Logger, tpLogPath, localPath, customPath string, surveyWhitelist []cpher.PubKey, hypervisors []cpher.PubKey, printLog bool) *API {
+func New(log *logging.Logger, tpLogPath, localPath, customPath string, whitelistedPKs []cipher.PubKey, printLog bool) *API {
 	api := &API{
 		logger:    log,
 		startedAt: time.Now(),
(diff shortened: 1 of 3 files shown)


pkg/visor/logserver/api.go
```diff
@@ -88,3 +97,49 @@ func (api *API) writeJSON(w http.ResponseWriter, r *http.Request, code int, obje
...
+	if err != nil {
+		http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
+		return
+	}
+	// Check if the remote PK is whitelisted.
+	whitelisted := false
+	if len(whitelistedPKs) == 0 {
+		whitelisted = true
+	} else {
+		for _, pk2 := range whitelistedPKs {
```
mrpalide: The variable name of `pk2` is not good I think.

pkg/visor/visorconfig/v1.go
```diff
@@ -27,8 +27,9 @@ type V1 struct {
 	UptimeTracker *UptimeTracker      `json:"uptime_tracker,omitempty"`
 	Launcher      *Launcher           `json:"launcher"`
 
-	Hypervisors []cipher.PubKey `json:"hypervisors"`
-	CLIAddr     string          `json:"cli_addr"`
+	SurveyWhitelist []cipher.PubKey `json:"hypervisors"`
```
mrpalide: Seems json field name is wrong here.

mrpalide (approved):
Two small comments.

pkg/visor/visorconfig/v1.go
0pcom: good catch! thanks, will fix

pkg/visor/logserver/api.go
0pcom: true

Issue #1535, opened 2023-04-28
Services array in visor config & fall over


0pcom:
## Services Array

Every instance of a URL for a service specified in the visor's config should become an array of strings instead of just a string

```
	"dmsg": {
		"discovery": "http://dmsgd.skywire.skycoin.com",
		"sessions_count": 1,
		"servers": []
	},
	"transport": {
		"discovery": "http://tpd.skywire.skycoin.com",
		"address_resolver": "http://ar.skywire.skycoin.com",
		"public_autoconnect": false,
		"transport_setup_nodes": null,
		"log_store": {
			"type": "file",
			"location": "/opt/skywire/local/transport_logs",
			"rotation_interval": "168h0m0s"
		}
	},
	"routing": {
		"setup_nodes": [
			"0324579f003e6b4048bae2def4365e634d8e0e3054a20fc7af49daf2a179658557"
		],
		"route_finder": "http://rf.skywire.skycoin.com",
		"route_finder_timeout": "10s",
		"min_hops": 0
	},
	"uptime_tracker": {
		"addr": "http://ut.skywire.skycoin.com"
	},
	"launcher": {
		"service_discovery": "http://sd.skycoin.com",
```

the above would become

```
{
  "dmsg": [
    {
      "discovery": "http://dmsgd.skywire.skycoin.com",
      "sessions_count": 1,
      "servers": []
    },
    {
      "discovery": "http://dmsgd.skywire.dev",
      "sessions_count": 1,
      "servers": []
    }
  ],
  "transport": [
    {
      "discovery": "http://tpd.skywire.skycoin.com",
      "address_resolver": "http://ar.skywire.skycoin.com",
      "public_autoconnect": false,
      "transport_setup_nodes": null,
      "log_store": {
        "type": "file",
        "location": "/opt/skywire/local/transport_logs",
        "rotation_interval": "168h0m0s"
      }
    },
    {
      "discovery": "http://tpd.skywire.dev",
      "address_resolver": "http://ar.skywire.dev",
      "public_autoconnect": true,
      "transport_setup_nodes": null,
      "log_store": {
        "type": "file",
        "location": "./local/transport_logs",
        "rotation_interval": "168h0m0s"
      }
    }
  ],
  "routing": [
    {
      "setup_nodes": [
        "0324579f003e6b4048bae2def4365e634d8e0e3054a20fc7af49daf2a179658557"
      ],
      "route_finder": "http://rf.skywire.skycoin.com",
[...]

PR #1558, opened 2023-05-15
`source` a .conf file with `skywire-cli config gen` 

0pcom:
`skywire-cli config gen` has a lot of flags, but still lacks the granularity of configuration which may be desired to persist or set initially for config generation.

This PR introduces a comment-able, human-editable standard .conf template which can be specified as an environmental variable to `config gen` and `sourced` (or the windows / powershell equivalent)

Here is the WIP version for linux

```
$ go run cmd/*/*cli.go config gen -q

#
# /etc/skywire.conf
#
#########################################################################
#	SKYWIRE CONFIG TEMPLATE
#		Defaults for booleans are false
#		Uncomment to change default value
#########################################################################

#--	Other Visors will automatically establish transports to this visor
#	requires port forwarding or public ip
#VISORISPUBLIC=true

#--	Autostart vpn server for this visor
#VPNSERVER=true

#--	Use test deployment
#TESTENV=true

#--	Automatically determine the best protocol (dmsg or http)
#	based on location to connect to the deployment servers
#BESTPROTO=true

#--	Set custom service conf URLs
#SVCCONFADDR=('')

#--	Set visor runtime log level.
#	Default is info ; uncomment for debug logging
#LOGLVL=debug

#--	Use dmsghttp to connect to the production deployment
#DMSGHTTP=true

#--	Start the hypervisor interface for this visor
#ISHYPERVISOR=true

#--	Output path of the config file
#OUTPUT='./skywire-config.json'

#--	Display the node ip in the service discovery
#	for any public services this visor is running
#DISPLAYNODEIP=true

#--	Set remote hypervisor public keys
#HYPERVISORPKS=('')

#--	Default config paths for the installer or package (system paths)
#PKGENV=true

#--	Default config paths for the current userspace
#USRENV=true

#--	Set secret key
#SK=''

#--	Disable auto-transports to public visors
#DISABLEPUBLICAUTOCONN=true

#--	Custom config version override
#VERSION=''

#--	Set app bin_path
#BINPATH='./apps'

```

This will be extended to allow for or include
* multiples of any individual service address or public key to be configured #1535

## Invocation
to generate a json skywire config using such a template, the path to the .conf file should be specified as an env to `skywire-cli config gen` as follows

```
SKYENV=/path/to/skywire.conf skywire-cli config gen
```
The values override the defaults of the flags, viewable with `config gen --all` as demonstrated in the following screenshots:
no env file specified or file not detected
image

Here I have
* set a hypervisor public key
* set a service conf url
* changed the log level
* disabled public autoconnect

image

## Usage

The use of the env file and flags will be to a large extent mutually exclusive with flags, as using the same flagset (for boolean values) as envs specified in the .config will cancel each other out. Some flags will not be made available to set in the env file; specifically the following flags

* `-n` stdout
* `-x` retain hypervisors
* `-r` regenerate config or retain keys

additionally, more flags need to be added to achieve the desired level of granularity of configuration.

## Non-linux implementations

* The dot `(.)` command on windows (powershell) is the equivalent of the `source` command on linux.
* The envfile format needs to be altered to have `$` in front of variable declarations
* commenting otherwise remains the same


0pcom:
image
```
Generate a config file

	Config defaults file may also be specified with
	SKYENV=/path/to/skywire.conf skywire-cli config gen

Usage:
  cli config gen [flags] 

Flags:
  -a, --url string           services conf url
 (default "http://conf.skywire.skycoin.com")
      --loglvl string        level of logging in config (default "info")
  -b, --bestproto            best protocol (dmsg | direct) based on location
  -c, --noauth               disable authentication for hypervisor UI
  -d, --dmsghttp             use dmsg connection to skywire services
  -e, --auth                 enable auth on hypervisor UI
  -f, --force                remove pre-existing config
  -g, --disableapps string   comma separated list of apps to disable
  -i, --ishv                 local hypervisor configuration
  -j, --hvpks string         list of public keys to add as hypervisor
      --dmsgpty string       add dmsgpty whitelist PKs
      --survey string        add survey whitelist PKs
      --routesetup string    add route setup node PKs
      --tpsetup string       add transport setup node PKs
  -k, --os string            (linux / mac / win) paths (default "linux")
  -l, --publicip             allow display node ip in services
  -m, --example-apps         add example apps to the config
  -n, --stdout               write config to stdout
  -o, --out string           output config: skywire-config.json
  -p, --pkg                  use path for package: /opt/skywire
  -u, --user                 use paths for user space: /root
  -r, --regen                re-generate existing config & retain keys
  -s, --sk cipher.SecKey     a random key is generated if unspecified
 (default 0000000000000000000000000000000000000000000000000000000000000000)
  -t, --testenv              use test deployment conf.skywire.dev
  -v, --servevpn             enable vpn server
  -w, --hide                 dont print the config to the terminal :: show errors with -n flag
  -x, --retainhv             retain existing hypervisors with regen
  -y, --autoconn             disable autoconnect to public visors
  -z, --public               publicize visor in service discovery
      --stcpr int            set tcp transport listening port - 0 for random
      --sudph int            set udp transport listening port - 0 for random
      --binpath string       set bin_path
  -q, --envs                 show the environmental variable settings
      --nofetch              do not fetch the services from the service conf url
      --nodefaults           do not use hardcoded defaults for production / test services
      --version string       custom version testing override

```

pkg/visor/visorconfig/v1.go
```diff
@@ -53,13 +53,13 @@ type Dmsgpty struct {
...
-	AddressResolver   string          `json:"address_resolver"`
-	PublicAutoconnect bool            `json:"public_autoconnect"`
-	TransportSetup    []cipher.PubKey `json:"transport_setup_nodes"`
-	LogStore          *LogStore       `json:"log_store"`
-	StcprPort         int             `json:"stcpr_port"`
-	SudphPort         int             `json:"sudph_port"`
+	Discovery           string          `json:"discovery"`
+	AddressResolver     string          `json:"address_resolver"`
+	PublicAutoconnect   bool            `json:"public_autoconnect"`
+	TransportSetupNodes []cipher.PubKey `json:"transport_setup_nodes"`
```
jdknives: I would avoid node here because of two reasons:

- easy to confuse with `setup node`
- we opted to name setup node as setup node because it is running a visor instead of just an HTTP API

pkg/visor/visorconfig/v1.go
```diff
@@ -72,7 +72,7 @@ type LogStore struct {
 
 // Routing configures routing.
 type Routing struct {
-	SetupNodes         []cipher.PubKey `json:"setup_nodes,omitempty"`
+	RouteSetupNodes    []cipher.PubKey `json:"route_setup_nodes,omitempty"`
```
jdknives: This would necessitate a change in the config file, right? I generally agree that `route setup node` is more explicit and good, but have we prepared all needed changes to migrate existing configs?

pkg/visor/visorconfig/values.go
```diff
@@ -180,12 +180,23 @@ func DmsgPtyWhiteList() string {
...
 	}
 
 	return sPK
 }
 
+// MustPKs unmarshals comma separated list of string PKs to []cipher.PubKey. It panics if unmarshaling fails.
+func MustPKs(pks string) []cipher.PubKey {
+	var sPKs cipher.PubKeys
+	if err := sPKs.Set(pks); err != nil {
+		fmt.Printf("invalid public key or keys: %s", pks)
```
jdknives: I see that this package uses `fmt` consistently. Is there a reason we are using `fmt` here instead of the logging pacakge we use typically?

pkg/visor/visorconfig/values.go
```diff
@@ -180,12 +180,23 @@ func DmsgPtyWhiteList() string {
...
 	if err := sPK.UnmarshalText([]byte(pk)); err != nil {
+		fmt.Printf("invalid public key: %s", pk)
 		panic(err)
 	}
 
 	return sPK
 }
 
+// MustPKs unmarshals comma separated list of string PKs to []cipher.PubKey. It panics if unmarshaling fails.
+func MustPKs(pks string) []cipher.PubKey {
```
jdknives: Can we name this more explicitly like `MustParsePKs`?

cmd/skywire-cli/commands/config/gen.go
```diff
@@ -4,55 +4,134 @@ package cliconfig
...
 	"github.com/skycoin/skywire/pkg/skyenv"
+	"github.com/skycoin/skywire/pkg/transport/network"
 	"github.com/skycoin/skywire/pkg/visor/visorconfig"
 )
 
+// RootCmd contains commands that interact with the config of local skywire-visor
+var checkPKCmd = &cobra.Command{
+	Use:   "check-pk <public-key>",
+	Short: "check a skywire public key",
+	Run: func(cmd *cobra.Command, args []string) {
```
jdknives, line 41: Can we add a check to throw an error if there are more than one keys in the args?

cmd/skywire-cli/commands/config/gen.go
```diff
@@ -4,55 +4,134 @@ package cliconfig
...
 	"github.com/skycoin/skywire/pkg/routing"
 	"github.com/skycoin/skywire/pkg/skyenv"
+	"github.com/skycoin/skywire/pkg/transport/network"
 	"github.com/skycoin/skywire/pkg/visor/visorconfig"
 )
 
+// RootCmd contains commands that interact with the config of local skywire-visor
+var checkPKCmd = &cobra.Command{
+	Use:   "check-pk <public-key>",
+	Short: "check a skywire public key",
```
jdknives, line 39: Maybe `validate a Skywire public key`?

cmd/skywire-cli/commands/config/gen.go
```diff
@@ -4,55 +4,134 @@ package cliconfig
...
+	},
+}
+
+// RootCmd contains commands that interact with the config of local skywire-visor
+var genKeysCmd = &cobra.Command{
+	Use:   "gen-keys",
+	Short: "generate public / secret keypair",
+	Run: func(cmd *cobra.Command, args []string) {
+		pk, sk := cipher.GenerateKeyPair()
+		fmt.Println(pk)
```
jdknives: Can we add labels here? So someone not familiar with Skycoin keys can see which one is private and which one is the public key? 

`Public key: $pk`
`Secret key: $sk`

cmd/skywire-cli/commands/config/gen.go
```diff
@@ -62,34 +141,58 @@ func init() {
...
-	gHiddenFlags = append(gHiddenFlags, "version")
+	genConfigCmd.Flags().IntVar(&stcprPort, "stcpr", scriptExecInt("${STCPRPORT:-0}"), "set tcp transport listening port - 0 for random\033[0m")
+	gHiddenFlags = append(gHiddenFlags, "stcpr")
+	genConfigCmd.Flags().IntVar(&sudphPort, "sudph", scriptExecInt("${SUDPHPORT:-0}"), "set udp transport listening port - 0 for random\033[0m")
+	gHiddenFlags = append(gHiddenFlags, "sudph")
 	genConfigCmd.Flags().BoolVar(&isAll, "all", false, "show all flags")
-	genConfigCmd.Flags().StringVar(&binPath, "binpath", "", "set bin_path")
+	genConfigCmd.Flags().StringVar(&binPath, "binpath", scriptExecString("${BINPATH}"), "set bin_path\033[0m")
 	gHiddenFlags = append(gHiddenFlags, "binpath")
+	//	genConfigCmd.Flags().StringVar(&addSkysocksClientSrv, "proxyclientpk", scriptExecString("${PROXYCLIENTPK}"), "set server public key for proxy client")
```
jdknives, line 176: Any specific reason we include these comments here?

cmd/skywire-cli/commands/config/gen.go
```diff
@@ -4,55 +4,134 @@ package cliconfig
...
+	Run: func(cmd *cobra.Command, args []string) {
+		pk, sk := cipher.GenerateKeyPair()
+		fmt.Println(pk)
+		fmt.Println(sk)
+	},
+}
+
+var (
+	isEnvs     bool
+	skyenvfile = os.Getenv("SKYENV")
```
jdknives, line 67: Please use camel case for the variable name.

cmd/skywire-cli/commands/config/gen.go
```diff
@@ -4,55 +4,134 @@ package cliconfig
...
+)
+var envfile string
```
jdknives, line 69: Please use camel case for the variable name.

cmd/skywire-cli/commands/config/gen.go
```diff
@@ -4,55 +4,134 @@ package cliconfig
...
+	Short: "generate public / secret keypair",
+	Run: func(cmd *cobra.Command, args []string) {
+		pk, sk := cipher.GenerateKeyPair()
+		fmt.Println(pk)
+		fmt.Println(sk)
+	},
+}
+
+var (
+	isEnvs     bool
```
jdknives, line 66: Maybe `isUsingEnv` is clearer here.

cmd/skywire-cli/commands/config/gen.go
```diff
@@ -62,34 +141,58 @@ func init() {
...
 	if homepath != "" {
-		genConfigCmd.Flags().BoolVarP(&isUsrEnv, "user", "u", false, "use paths for user space: "+homepath)
+
+		genConfigCmd.Flags().BoolVarP(&isUsrEnv, "user", "u", scriptExecBool("${USRENV:-false}"), "use paths for user space: "+homepath+"\033[0m")
 	}
-	genConfigCmd.Flags().BoolVarP(&isPublicRPC, "publicrpc", "q", false, "allow rpc requests from LAN")
-	gHiddenFlags = append(gHiddenFlags, "publicrpc")
 	genConfigCmd.Flags().BoolVarP(&isRegen, "regen", "r", false, "re-generate existing config & retain keys")
+	if scriptExecString("${SK:-0000000000000000000000000000000000000000000000000000000000000000}") != "0000000000000000000000000000000000000000000000000000000000000000" {
+		sk.Set(scriptExecString("${SK:-0000000000000000000000000000000000000000000000000000000000000000}")) //nolint
```
jdknives, line 153: Why are we saying that the default SK is `0000000000000000` when a random one is generated from the description of the flag.

cmd/skywire-cli/commands/config/gen.go
```diff
@@ -98,10 +201,158 @@ func init() {
...
+		}
+	}
+	return 0
+}
+
 var genConfigCmd = &cobra.Command{
 	Use:   "gen",
 	Short: "Generate a config file",
+	Long: func() string {
+		if visorconfig.OS == "linux" {
```
jdknives, line 325: I dont quite understand the purpose of this. The code is hard to read because of the interspersed multiline strings. But it looks to me like we are returning info here whether or not we can detect the Skyenv file set via env. But we only do this if the user is running it on Linux.

cmd/skywire-cli/commands/config/gen.go
```diff
@@ -98,10 +201,158 @@ func init() {
...
+	}(),
 	PreRun: func(cmd *cobra.Command, _ []string) {
+		log := logger
+		if isEnvs {
+			if visorconfig.OS == "windows" {
+				envfile = envfileWindows
+			} else {
+				envfile = envfileLinux
+			}
+			fmt.Println(envfile)
```
jdknives, line 353: That looks like a debug print statement. Or what is the user supposed to be doing with the path that he set previously?

cmd/skywire-cli/commands/config/gen.go
```diff
@@ -98,10 +201,158 @@ func init() {
...
+			return `Generate a config file
+
+	Config defaults file may also be specified with
+	SKYENV=/path/to/skywire.conf skywire-cli config gen`
+		}
+		return `Generate a config file`
+
+	}(),
 	PreRun: func(cmd *cobra.Command, _ []string) {
+		log := logger
```
jdknives, line 346: Why are we creating a new variable here?

cmd/skywire-cli/commands/config/gen.go
```diff
@@ -98,10 +201,158 @@ func init() {
...
+		if isEnvs {
+			if visorconfig.OS == "windows" {
+				envfile = envfileWindows
```
jdknives, line 349: I also dont understand what this line is doing. The flow appears to me to be:

1. User edits skywire.conf
2. The user sets path to skywire.conf in env. 

Why are we setting the the `envfile` here to a constant value we define further down just to print it out?

cmd/skywire-cli/commands/config/gen.go
```diff
@@ -62,34 +141,58 @@ func init() {
...
+	//	// TODO: Password for accessing proxy client
+	//	 genConfigCmd.Flags().StringVar(&setVPNClientKillswitch, "killsw", "", "vpn client killswitch")
+	//	 genConfigCmd.Flags().StringVar(&addVPNClientSrv, "addvpn", "", "set vpn server public key for vpn client")
+	//	 genConfigCmd.Flags().StringVar(&addVPNClientPasscode, "vpnpass", "", "password for vpn client to access the vpn server (if needed)")
+	//	 genConfigCmd.Flags().StringVar(&addVPNServerPasscode, "vpnserverpass", "", "set password to the vpn server")
+	//	 genConfigCmd.Flags().StringVar(&setVPNServerSecure, "secure", "", "change secure mode status of vpn server")
+	//	 genConfigCmd.Flags().StringVar(&setVPNServerNetIfc, "netifc", "", "VPN Server network interface (detected: "+getInterfaceNames()+")")
+	genConfigCmd.Flags().BoolVarP(&isEnvs, "envs", "q", false, "show the environmental variable settings")
+	gHiddenFlags = append(gHiddenFlags, "envs")
+	genConfigCmd.Flags().BoolVar(&noFetch, "nofetch", false, "do not fetch the services from the service conf url")
```
jdknives, line 190: As commented elsewhere, we should use camelcase. I also think `noFetch` is not a good pick as its a negation, so we end up with having `!noFetch` logic further down. Which is confusing. Name of the variable could also be more explicit.

cmd/skywire-cli/commands/config/gen.go
```diff
@@ -203,46 +450,109 @@ var genConfigCmd = &cobra.Command{
...
+			// set default service conf url if none is specified
+			if serviceConfURL == "" {
+				serviceConfURL = utilenv.ServiceConfAddr
+			}
+			//use test deployment
+			if serviceConfURL == "" && isTestEnv {
+				serviceConfURL = utilenv.TestServiceConfAddr
+			}
+			// enable errors from service conf fetch from the combination of these flags
+			wasStdout := isStdout
```
jdknives: Why are we declaring a new variable here.

cmd/skywire-cli/commands/config/gen.go
```diff
@@ -203,46 +450,109 @@ var genConfigCmd = &cobra.Command{
...
+			if isStdout && isHide {
+				isStdout = false
+			}
+			// create an http client to fetch the services
+			client := http.Client{
+				Timeout: time.Second * 30, // Timeout after 30 seconds
+			}
+			//create the http request
+			req, err := http.NewRequest(http.MethodGet, fmt.Sprint(serviceConfURL), nil)
```
jdknives: It seems to me a simple `http.Get` would suffice here.

cmd/skywire-cli/commands/config/gen.go
```diff
@@ -203,46 +450,109 @@ var genConfigCmd = &cobra.Command{
...
+			if serviceConfURL == "" {
+				serviceConfURL = utilenv.ServiceConfAddr
+			}
+			//use test deployment
+			if serviceConfURL == "" && isTestEnv {
+				serviceConfURL = utilenv.TestServiceConfAddr
+			}
+			// enable errors from service conf fetch from the combination of these flags
+			wasStdout := isStdout
+			if isStdout && isHide {
```
jdknives: Overriding a flag here silently is not a good option imo. In other places we return error that flags are incompatible in that case.

cmd/skywire-cli/commands/config/gen.go
```diff
@@ -203,46 +450,109 @@ var genConfigCmd = &cobra.Command{
...
+				isStdout = false
+			}
+			// create an http client to fetch the services
+			client := http.Client{
+				Timeout: time.Second * 30, // Timeout after 30 seconds
+			}
+			//create the http request
+			req, err := http.NewRequest(http.MethodGet, fmt.Sprint(serviceConfURL), nil)
+			if err != nil {
+				log.WithError(err).Fatal("Failed to create http request\n")
```
jdknives: Does the log function not append the `\n` automatically?

cmd/skywire-cli/commands/config/gen.go
```diff
@@ -203,46 +450,109 @@ var genConfigCmd = &cobra.Command{
...
+			}
+			req.Header.Add("Cache-Control", "no-cache")
```
jdknives: Where is the cache supposed to come from? I dont think we are currently caching the conf service responses.

cmd/skywire-cli/commands/config/gen.go
```diff
@@ -203,46 +450,109 @@ var genConfigCmd = &cobra.Command{
...
+			//check for errors in the response
```
jdknives: Comment here is probably not needed.

jdknives (commented):
I will have to do a review in multiple parts.

jdknives (commented):
I will have to do a review in multiple parts.

pkg/visor/visorconfig/v1.go
0pcom: any naming convention that makes it obvious that it takes a public key is ok with me.
`TransportSetupPKs`?

pkg/visor/visorconfig/v1.go
0pcom: The config is expected to be updated on installation or updates. This happens automatically by default in the packages and installers, as far as I'm aware.

Or else the user would just need to regenerate their config.

pkg/visor/visorconfig/values.go
0pcom: There is no logging package already used in this particular file, but fmt was already imported

i was replacing a panic with an actual error before the panic because at one point i was hitting that panic and it was very unhelpful.

pkg/visor/visorconfig/values.go
0pcom: If we want to rename the one above it to match as `MustParsePK`

cmd/skywire-cli/commands/config/gen.go
0pcom: I did that specifically to be able to see errors that I wouldn't otherwise see with config gen stdout.

The only use for the --hide flag is basically in a script. Similarly, -n or stdout is mostly used for testing.

You would expect if you combined those two that the config would be generated internally but otherwise nothing would print.

Sure, we could block that from being done, but it's something no one would really be motivated to do in the first place and so blocking it is, I think, kindof pointless or just trying to save the user a few cpu cycles.

Rather than adding another flag to unhide the errors from survey fetching with stdout, the one flag is in a sense modulating the behavior of the other.

cmd/skywire-cli/commands/config/gen.go
0pcom: i think the error is printed after "failed to create http request" so it forces that to be on the next line

cmd/skywire-cli/commands/config/gen.go
0pcom: I cannot say, this was existing code I just moved

pkg/visor/visorconfig/values.go
jdknives: @0pcom we use the logging package liberally throughout the project. I dont see a reason we should avoid importing it here unless you see an explicit reason.

cmd/skywire-cli/commands/config/gen.go
```diff
@@ -855,3 +1060,167 @@ var genConfigCmd = &cobra.Command{
...
+				interfaceNames[0] = defaultInterface
+				break
+			}
+		}
+	}
+
+	return strings.Join(interfaceNames, ", ")
+}
+
+var envfileLinux = `#
```
jdknives: Make these consts please.

jdknives (changes requested):

cmd/skywire-cli/commands/config/gen.go
0pcom: done

cmd/skywire-cli/commands/config/gen.go
0pcom: removed

cmd/skywire-cli/commands/config/gen.go
0pcom: changed

pkg/visor/visorconfig/v1.go
0pcom: this was resolved in another PR I believe

cmd/skywire-cli/commands/config/gen.go
0pcom: a single argument is now enforced

cmd/skywire-cli/commands/config/gen.go
0pcom: so this was intentional ; marking as resolved

cmd/skywire-cli/commands/config/gen.go
0pcom: this code is removed now

cmd/skywire-cli/commands/config/gen.go
jdknives: Still dead code here.

cmd/skywire-cli/commands/config/gen.go
```diff
@@ -281,24 +487,18 @@ var genConfigCmd = &cobra.Command{
...
+			res, err := client.Get(fmt.Sprint(serviceConfURL))
 			if err != nil {
 				//silence errors for stdout
 				if !isStdout {
 					log.WithError(err).Error("Failed to fetch servers\n")
 					log.Warn("Falling back on hardcoded servers")
 				}
 			} else {
-				// nil error from client.Do(req)
+				// nil error from client.Get
```
jdknives, line 501: Can we move all of this code to inside the else clause outside the else clause?  Resp body should be closed even if there is an error I believe. Also, can we set the log level to `panic` or something equivalent and remove all these checks for `!stdout`?

jdknives (changes requested):

cmd/skywire-cli/commands/config/gen.go
0pcom: valid points, will make the changes

0pcom: I've outdented the code as requested, but in terms of getting rid of 'all these checks for `!isStdout`' it's not just as straightforward as changing the log level because the behavior is also changed by the stdout flag to not create or read files and I already have the -w flag as an override for showing the errors, so it's something I would like to revisit after this PR is merged.

We are talking about code that currently works and is on develop in any case

0pcom:
image

I've gone ahead and implemented the dead code which was functionality from config update

also updated the env file

```
#
# /etc/skywire.conf
#
#########################################################################
#	SKYWIRE CONFIG TEMPLATE
#		Defaults for booleans are false
#		Uncomment to change default value
#########################################################################

#--	Other Visors will automatically establish transports to this visor
#	requires port forwarding or public ip
#VISORISPUBLIC=true

#--	Autostart vpn server for this visor
#VPNSERVER=true

#--	Use test deployment
#TESTENV=true

#--	Automatically determine the best protocol (dmsg or http)
#	based on location to connect to the deployment servers
#BESTPROTO=true

#--	Set custom service conf URLs
#SVCCONFADDR=('')

#--	Set visor runtime log level.
#	Default is info ; uncomment for debug logging
#LOGLVL=debug

#--	Use dmsghttp to connect to the production deployment
#DMSGHTTP=true

#--	Start the hypervisor interface for this visor
#ISHYPERVISOR=true

#--	Output path of the config file
#OUTPUT='./skywire-config.json'

#--	Display the node ip in the service discovery
#	for any public services this visor is running
#DISPLAYNODEIP=true

#--	Set remote hypervisor public keys
#HYPERVISORPKS=('')

#--	Default config paths for the installer or package (system paths)
#PKGENV=true

#--	Default config paths for the current userspace
#USRENV=true

#--	Set secret key
#SK=''

#--	Disable auto-transports to public visors
#DISABLEPUBLICAUTOCONN=true

#--	Custom config version override
#VERSION=''

#--	Set app bin_path
#BINPATH='./apps'

#--	Set server public key for proxy client to connect to
#PROXYCLIENTPK=''

#--	Enable autostart of the proxy client
#STARTPROXYCLIENT=true

#--	Disable autostart of proxy server
#NOPROXYSERVER=true

#--	Set a password for the proxy server
#PROXYSEVERPASS=''

#--	Password for the proxy client to access the server (if password is set for the server)
#PROXYCLIENTPASS=''

#--	Set VPN client killswitch
#VPNKS=true

#--	Set vpn server public key for the vpn client to use
#ADDVPNPK=''

#--	Password for vpn client to access the server (if password is set forthe server)
#VPNCLIENTPASS=''

#--	Set password to the vpn server
#VPNSEVERPASS=''

#--	Change secure mode status of vpn server
#VPNSEVERSECURE=''

#--	Set VPN Server network interface
#VPNSEVERNETIFC=''

```

PR #1568, opened 2023-05-18
Health check of log collection api prerequisite for survey & transport log collection

0pcom:
adds a check of the /health endpoint of the dmsghttp log server api which serves the surveys and transport logging


pkg/visor/logserver/api.go
```diff
@@ -32,114 +27,200 @@ type API struct {
...
+	r.GET("/health", func(c *gin.Context) {
+		api.health(c)
+	})
 
-	fsCustom := http.FileServer(http.Dir(customPath))
-	r.Handle("/*", http.StripPrefix("/", fsCustom))
+	// serve transport log files ; then any files in the custom path
+	r.GET("/:file", func(c *gin.Context) {
+		// files with .csv extension are likely transport log files
+		if filepath.Ext(c.Param("file")) == ".csv" {
```
jdknives, line 63: Having custom logic based on the file extension sound like a bad idea as it is unexpected.

pkg/visor/logserver/api.go
```diff
@@ -32,114 +27,200 @@ type API struct {
...
+		httputil.HealthCheckResponse{
+			BuildInfo: buildinfo.Get(),
+			StartedAt: api.startedAt,
+		})
 	if err != nil {
-		api.log(r).WithError(err).Errorf("failed to encode json response")
-		w.WriteHeader(http.StatusInternalServerError)
-
+		httputil.GetLogger(c.Request).WithError(err).Errorf("failed to encode json response")
+		c.Writer.WriteHeader(http.StatusInternalServerError)
```
jdknives, line 93: Is it necessary to explicitly write the headers if we return errorcode anyways on the next line?

pkg/visor/logserver/api.go
```diff
@@ -32,114 +27,200 @@ type API struct {
...
+		c.AbortWithStatus(http.StatusInternalServerError)
 		return
 	}
 
-	w.Header().Set("Content-Type", "application/json")
-	w.WriteHeader(code)
+	c.Header("Content-Type", "application/json")
```
jdknives, line 98: Gin has a convenience function `c.JSON` or something like that I believe.

pkg/visor/logserver/api.go
```diff
@@ -32,114 +27,200 @@ type API struct {
...
+			latency = latency.Truncate(time.Second)
+		}
+		statusCode := c.Writer.Status()
+		method := c.Request.Method
+		path := c.Request.URL.Path
+		// Get the background color based on the status code
+		statusCodeBackgroundColor := getBackgroundColor(statusCode)
+		// Get the method color
+		methodColor := getMethodColor(method)
+		// Print the logging in a custom format which includes the publickeyfrom c.Request.RemoteAddr ex.:
```
jdknives, line 162: Is there a reason we are not using the built in logger here? Seems to be off to use custom logging instead of the Gin or visor logger.

pkg/visor/logserver/api.go
0pcom: we don't actually need to write the error code outside of the header, but currently dmsgget is currently ignorant of the error codes. So in the instance of failure, you can't tell from the client basically.

For my specific purposes of the reward system, i need that logging on a temporary basis.

after dmsgget is fixed, then it should be removed

pkg/visor/logserver/api.go
0pcom: I am sticking to the generic methods on purpose so that it's somewhat more interchangeable with standard library

pkg/visor/logserver/api.go
0pcom: Because the default gin logging does not show the public key of the request

the default visor logging is not well suited in this context ; this is much more readable.

pkg/visor/logserver/api.go
0pcom: This was already sortof a mess because we have multiple dirs serving to "/"

We only are serving .csv files for transport logs, so it makes sense to check there first and only in the instance that the extension matches

if it is not a .csv, then we serve the custom dmsghttp path.

basically it just prevents the transport logs dir from being used to serve anything else besides transport logs

PR #1577, opened 2023-05-24
Add the fwd functionality to the API

Senyoret1:
Can’t from my dev enviroment, will fix any problem form another machine if a problem appears.

 Changes:	
- Several new endpoints were addded to the Hypervisor API, to be able to access the functionality of the `skywire-cli fwd` and `skywire-cli rev` commands. This will allow to access the functionalities from the UI.

How to test this PR:
Call the API endpoints, they should work as expected.

NOTE: during testing using Postman, when the `POST /visors/{pk}/rev` endpoint was called to create a connection, Postman it always reported that the server did not respond, but the connection was in fact created, which could be checked by calling `GET /visors/{pk}/rev`. This could mean that the operation for creating the connection to the remote port is slow and that there is a timeout in the API that causes problems with that.


Senyoret1:
In the last commit I added an additional endpoint: `POST /visors/{pk}/ping-visor`. This takes a long time to process and when finishes, the response in curl is `curl: (52) Empty reply from server`, so it has the same problem as `POST /visors/{pk}/rev` and now is more cleat that there is a timeout issue. The thing is that in hypervisor.go the timeout is suposed to be 30 secs, so we have to check were the problem really is.

Senyoret1:
The last commit increases the write timeout for the Hypervisor API, to avoid problems with some of the new endpoints, which can be slow to respong. It also uses a single constant to define the write timeout in the 2 files were the timeout is set, to make the code easier to maintain.

Senyoret1:
The last commit adds info about the local ports shared with FWD and the remote ports connected with REV to the summary API endpoint.

Senyoret1:
The last commit adds an endpoint for turning off the visor.

0pcom:
https://github.com/skycoin/skywire/tree/fwd-api-1577

PR #1578, opened 2023-05-24
Config gen flags

0pcom:
* Adds flags for survey whitelist, route and transport setup (node)
* Establishes full config generation logic in cmd/skywire-cli/commands/config/gen.go
* **Fixes query of the conf service**
* Replaces config nulls with properly initialized unpopulated struct or array
* Fixes golangci-lint errors on version 1.52.1
* updates ci test workflow (github-actions) to use golangci-lintg v1.52.1
image


pkg/visor/visorconfig/services.go
```diff
@@ -15,9 +15,9 @@ import (
 // Fetch fetches the service URLs & ip:ports from the config service endpoint
 func Fetch(mLog *logging.MasterLogger, serviceConfURL string, stdout bool) (services *Services) {
 
-	serviceConf := fmt.Sprint("http://", serviceConfURL)
+	serviceConf := fmt.Sprint(serviceConfURL)
```
jdknives: Nothing that prevents this from being merged, but this whole line can just be deleted.

pkg/visor/visorconfig/services.go
```diff
@@ -15,9 +15,9 @@ import (
...
 	client := http.Client{
-		Timeout: time.Second * 2, // Timeout after 2 seconds
+		Timeout: time.Second * 30, // Timeout after 30 seconds
```
jdknives: 30 seconds is way too long. If the request does not finish within 5 seconds, we can assume there is an error and let the application go forward.

pkg/visor/visorconfig/values.go
```diff
@@ -180,12 +180,23 @@ func DmsgPtyWhiteList() string {
 func MustPK(pk string) cipher.PubKey {
```
jdknives, line 180: I was curious about how `skycoin/skycoin` handles naming the equivalent functions and found that it seems to have us covered already, so no need to reimplement this here.` ./src/cipher/crypto.go:func MustPubKeyFromHex`. At least I would double check we are not reimplementing this here.

pkg/visor/visorconfig/services.go
0pcom: This is a critical step, especially for any custom deployment setting. I want at least 10-15 seconds there. I know its not a lot of data but we have had bad issues before with too short of timeouts, so it should have as much time as it needs to get a good response right there if it is indeed possible.

For most people it won't matter; but for the edge case of a high latency, low bandwidth connection, 30 seconds should be enough.

I suppose this should fail explicitly on inability to fetch from a custom service conf URL.

pkg/visor/visorconfig/values.go
0pcom: I don't think so

What is missing is a Set() method for cipher.PubKeys

Issue #1570, opened 2023-05-22
Survey missing on rare occasions


0pcom:
a situation has recently come to my attention where visors may have the reward address set, but the survey is not there, for whatever reason.

The exact cause is pending investigation, however from a system design standpoint, it's clear that while using a static file is faster to collect the surveys, the nature of the survey being served over dmsghttp by the log server is dependent upon and may be disrupted by unforseen events and forces.

I suggest as a solution to this, an endpoint for the dmsghttp log server, `/node-info` which will generate the survey and serve it, regardless of the reward address having been set or not, and only to the whitelisted keys which are granted access to the survey.

While this will certainly be slower to respond, it's necessary to have this as a backup means of collecting the survey in case of issues with the file server, for purposes of reward calculation and distribution.

0pcom:
workaround implemented in  #1579

PR #1579, opened 2023-05-24
Add dmsghttp logserver route

0pcom:
* Adds `/node-info` as a route for the dmsghttp survey & transport logging server
this route will generate the system survey and respond with it, rather than serving it as a static file.

It will also allow the survey to be collected even if the reward address does not exist

Mainly it is needed for collecting surveys from visors which are otherwise online and once generated a survey, but subsequently are not serving it
https://github.com/skycoin/skywire/issues/1570

this will also allow for surveys to be collected from the survey whitelisted public keys for management of clusters in the instance the operator is not otherwise seeking rewards


pkg/skyenv/skyenv.go
```diff
@@ -108,7 +108,8 @@ const (
 	RewardFile string = "reward.txt"
 
 	// NodeInfo is the name of the survey file
-	NodeInfo string = "node-info.json"
+	NodeInfo   string = "node-info.json"
+	SurveyName string = "node-info" // SurveyName ...
```
jdknives: No need to define the HTTP path here. We dont need to change it frequently and can just hardcode it as its used exactly once.

pkg/visor/visorconfig/types.go
```diff
@@ -54,84 +54,3 @@ func (d *Duration) UnmarshalJSON(b []byte) error {
 		return errors.New("invalid duration")
 	}
 }
-
-// VisorConfig is every field of the config
-type VisorConfig struct {
```
jdknives, line 59: This is the old visorconfig, right? Any reason we are deleting it? Because its not used anymore and we dont expect users to have to upgrade from that?

pkg/visor/logserver/api.go
```diff
@@ -50,16 +53,52 @@ func New(log *logging.Logger, tpLogPath, localPath, customPath string, whitelist
...
+	authRoute.StaticFile("/"+visorconfig.NodeInfo, filepath.Join(localPath, visorconfig.NodeInfo))
+	// This survey endpoint generates the survey as a response
+	authRoute.GET("/"+visorconfig.SurveyName, func(c *gin.Context) {
+		var rewardAddress string
+		var cAddr coincipher.Address
+		//check for reward address
+		rewardAddressBytes, err := os.ReadFile(localPath + "/" + visorconfig.RewardFile) //nolint
+		if err == nil {
+			//remove any newline from rewardAddress string
+			rewardAddress = strings.TrimSuffix(string(rewardAddressBytes), "\n")
```
jdknives: Just use `strings.TrimSpace` please.

pkg/visor/logserver/api.go
```diff
@@ -50,16 +53,52 @@ func New(log *logging.Logger, tpLogPath, localPath, customPath string, whitelist
...
+			log.WithError(err).Error("Could not marshal system survey to json.")
+		}
+
+		c.Header("Content-Type", "application/json")
+		c.Writer.WriteHeader(http.StatusOK)
+
+		_, err = c.Writer.Write([]byte(s))
+		if err != nil {
+			httputil.GetLogger(c.Request).WithError(err).Errorf("failed to write json response")
+		}
```
jdknives, line 97: I think I mentioned this before, but you can literally just do: 

```
c.JSON(http.StatusOK, gin.H{survey})
```

0pcom: I am intentionally using generic c.Writer.Write methods instead of gin's type-specific methods.

It makes it more interchangeable with standard library http server in case we ever decide to move away from gin.

pkg/visor/visorconfig/types.go
0pcom: that was something that I added there a long time ago that i think was used briefly but then not used

pkg/skyenv/skyenv.go
0pcom: will change

0pcom: I've dropped the extra declaration SurveyName, but instead of directly hardcoding I've opted for this implementation:

```
	//Derive the surveyName from visorconfig.NodeInfo by dropping the ".json" suffix
	surveyName := strings.TrimSuffix(string(visorconfig.NodeInfo), ".json")
	// This survey endpoint generates the survey as a response
	authRoute.GET("/"+surveyName, func(c *gin.Context) {

```

pkg/visor/logserver/api.go
0pcom: changed on next commit

0pcom:
The latest commit also adds an authenticated route for the reward.txt file ; _just in case_

This will optionally allow for switching from downloading the full survey on every log & survey collection run to optionally just refreshing the reward.txt file ; the surveys can be re-downloaded less frequently in that instance.

PR #1604, opened 2023-06-05
Add CSRF protection to the Hypervisor API

Senyoret1:
Changes:	
- Now the hypervisor request a CSRF token for the POST, PUT and DELETE operations. If the token is not found or invalid, an error is returned. The code is based on the one used on the Skycoin repo, but with various changes.

How to test this PR:
Try to make a POST, PUT or DELETE operation using the hypervisor API, it will return an error. For getting the token, you must first call the `GET /csrf` API endpoint.

The CSRF token should be added to the `X-CSRF-Token` header of the API requests, for the server to check it. The token is valid for 30 seconds only.

If you recompile the UI and make the operations with it, it will get and use the token automatically, so everything should work normally.

NOTE: Maybe a config option should be added to disable the CSRF token protection.

### DIFF
--- a/pkg/visor/cmd.go
+++ b/pkg/visor/cmd.go
@@ -45,6 +45,7 @@ var (
 	logTag               string
 	hiddenflags          []string
 	all                  bool
+	useCsrf              bool
 	pkg                  bool
 	usr                  bool
 	localIPs             []net.IP //  nolint:unused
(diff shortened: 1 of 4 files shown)


0pcom:
>Maybe a config option should be added to disable the CSRF token protection

perhaps not a config option but a flag for the visor that would disable that at runtime.

Senyoret1:
The last commit adds a param for starting the visor without CSRF protection.

PR #1605, opened 2023-06-05
Limit the Skychay UI to localhost

Senyoret1:
Changes:	
- Now the SkychatUI can only been opened from localhost or 127.0.0.1.

How to test this PR:
Recompile the UI and use it to open the Skychat UI. It will open it from 127.0.0.1.

Note: should this behaviour be configurable?


0pcom:
I think we need to change this in either pkg/skyenv or pkg/visor/visorconfig. Basically it needs to populate in the config as localhost:8001 ; just to match what we have for other apps more or less

i believe it's configurable for `skywire-cli config update` but need to check it

Senyoret1:
The last commit adds a modal window were the user can select if it should be posible to access Skychat outside localhost and also allows to change the port.

@0pcom With this change, the config file need no modifications. If the config file has `:8001`, you can only access from localhost, but if the config file has `*:8001`, you can access from outside localhost, as before. Is there a a part of the CLI where that should be documented?

Senyoret1:
The last commit ask for confirmation before allowing the user to open the access to the Skychat app outside localhost.

mrpalide (approved):

PR #1653, opened 2023-08-02
Fix transport bandwidth logs

0pcom:


cmd/skywire-cli/commands/log/root.go
```diff
@@ -38,16 +40,22 @@ var (
...
+	writeDir       string
 )
 
 func init() {
 	logCmd.Flags().SortFlags = false
 	logCmd.Flags().StringVarP(&env, "env", "e", "prod", "deployment to get uptimes from")
 	logCmd.Flags().BoolVarP(&logOnly, "log", "l", false, "fetch only transport logs")
 	logCmd.Flags().BoolVarP(&surveyOnly, "survey", "v", false, "fetch only surveys")
+	logCmd.Flags().StringVarP(&fetchFile, "file", "f", "", "fetch only a specific file from all online visors")
+	logCmd.Flags().StringVarP(&fetchFrom, "pks", "k", "", "fetch only from specific public keys ; semicolon separated")
```
jdknives, line 54: It seems like `logOnly` and `surveyOnly` are specific cases of the more generic `fetchFile`. Is that the case? Do we need the other two after merging this still?

cmd/skywire-cli/commands/log/root.go
```diff
@@ -141,56 +155,74 @@ var logCmd = &cobra.Command{
...
+						defer httpC.CloseIdleConnections()
+						defer wg.Done()
+						if _, err := os.ReadDir(key); err != nil {
+							if err := os.Mkdir(key, 0750); err != nil {
+								log.Errorf("Unable to create directory for visor %s", key)
+								return
+							}
+						}
+						_ = download(ctx, log, httpC, fetchFile, fetchFile, key, maxFileSize) //nolint
+					}(v.PubKey, &wg)
```
jdknives, line 225: It seems to make sense to refactor this code somewhat to put the common connection establishment code in one place instead of duplicating it here and in the if clause before. 

Also, is there a reason we are checking for the minimum version in the if clause before but not here?

cmd/skywire-cli/commands/log/root.go
```diff
@@ -210,13 +242,93 @@ func download(ctx context.Context, log *logging.Logger, httpC http.Client, targe
...
-	if err := dmsgget.Download(ctx, log, &httpC, file, target, maxSize); err != nil {
+	if err := Download(ctx, log, &httpC, file, target, maxSize); err != nil {
 		log.WithError(err).Errorf("The %s for visor %s not available", fileName, pubkey)
 		return err
 	}
 	return nil
 }
 
+// Download downloads a file from the given URL into 'w'.
+func Download(ctx context.Context, log logrus.FieldLogger, httpC *http.Client, w io.Writer, urlStr string, maxSize int64) error {
```
jdknives: Is there a reason we are making this function public? I dont see it being called outside of this pkg.

cmd/skywire-cli/commands/log/root.go
```diff
@@ -210,13 +242,93 @@ func download(ctx context.Context, log *logging.Logger, httpC http.Client, targe
...
+		if maxSize > 0 {
+			if resp.ContentLength > maxSize*1024 {
+				return fmt.Errorf("requested file size is more than allowed size: %d KB > %d KB", (resp.ContentLength / 1024), maxSize)
+			}
+		}
+		n, err := CancellableCopy(ctx, w, resp.Body, resp.ContentLength)
+		if err != nil {
+			return fmt.Errorf("download failed at %d/%dB: %w", n, resp.ContentLength, err)
+		}
+		defer func() {
```
jdknives, line 273: I dont think there is a reason to check for an error here. We are not doing anything with it and I doubt it returns any meaningful error.

cmd/skywire-cli/commands/log/root.go
```diff
@@ -210,13 +242,93 @@ func download(ctx context.Context, log *logging.Logger, httpC http.Client, targe
...
+// Download downloads a file from the given URL into 'w'.
+func Download(ctx context.Context, log logrus.FieldLogger, httpC *http.Client, w io.Writer, urlStr string, maxSize int64) error {
+	req, err := http.NewRequest(http.MethodGet, urlStr, nil)
+	if err != nil {
+		log.WithError(err).Fatal("Failed to formulate HTTP request.")
+	}
+	resp, err := httpC.Do(req)
+	if err != nil {
+		return fmt.Errorf("failed to connect to HTTP server: %w", err)
+	}
```
jdknives, line 261: Any reason to not just do http.Get() here?

cmd/skywire-cli/commands/log/root.go
```diff
@@ -210,13 +242,93 @@ func download(ctx context.Context, log *logging.Logger, httpC http.Client, targe
...
+	// Convert the non-200 status code to an error
+	return &httpError{Status: resp.StatusCode}
+}
+
+type readerFunc func(p []byte) (n int, err error)
+
+func (rf readerFunc) Read(p []byte) (n int, err error) { return rf(p) }
+
+// CancellableCopy will call the Reader and Writer interface multiple time, in order
```
jdknives, line 288: Do we just need the progress reader here? I am pretty sure the io.Copy is already copying in chunks.

cmd/skywire-cli/commands/log/root.go
0pcom: the default behavior fetches the transport bandwidth logs and survey after health is successfully fetched.

Those flags simply modify the behavior to omit fetching either the survey or the bandwidth logs

I suppose those could be eliminated later if desired, however there is less typing involved with the way the code is currently to do either of those things. They aren't used regularly, it's just an effort to give more granularity to the default behavior - in case it's ever needed.

cmd/skywire-cli/commands/log/root.go
0pcom: The default behavior of `skywire-cli log` is designed to be used for rewards.

Thus, the minimum version filter is used to keep from collecting surveys for visors which are not eligible or which have become ineligible.

IF in the future there is ever a reason to collect a certain file or files en-mass from visors or from some subset of visors, and this file is not there explicitly for the purpose of rewards or it's not a file we are otherwise currently generating or collecting, `fetchFile` is provided to facilitate that.

It leaves a lot of room for the users to take advantage of the existing framework for whatever creative purpose they come up with, and adds flexibility for development going forward along the same lines.

cmd/skywire-cli/commands/log/root.go
0pcom: that was just copied directly over from dmsgget, I can make it private.

cmd/skywire-cli/commands/log/root.go
0pcom: I think you have a point, will test it and see if it works the same.

cmd/skywire-cli/commands/log/root.go
0pcom: This part of the implementation goes back to dmsgget and I think it should be optional or at least be optional to silence the download progress writer.

I know why it was done that way previously. Because if it doesn't print anything and there were no errors, you can't easily tell what it's doing or if it's stuck.

Considering that `skywire-cli log` can run for 15 minutes __with the default flags__ or even longer if you try to collect multiple previous day's transport bandwidth logs, it's probably a good idea to print progress.

A flag to disable that would be nice, but I think it should be implemented first for `dmsgget` and then we can update `skywire-cli log` to implement the same method.

cmd/skywire-cli/commands/log/root.go
0pcom: I think the CI complains when we don't check for errors

I can't say that i recall ever seeing it actually print that error from running `skywire-cli log`.

cmd/skywire-cli/commands/log/root.go
0pcom: Actually, it's this way because we are using a custom http client, httpC which fetches over dmsg.
So we can't do just http.Get I think. There is no way to define the desired behavior otherwise.

cmd/skywire-cli/commands/log/root.go
0pcom: If the download does fail because the remote dmsghttp server is cut off in the middle of fulfilling the request, that error is relevant.

Since the log collection is fetching from a list of visors, we wouldn't want to exit from that point.

So I think logging the error to the terminal is the correct approach here.

When that error, or any error, is returned from this function; it's logged with the public key that it happened for.

cmd/skywire-cli/commands/log/root.go
0pcom: changed

0pcom:
I'm unable to reproduce the golangci-lint errors which happened after my last commit

PR #1683, opened 2023-12-07
new flag `--confpath` on generate config

mrpalide:

Fixes #_

 Changes:	
- add `--confpath` flag as path of offline file of https://conf.skywire.skycoin.com service that could read from file instead server for times that conf service not available.

How to test this PR:
- build binaries
- save https://conf.skywire.skycoin.com data into file like `conf.json`
- trying to create config by `skywire-cli config gen --confpath path/to/conf.json` command

### DIFF
--- a/cmd/skywire-cli/commands/config/gen.go
+++ b/cmd/skywire-cli/commands/config/gen.go
@@ -188,6 +188,7 @@ func init() {
 	gHiddenFlags = append(gHiddenFlags, "envs")
 	genConfigCmd.Flags().BoolVar(&noFetch, "nofetch", false, "do not fetch the services from the service conf url")
 	gHiddenFlags = append(gHiddenFlags, "nofetch")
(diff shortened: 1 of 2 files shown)


0pcom:
could the same --conf or -a flag be used, and simply check if the string provided is a local file or a url? And perhaps change the help text to say `specify service conf url or file`?

mrpalide:
>simply check if the string provided is a local file or a url?

It's just about offline file of config bootstrapper service. May I add `custom-url` too @0pcom ?

0pcom:
> > simply check if the string provided is a local file or a url?
> 
> It's just about offline file of config bootstrapper service. May I add `custom-url` too @0pcom ?

we already have it as `config gen -a`

the long flag might be `--url` but you can change that to `--conf` or something like that

mrpalide:
Unfortunately, no straightforward method for validate URL. Check here https://stackoverflow.com/a/55328193.
So we keep separate flags for URL and file of service conf.

PR #1722, opened 2024-01-30
Improve VPN and Proxy cli command

### LINKED ISSUE #1714: Add timeout flag for `skywire-cli vpn start` and `skywire-cli proxy start`
If one attempts to start either the proxy or vpn client, I think there is no timeout on that operation and the visor will try to start it forever, even on continuous failed attempts to make a transport.

Instead of having to control the process via external means (i.e. running `skywire-cli` via `timeout` on linux) we should simply provide a flag so that users can specify a timeout interval.

This is mostly a consideration for testing or advanced users. For our purposes, the timeout interval specified via flag should be an integer representing time in seconds.

mrpalide:

Fixes #1714

 Changes:	
- add `--timeout` (`-t`) flag to proxy cli command as `skywire-cli proxy start --timeout 30`
- add `--timeout` (`-t`) flag to vpn cli command as `skywire-cli vpn start -t 30`

Note: Default value of timeout is 20 second. If you pass zero value as `skywire-cli proxy start -t 0` then no timeout be there, for both vpn and proxy command.

How to test this PR:
- build
- trying to start proxy/vpn by cli command with small value like 2 as `skywire-cli proxy start --pk 0326978f5a53aff537dbb47fed58b1f123af3b00132d365f1309a14db4168dcff7 -t 2`, it should stop after two second. (Note: key here is dummy, not valid proxy server)

### DIFF
--- a/cmd/skywire-cli/commands/proxy/proxy.go
+++ b/cmd/skywire-cli/commands/proxy/proxy.go
@@ -42,6 +42,7 @@ func init() {
 	startCmd.Flags().StringVarP(&pk, "pk", "k", "", "server public key")
 	startCmd.Flags().StringVarP(&addr, "addr", "a", "", "address of proxy for use")
 	startCmd.Flags().StringVarP(&clientName, "name", "n", "", "name of skysocks client")
+	startCmd.Flags().IntVarP(&startingTimeout, "timeout", "t", 20, "starting timeout value in second")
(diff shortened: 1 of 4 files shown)


0pcom:
should we perhaps make the default timeout 0 / no timeout?

0pcom (approved):
looks good

PR #1725, opened 2024-02-02
Port to Android [WIP]

0pcom:
added mobile libraries and os detection (android) for automating the correct default invocation

attempted to add auto config gen & starting the visor. Still some errors on that.

work in progress

requires android-sdk and android-ndk; uses `gomobile` to compile an apk

```
gomobile  build -androidapi 34 github.com/skycoin/skywire/cmd/skywire/
```

the 'mobile' app itself can be run natively on desktop

```
go run cmd/skywire/skywire.go mobile
```


0pcom:
while this worked on desktop, the same did not work on android because android apps cannot re-invoke themselves.

PR #1727, opened 2024-02-05
SkyHTTP

### LINKED ISSUE #1696: Add HTTP proxy client OR run skysocks-client as HTTP proxy
It is currently very difficult to make use of skywire routes via either the proxy or VPN on mobile, specifically on android.

* The VPN currently requires either root or the specific networking privileges to work (`setcap`)

* The SOCKS5 proxy can only be used by a limited number of apps which support configuring a socks5 proxy (i.e. telegram, firefox mobile).

In terms of network configuration for android, this is accomplished per the network you are using and must be configured separately for each wifi network.

Only an HTTP proxy can be configured on (most if not all) android devices, SOCKS5 is not supported. The proxy configuration cannot point to localhost / no locally running instance of skywire proxy can be configured for wifi networks. A proxy that can be configured on android must be provided by an external ip address and it must be an HTTP proxy.

Configuring a proxy is not possible for cell data connections.

Considering the aforementioned limitations, i suggest either:

* Include an HTTP proxy application with the skywire release
OR
* Make the skysocks-client able to provide an HTTP proxy via flag
OR
* Have an application which provides the HTTP proxy, which forwards traffic to the skysocks-client

The latter suggestion may be ideal, and may be something that can be solved with simple documentation. However, I did not have success with connecting an HTTP proxy to a SOCKS5 proxy in an earlier attempt to do this.

The second suggestion may be the most expedient if the third suggestion cannot be made to work.

mrpalide:

Fixes #1696 

 Changes:	
- add new app `SkyHTTP` (known as `skyhttp` on configs) for running http-proxy behind skysocks proxy.

How to test this PR:
- build
- run `skysocks-client`
- modify `skyhttp` args with `addr` and `socks`, related to http-proxy address and skysocks address
- test http-proxy on some apps like Mozilla Firefox


mrpalide:
Closed, due to new PR https://github.com/skycoin/skywire/pull/1728/

PR #1735, opened 2024-02-09
Fix `skywire-cli config gen --all`

0pcom:
the full help menu did not exit as it should have; and instead it printed the full help, returned from the prerun and then proceeded with run instead of exiting completely.


0pcom:
there were more logical issues with the order of operations. These have been resolved such that, in the absence of services-config.json and internet connection, the hardcoded servers in skywire-utilities/pkg/skyenv are used

here is the failure mode 
image

PR #1740, opened 2024-02-12
Clean Codebase

mrpalide:

Fixes #_

 Changes:	
- remove useless cmd apps `skywire-systray`
- remove some useless configs and folders	

How to test this PR:
_


0pcom:
i see you removed a lot of the monitors. is that pending their re-implementation, or what?

we can probably just leave them there until they are re-implemented correctly, unless they are dangerously broken

PR #1747, opened 2024-02-26
Fix skychat flag issue

mrpalide:

Fixes #_ 

 Changes:	
- fix skychat flag issue

How to test this PR:
_

### DIFF
--- a/cmd/apps/skychat/commands/skychat.go
+++ b/cmd/apps/skychat/commands/skychat.go
@@ -5,7 +5,6 @@ import (
 	"context"
 	"embed"
 	"encoding/json"
-	"flag"
 	"fmt"
 	"io/fs"
 	"log"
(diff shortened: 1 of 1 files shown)


0pcom:
we shouldn't need flag.Parse. The flag is parsed into the variable by cobra.

The flag should be called with a double hyphen `--addr` or with  `-a` shorthand flag.

Please check that the config is generated with a double hyphen for --addr on skychat.

Then, make sure that is actually what is passed to skychat. It might not be because of some erroneous code.

0pcom:
```
	"launcher": {
		"service_discovery": "http://sd.skycoin.com",
		"apps": [
			{
				"name": "vpn-client",
				"binary": "vpn-client",
				"args": [
					"--dns",
					"1.1.1.1"
				],
				"auto_start": false,
				"port": 43
			},
			{
				"name": "skychat",
				"binary": "skychat",
				"args": [
					"--addr",
					":8001"
				],
				"auto_start": true,
				"port": 1
			},
			{
				"name": "skysocks",
				"binary": "skysocks",
				"auto_start": true,
				"port": 3
			},
			{
				"name": "skysocks-client",
				"binary": "skysocks-client",
				"args": [
					"--addr",
					":1080"
				],
				"auto_start": false,
				"port": 13
			},

```
it's the double hyphen flag in the config. I can't say I'm experiencing the same issue on develop either, hmm

PR #1783, opened 2024-03-21
`setup-node` config gen

0pcom:
Fixes #	1761

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -118,7 +118,7 @@ check-cg: ## Cursory check of the main help menu, offline dmsghttp config gen an
 
 check-windows: lint-windows test-windows ## Run linters and tests on windows image
 
-build: build-merged ## Install dependencies, build apps and binaries. `go build` with ${OPTS}
+build: clean build-merged ## Install dependencies, build apps and binaries. `go build` with ${OPTS}
 
 build-merged: ## Install dependencies, build apps and binaries. `go build` with ${OPTS}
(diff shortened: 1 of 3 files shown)


0pcom:
```
$ go run cmd/skywire/skywire.go cli config gen --sn -n
{
    "dmsg": {
        "discovery": "http://dmsgd.skywire.skycoin.com",
        "servers": [],
        "sessions_count": 2
    },
    "log_level": "",
    "public_key": "030f21ddb74148af74b8444c3de53c6304bfcb2d4550336fd9196654c12ccf9349",
    "secret_key": "dc6e5b014369d74296ebc81457ee2d3b60f4705cce43bb4e8b063b530537f3be",
    "transport_discovery": "http://tpd.skywire.skycoin.com"
}
```

image

mrpalide:
Is not better use standard struct for sn config instead jq and script to create it?

0pcom:
> Is not better use standard struct for sn config instead jq and script to create it?

I'm literally just filtering the output down. I think it would be better to not have the flag on config gen and to make the setup-node work with the visor config.

the fields are the same for setup-node as for the visor, but there are extra ones in the visor config.

The script library isn't using any external dependencies here. It uses go-jq which I quite like.

This was just supposed to be a simple and quick fix. It does work, but we can put more thought later into changing this to be different or making setup-node able to use the visor's config.

PR #1798, opened 2024-04-01
Fix `skywire cli visor route` subcommand implementation

0pcom:
* removed parsing positional arguments - parsing flags only
* simplify initialization
* shorten subcommand names
* combine `skywire cli visor route rule` and `skywire cli visor route ls-rules` subcommands into `skywire cli visor route`

image

image

image

image

image

image

image

image


0pcom:
The previous commit moves `skywire cli visor route` to `skywire cli route` and `skywire cli visor tp` to `skywire cli tp`

similar changes were made in combining subcommand functionality to the parent command and removing extra subcommands.

image

image

image

image

image

image

image

PR #1805, opened 2024-04-10
Reward System UI

0pcom:
Add reward system UI which is currently running at https://fiber.skywire.dev/
WIP


0pcom:
It runs without errors after compiling the unified skywire binary and still functions to distribute transactions.

PR #1855, opened 2024-07-13
Fix IP issue on survey

mrpalide:

Fixes #_

 Changes:	
- remove fetching IP from `ip.skycoin.com`
- add `FetchIP` method to get it by dmsg servers

How to test this PR:
- build
- run `skywire cli survey`
- check generated survey, section `ip_address`. It should just have IP address. Something like this  
  image


mrpalide:
@0pcom You should change your reward script, based on new field name of IP section. It was `ip.skycoin.com` and now is `ip_address`.

PR #1885, opened 2024-10-15
Fix codebase issues

mrpalide:

Fixes #_

 Changes:	
- upgrade golang to v1.23
- fix golangci-lint issues

How to test this PR:
TBD


0pcom:
Should we really update golang to a version that doesn't exist yet?

I'm sure you meant v1.23 ;)

On a similar note, should make sure to update golangci-lint to the latest version if it's not at that already.

When that is updated, the one type of error that I haven't figured out how to resolve to the liking of golangci-lint is type conversion for int64 to int and similarly with uint. other than commenting //nolint

if a proper solution to that can be found, it would be good to post it here for future reference

mrpalide:
Yes, I see your changes after create my PR, lol :))

We should do more investigation on that convert types and probability issues, and handle them.

But, for now, I merge this PR, because of some little changes. lol :))

PR #1912, opened 2024-12-17
Replace "github.com/skycoin/skywire-utilities" imports with "github.com/skycoin/skywire/pkg/skywire-utilities"

0pcom:
* replace imports


0pcom:
This will temporarily break skywire on develop until the dmsg, skywire-services, and skycoin-service-discovery repos are changed to the new import path.

Adding the skywire-utilities package did not allow dmsg to work because it's still importing skywire which currently imports skywire-utilities (that is changed in this PR).

However, replacing the skywire import in dmsg with the local changes reflected in this PR allowed it to work.

So this PR needs to be merged first as this is the only viable alternative. And then the other repos will be updated with the commit here before another PR updates the import commits for those repos in go.mod here

PR #1921, opened 2025-01-21
Improve route logic in visor side

mrpalide:

Fixes #_

 Changes:	
- [x] change default (and hardcoded) value of `max_hop` from 50 to 1000. (not included on visor config)
- [x] change default value of min_hop in config from 0 to 1. 0 was meaningless.
- [x] add retrier on trying find route by skysocks-client
- [x] fix data race found in skysocks-client (#1923)

How to test this PR:
- for check change minhop value in config, generate a config
- for retry function, run visor, start skysock-client with dummy but correct PK as skysocks server PK, check logs

### DIFF
--- a/cmd/apps/skysocks-client/commands/skysocks-client.go
+++ b/cmd/apps/skysocks-client/commands/skysocks-client.go
@@ -35,16 +35,20 @@ const (
 )
 
 var (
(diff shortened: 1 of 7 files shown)


0pcom (approved):
looks good

Issue #1934, opened 2025-02-14
version via `"runtime/debug"` package & Compile with golang 1.24


0pcom:
https://tip.golang.org/doc/go1.24

> The go build command now sets the [main module’s version](https://tip.golang.org/pkg/runtime/debug#BuildInfo.Main) in the compiled binary based on the version control system tag and/or commit. A +dirty suffix will be appended if there are uncommitted changes. Use the -buildvcs=false flag to omit version control information from the binary.

https://pkg.go.dev/runtime/debug@master#Module

If skywire is compiled with go 1.24, the version will be available from `"runtime/debug"` import:

```
fmt.Println(runtime.BuildInfo.Main.Version)

```

We should augment the way we version skywire and other related repos to use this new method, and drop the manual setting of the version via ldflags

We should upgrade the go version used in the CI and used to create releases to go 1.24

We must also update golangci-lint to the latest version & ensure it is built with the latest go version 1.24

0pcom:
https://github.com/skycoin/skywire/pull/1938

The test of this PR succeeds:

```
$ go run github.com/skycoin/skywire/cmd/version@develop
go: downloading github.com/skycoin/skywire v1.3.29-rc5.0.20250226185533-ac6ecde1de31
github.com/skycoin/skywire/cmd/version version v1.3.29-rc5.0.20250226185533-ac6ecde1de31
$ go run github.com/skycoin/skywire/cmd/version@develop
github.com/skycoin/skywire/cmd/version version v1.3.29-rc5.0.20250226185533-ac6ecde1de31

```

What is more convenient, the build is cached.

While it may take a moment to download the latest commit to develop, running the same command again works much faster and is basically seamless

## Further changes

We need to integrate this versioning scheme so that the other places in the source code which reference the version will use this & either maintain reverse compatibility with the previous versioning mechanisms or we need to state that go v1.24 is the minimum version required to build

## Future implications

The versioning was the one thing which previously required `git` as a dependency, and the user to clone the source code. With this change, it's possible to get a correctly versioned skywire binary with `go run` or `go install`

now it's possible to easily update a locally installed binary with those commands.

PR #1938, opened 2025-02-26
Add Version command

0pcom:
https://github.com/skycoin/skywire/issues/1934

This is difficult to test outside of the develop branch here because the module declares it's path differently from what it's required as on forks.

So the version command will be added, at least temporarily, until it's ombined with the buildinfo package - in a way which is hopefully reverse compatible with previous versioning approaches.

## Testing (after merge)

it should be possible to either install or run skywire from the develop branch and to run `skywire version` to print the version.

```
go install github.com/skycoin/skywire/cmd/version@develop
# OR
go run  github.com/skycoin/skywire/cmd/version@develop
```


0pcom:
* the golang & golangci-lint versions used by the CI were also updated in this PR

* replace directives were omitted from go.mod to allow the versioning scheme to work correctly via `go install`

* vestigial code was removed

* CI errors shown on the latest version of golangci-lint were fixed

Issue #1945, opened 2025-03-05
Use address resolver to determine if certain transport types can be established to remote visors


0pcom:
Currently there exists logic to prevent automatic stcpr transport creation to non-public visors - which may be overridden by the `-f` flag of `skywire cli tp add`

This logic is actually implemented incorrectly as it blocks creation of stcpr transports regardless of if the visor is public or not.

In the second place, it was the wrong choice to use the public visor service discovery to make this determination.

the [address resolver](https://ar.skywire.skycoin.com/transports) should be used to predetermine what transport types may be created to remote visors. including sudph. And we should show an error message based on if the visor attempting to be transported is shown to be available for certain transport types.

the -f flag should force a transport creation attempt for certain types of transports.

0pcom:
fixed in https://github.com/skycoin/skywire/pull/1947

PR #1947, opened 2025-03-10
`skywire cli tp add` use address resolver instead of public visor service discovery

0pcom:
Fixes: https://github.com/skycoin/skywire/issues/1945


0pcom:
@mrpalide integration test is broken for linux CI. it works for me locally. Merging.

PR #1948, opened 2025-03-12
Remove replace directive from go.mod

0pcom:
replace directives will break `go install` run from outside cloned sources. i.e. `go install github.com/skycoin/skywire/cmd/skywire@develop`


0pcom:
tests still broken @mrpalide - merging

PR #1953, opened 2025-03-21
transition reward system UI to use cogentcore.org/core [WIP]

0pcom:
use cogentcore.org/core for reward system UI - WORK IN PROGRESS
```
go run cmd/skywire/skywire.go cli rewards-ui -p 8082
```
image

```
 go run cmd/skywire-cli/commands/rewards-ui/ui/ui.go
```

image

* There is currently a bit of discrepancy between the behavior of the web and desktop versions of the UI, I'm working with upstream devs on these issues such as image display, etc.

I've devised a method for including the source files for the UI which are then compiled when the reward system server is run. This is efficient in that it does not bloat the size of the binary, however it presents a challenge in including different versions of the UI. For instance, we could make it possible for users to run the desktop or web UI locally, using dmsghttp to connect to the reward system server API over dmsg ; or we could include it as it's own visor app and use the visor's dmsg client.

These methods present a challenge to the inclusion of the UI source files because the UI is currently a main package. if it were a library, compiled on the user end instead of embedding the wasm into the source files, then it would be impossible to include development changes to that library at compile time, for testing.

So we may just be better off with embedding the wasm in the future, if we want users to be able to run this locally. It's not currently possible to include the UI as it's own subcommand of skywire-cli while maintaining embed-able source code files without duplicating the code.


0pcom:
https://github.com/user-attachments/assets/14ab298d-2bfe-446e-8b2f-7b66f27f17d7

still WIP

0pcom:
Tests are somehow broken by these changes.

```
Run make check
golangci-lint --version
golangci-lint has version 1.64.[5](https://github.com/skycoin/skywire/actions/runs/14379838813/job/40320768640?pr=1953#step:5:6) built with go1.24.0 from 0a[6](https://github.com/skycoin/skywire/actions/runs/14379838813/job/40320768640?pr=1953#step:5:7)03e49 on 2025-02-13T21:19:55Z
GO111MODULE=on golangci-lint run -c .golangci.yml skywire.go
GO111MODULE=on golangci-lint run -c .golangci.yml ./cmd/...
level=warning msg="[runner] Can't run linter goanalysis_metalinter: buildir: failed to load package glfw: could not load export data: no export data for \"github.com/go-gl/glfw/v3.3/glfw\""
level=error msg="Running error: can't run linter goanalysis_metalinter\nbuildir: failed to load package glfw: could not load export data: no export data for \"github.com/go-gl/glfw/v3.3/glfw\""
level=error msg="Timeout exceeded: try increasing it by passing --timeout option"
make: *** [Makefile:221: lint] Error 4
```

I'm trying to go ahead and deploy this new UI for the reward system. @mrpalide please fix the CI issue when you get a chance.

PR #1984, opened 2025-05-20
update deps & vendor latest commits from dmsg

0pcom:


0pcom:
@mrpalide i think it should be possible to update wireguard to the latest version, it will just require some changes

image

I'll downgrade it again to the last working commit, for now

PR #1989, opened 2025-05-29
Expand public autoconnect logic [WIP]

0pcom:
The current logic for the public autoconnect only makes STCPR transports to public visors.

While this should create enough transports to enable multihop routes, I believe that some improvement is possible with the following change:

* Visors should make SUDPH transports to those visors which are connected to public visors

The only issue with this proposal stems from the fact that: once a public visor reaches a certain number of transports, it should de-register itself as a public visor in the service discovery.

At the point where the public visor de-registers itself, there is not a good way to tell which public keys belong to public visors, except indirectly by number of transports, because the visor is no longer listed in the service discovery as a public visor.

Ideally we would have a similar setup as we have for dmsg servers listed in dmsg-discovery where we have `available` and `all` servers endpoints


0pcom:
I've significantly improved the logic for the public autoconnect.

* moved autoconnect logic out of `pkg/servicedisc` into `pkg/visor`
The autoconnect is really critical logic and probably should have been in pkg/visor to begin with. This change was necessary in order to use the different api clients such as ut client, ar client & tpdisc client

* maintain fresh data by limiting connection attempts to a 5 minute window.
This avoids the visor taking a long time to attempt to connect to many keys in a long list, where the data may be old

* autoconnect to public visors via stcpr AND visors which are already connected to public visors via sudph
This should significantly reduce the pressure on public visors

* randomize the list of visors to connect to (very important)
Avoid the situation where every visor is trying to connect to the same public visor at once!

* ensure that visors are online according to the uptime tracker before attempting to connect to them
This avoids attempting to connect to visors that have subsequently gone offline

* determine which transport type to use when connecting with address resolver client
This is important since the public autoconnect logic is now using multiple transport types.

* moved transport type constants to it's own package (from `pkg/transport/network` to `pkg/transport/types`)
There were lots of packages importing `pkg/transport/network` and only using the transport type constants, so this is a significant optimization

## NOTES

**The public autoconnect relies on at least one public visor running on the network**

No transports will be established to any visor unless a public visor is observed to be present on the network.

We should consider changing that - per the initial comment of this pull request - so that transports may be established to visors that are connected to unadvertised public visors.

PR #1992, opened 2025-05-31
Update vendor deps

0pcom:
```
go get -u ./...
```
* reset wireguard to previous version
```
go mod tidy && go mod vendor
```


0pcom:
I'm intending to add the dmsg servers config as `skywire dmsg conf` directly in the dmsg repo - and randomize the dmsg servers at startup so that dmsg utilities don't by default all attempt to connect via the same dmsg server initially.

* moved the skywire dmsghttp deployment config to `skywire svc conf dmsghttp`

skywire already has a mechanism for randomizing the dmsg servers used by the visor when it's run with dmsghttp-config, so it's unnecessary to do that here. But this is needed for dmsg utilities.

I wanted to update the deps here first, and will update the deps in dmsg as well, to include this commit in skywire and any updates to other code deps for dmsg.

PR #1996, opened 2025-06-05
add health check rpc method for route setup-node

0pcom:
WIP - add health check method to route setup-node rpc


0pcom:
tested, works. a cli command is included at `skywire svc sn health` which accepts the public key of the setup-node to dial to.

further diagnostic tooling such as checking dialing dmsg clients via specified servers will be added to dmsgweb to further aid troubleshooting of route setup-node dialing errors.

PR #1997, opened 2025-06-06
update vendor deps

0pcom:
update vendor deps


0pcom:
I've finally fixed the issue with wireguard for windows VPN.

Now the only dependency with a version that should not change is skycoin/skycoin - which uses the cobra-cli branch so that it can integrate as subcommand of skywire.

PR #2008, opened 2025-07-04
Remove default same-version filtering from proxy & vpn list subcommands

0pcom:
Remove default same-version filtering from proxy & vpn list subcommands

### DIFF
--- a/cmd/skywire-cli/commands/proxy/proxy.go
+++ b/cmd/skywire-cli/commands/proxy/proxy.go
@@ -7,7 +7,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
-	"strings"
 	"text/tabwriter"
 	"time"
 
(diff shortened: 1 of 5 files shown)


0pcom:
we still have issues with dead entries not being expired / deregistered in the service discovery


```
$ go run cmd/skywire/skywire.go cli proxy list --help
List proxy servers from service discovery
http://sd.skycoin.com/api/services?type=proxy
http://sd.skycoin.com/api/services?type=proxy&country=US

Set cache file location to "" to avoid using cache files



Flags:
  -m, --cfa int        update cache files if older than n minutes (default 5)
      --cfs string     SD cache file location (default "/tmp/proxysd.json")
      --cfu string     UT cache file location. (default "/tmp/ut.json")
  -o, --noton          do not filter by online status in UT
  -k, --pk string      check proxy service discovery for public key
  -r, --raw            print raw json data
  -a, --sdurl string   service discovery url (default "http://sd.skycoin.com")
  -s, --stats          return only a count of the results
  -w, --uturl string   uptime tracker url (default "http://ut.skywire.skycoin.com")


$ go run cmd/skywire/skywire.go cli proxy list -os
2800
$ go run cmd/skywire/skywire.go cli proxy list -s
1606

$ go run cmd/skywire/skywire.go cli vpn list -os
1635
$ go run cmd/skywire/skywire.go cli vpn list -s
896

```

0pcom:
I've updated the `ut` subcommand to correctly reflect the online status of visors

```
$ go run cmd/skywire/skywire.go cli ut -os
1811 visors online
$ go run cmd/skywire/skywire.go cli ut -s
2047 visors
```

only a little over 200 keys different, versus almost double for the service discovery entries where the visor providing the service is offline. Many of these are likely not even in the uptime tracker because they have been offline for so long.

PR #2067, opened 2025-10-05
[WIP] GeoIP Service

### LINKED ISSUE #2019: add source code for ip.skycoin.com & visor config field & update ip geolocation database
ip.skycoin.com is a somewhat critical yet often overlooked deployment service, however:

* it is hardcoded
* it is not open source
* the ip geolocation database is many years out of date

## Solutions

* make a field of the visor's config which allows ip.skycoin.com to be configured, instead of having it hardcoded in the places where it is hardcoded
 
*  add that configuration to the default embedded config & update conf service to include it as well

* add the source code for ip.skycoin.com ip geodata service to this repo

* update the ip geolocation database used by ip.skycoin.com service

## Potential Alternative

We do have a method for getting visor ip address from the dmsg server. We could expand that to include geolocation data, but it might require the ip database to be embedded or otherwise available for the dmsg server to look at. The database may be too large to conveniently embed. More investigation is warranted.

mrpalide:

Fixes #2019 

 Changes:	
- [x] implement GeoIP service, both by API and CLI
- [x] compatible with skywire as `./skywire svc ip`
- [x] add it on config (remove hardcoded)
- [x] add --geoip flag to service-discovery, uptime-tracker and `skywire visor ip` command 
- [x] deploy geoip on test network as http://ip.skywire.dev
- [x] deploy geoip on prod network on http://ip.skycoin.com  

How to test this PR:
- check http://ip.skywire.dev and http://ip.skywire.dev/?ip=x.x.x.x for functionality of geoip
- build skywire, generate config, change value on `geoip` field to `http://ip.skywire.dev`, run visor and check logs on register apps like skysocks for functionality of geoip
- check `skywire visor ip --geoip http://ip.skywire.dev` output
- check geoip service independetly by these two:
  - `skywire svc geoip <ip> --db ./GeoLite2-City.mmdb` and check output
  - `skywire sec geoid --api --db ./GeoLite2-City.mmdb` and check `localhost:8080/?ip=x.x.x.x`
  **note that, needs to download and put GeoLite2-City db beside skywire binary**


0pcom:
```
$ ./geoip --help
╔═╗ ┌─┐┌─┐╦   ╔═╗ 
║ ╦ ├┤ │ │║   ╠═╝ 
╚═╝ └─┘└─┘╩   ╩   

Note: GeoIP database should downloaded before start. You can get it from https://deb.skywire.dev/GeoLite2-City.mmdb
skywire svc geoip x.x.x.x --db ./GeoLite2-City.mmdb
skywire svc geoip --api --addr ":9093" --db ./GeoLite2-City.mmdb

Usage:
  geoip



Flags:
  -a, --addr string     address to bind to (default ":8080")
      --api             Run as API server
      --db string       Path to GeoLite2-City.mmdb database
  -l, --loglvl string   [info|error|warn|debug|trace|panic] (default "info")
      --tag string      logging tag (default "geoip")
  -v, --version         version for geoip

```

1) use all lowercase for the calvin ascii font
2) can or should we include (embed) the ip geolocation database instead of just linking it in the help like that? I think it would be better and save an extra step.

0pcom:
i just tested this and it works fine, by the way

mrpalide:
@0pcom 
1. lowercase geoip ascii handled
2. it is about 62mb db, and is large I think for repo, if you are ok, I can add it later.

0pcom:
> 2. it is about 62mb db, and is large I think for repo, if you are ok, I can add it here.

it should be ok,, I think.. the worst thing about it will be that it directly increases the size of the binary. But we already save a lot of space by compiling everything together.

I think we should go ahead with it for now. if it becomes problematic, we can remove it. Or if you have other strong sentiments against it, I'll defer to your judgement.

It would be somewhat useful, I think, if this database was also used by the dmsg server for the ip request. there is no reason why that should return different data than ip.skycoin.com.. but it's not exactly necessary or urgent to make that change, and I might have to modify the reward system to accommodate that change.

in fact the whole ip geolocation service could be included itself with the dmsg server, basically... hooked up to the ip method. But again, this is mostly future work. It just becomes convenient at the point that we have the database embedded.

0pcom:
also let us leave it such that the embedded database can still be overridden with path to database on the disk via flag. so that if no path is specified via that flag that it will use the embedded database.

PR #2079, opened 2025-10-27
Apps launcher revisions

0pcom:
* make the apps launcher able to start apps via function calls instead of system calls - maintaining reverse compatibility
* change default config to start apps via function calls instead of system calls.


mrpalide (approved):
Looks good. I test it too, but custom/unofficial external app test remains. We will do it when one app available later and if any issue be there then trying to solve it.

PR #2117, opened 2025-11-20
test improvements

0pcom:
continued


0pcom:
the e2e tests were reset to the commit when they had last passed but they are still failing. i need the fixes for the reward system server on develop right now ; will revisit the test failures in next PR

PR #2138, opened 2026-01-04
Fix release issues

0pcom:
* fix the CLOSE_WAIT open connection / resource leak / non-transportability issue
* fix apps logging messed up
* fix internal apps config causing visor crash


0pcom:
The majority of the issues have been resolved. There may still be an issue with the VPN client, but we need a patch release to replace v1.3.32. So the test failure will be resolved ASAP after the next release.

PR #2140, opened 2026-01-21
fix internal apps

0pcom:
- Update go dependencies to latest versions
- Fix log file permissions when visor runs as root in non-root owned directory (creates logs with 0644 instead of 0600 for readability)
- avoid `os.Exit()` in internal vpn client app
- fix multiple apps logging issue


0pcom:
Tests still not passing. I need to update a remote visor with these changes in order to continue testing. Merging...

PR #2145, opened 2026-01-25
Fix internal apps logging cross-contamination

0pcom:
Create isolated MasterLogger for each app in NewProcLogger instead of reusing the shared MasterLogger. When multiple apps start and share the same logger, each app's LogStore hook was being added to the shared logger, causing logs from one app to be written to all apps' log stores.


0pcom:
releases were building the wrong binary, the one without skycoin utilities - this was corrected to build releases from the repo root.

PR #2390, opened 2026-04-29
fix(dmsg-server): preload direct client with peer dmsg-server entries

0pcom:
## Summary

The dmsg-server's server-services dmsg client (which handles `/health`, `/pprof`, the embedded route-setup-node, and the DHT full node's network transport) was constructed with a `direct.NewClient` seeded with only the local server's own entry:

```go
entries := direct.GetAllEntries(cipher.PubKeys{conf.PubKey}, []*disc.Entry{serverEntry})
dClient := direct.NewClient(entries, log)
```

That keeps the dmsg-through-self services working without dmsg-discovery, but it also means the dmsg.Client cannot resolve any other PK. Every outbound dmsg dial from this process fails with `entry of public key is not found`.

The visible casualty is the DHT bootstrap. The dmsg-server runs a DHT full node on port 100 (when `enable_dht: true`) and uses this same dmsg.Client as its DHT network transport. Every bootstrap ping to a peer dmsg-server (the PKs from `deployment.Prod.DHTBootstrapPKs()` — i.e. all the *other* dmsg-servers) hits the resolution failure. After 3 failures the DHT caches the peer as non-DHT, so subsequent retries don't recover even after the failure mode is gone. The routing table on every dmsg-server stays empty and visor GetValue/GetItems traffic ends up served from each node's small in-memory cache — even though the per-salt entries are sitting in the shared Redis mirror that TPD/DMSGD/SD already populate.

Production state today (`139.162.160.227`, dmsg-server-6):

```
Redis dht:* keys (mirrored from disc): dmsg=1440, svc=1171, tp=1111
Local DHT store on dmsg-server-6:      ~16-32 items per salt
Bootstrap routing table:               0 peers (every ping cached as non-DHT)
```

## Fix

Mirror the `BootstrapDmsg` helper in `pkg/cmdutil/dmsg_bootstrap.go` that every other deployment service uses: seed the direct client with the embedded `dmsg.Prod.DmsgServers` list (excluding self) so peer PKs resolve without contacting dmsg-discovery. Direct-client architecture is preserved — no HTTP discovery dependency added.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./cmd/dmsg/dmsg-server/...`
- [x] `go test ./pkg/dmsg/direct/... ./pkg/dmsg/dmsgserver/... ./pkg/cmdutil/...` — all green
- [ ] Roll out the new image to the dmsg-server cluster; confirm:
  - Bootstrap ping logs change from `dmsg error 100 - entry is not found in discovery` to `DHT bootstrap succeeded`
  - `Network Size` on each DHT node grows past 0
  - Visor `dht get <pk> dmsg` / `dht get <pk> tp` start returning live data instead of falling through to HTTP

### DIFF
--- a/cmd/dmsg/dmsg-server/commands/start/root.go
+++ b/cmd/dmsg/dmsg-server/commands/start/root.go
@@ -162,7 +162,27 @@ var RootCmd = &cobra.Command{
 					AvailableSessions: conf.MaxSessions,
 				},
 			}
-			entries := direct.GetAllEntries(cipher.PubKeys{conf.PubKey}, []*disc.Entry{serverEntry})
+
+			// Build the direct client's static view: this server's own
+			// entry plus every peer dmsg-server from the embedded
+			// deployment list (excluding self). Without the peer entries
+			// the direct client could only resolve its own PK, so any
+			// outbound dmsg dial from this process — most importantly
+			// the DHT bootstrap pings and inter-server DHT replication
+			// over the dmsg DHT transport — failed with
+			// "entry of public key is not found", and every bootstrap
+			// peer wound up cached as non-DHT until restart. Mirrors
+			// the BootstrapDmsg helper that every other deployment
+			// service uses (pkg/cmdutil/dmsg_bootstrap.go).
+			servers := []*disc.Entry{serverEntry}
+			for i := range dmsg.Prod.DmsgServers {
+				peer := &dmsg.Prod.DmsgServers[i]
+				if peer.Static == conf.PubKey {
+					continue
+				}
+				servers = append(servers, peer)
+			}
+			entries := direct.GetAllEntries(cipher.PubKeys{conf.PubKey}, servers)
 			dClient := direct.NewClient(entries, log)
 
 			debugConfig := &dmsg.Config{
--- a/pkg/dht/tpd_adapter.go
+++ b/pkg/dht/tpd_adapter.go
@@ -29,27 +29,38 @@ func NewTPDAdapter(node *Node, log *logging.Logger) *TPDAdapter {
 }
 
 // RegisterTransports publishes the given transport entries to the DHT.
-// All entries are stored as a list under the node's own PK.
+// All entries are stored as a list of bare transport.Entry under the
+// node's own PK. The SignedEntry wrapping that legacy callers pass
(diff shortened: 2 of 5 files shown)


0pcom:
Pushed an additional commit on this branch (`5ba702d1b`) for three independent visor-side bugs that were also leaving DHT/TPD with state that didn't match local truth — complementary to the dmsg-server bootstrap fix here.

**Changes** (`pkg/transport/manager.go`, `pkg/dht/tpd_adapter.go`):

1. **Self-loop transports were leaking into discovery.** `saveTransportInternal` already swaps a no-op DiscoveryClient when `remote == self`, but `reRegisterTransports` iterates `tm.tps` and republishes every entry through the live `DiscoveryClient` — including self-loops. Now filtered on `Edges[0] == Edges[1]`.

2. **Empty publishes were skipped.** When the last real transport went away, `len(bareEntries) == 0` made `reRegisterTransports` early-return, so the DHT entry under the visor's PK kept serving the previous list forever. Drop the early-out: empty list with a fresh sequence is exactly the cleanup payload readers need.

3. **`TPDAdapter.RegisterTransportsV3` was producing zero-signature stubs.** It wrapped each `Entry` in `&SignedEntry{Entry: e, Version: v}` without filling `Signatures`, so every DHT tp publish landed with `"signatures":[…00..00, …00..00]`. Switched the DHT-side storage shape to bare `[]*Entry` — BEP44-style under-PK publication already authenticates the writer at the DHT layer, so per-entry signatures are redundant. `GetTransportsByEdge` prefers the bare shape and falls back to the legacy `SignedEntry` shape so peers that haven't republished since this commit still resolve. HTTP TPD path is unchanged.

Also drops dead code: `tm.dhtHandlesRegistration` / `SetDHTHandlesRegistration` were defined but never called — the skip branch they guarded was unreachable.

**Verified live** on a fixed visor after deploying:
- `DHT[tp, self]` only contains non-self-loop transports
- Each entry uses the new bare-Entry shape (no zero-sig artifacts)
- `Re-registering 0 transports` appears in the log on transport-free ticks, confirming the cleanup-publish path runs

Together with the dmsg-server bootstrap fix on this branch, the visor's data going INTO the DHT now matches local truth, and the dmsg-server cluster can actually bootstrap to receive and serve it.

PR #2529, opened 2026-05-13
  fix(visor): bound --dmsg-server startup + support pk@host:port form

### LINKED ISSUE #2528: fault functionality on hypervisor
1. When trying to stop visor with enabled hypervisor feature, Ctrl+C response as:
```
[2026-05-13T13:26:55.3252+03:30] WARN [dmsgctrl]: Failed to accept dmsgctrl conn, continuing: listening on closed connection
[2026-05-13T13:26:55.3252+03:30] WARN [dmsgctrl]: Failed to accept dmsgctrl conn, continuing: listening on closed connection
[2026-05-13T13:26:55.3252+03:30] WARN [dmsgctrl]: Failed to accept dmsgctrl conn, continuing: listening on closed connection
```
in tons of same line, and without shutting down complete, stop visor.

2. When trying to use `--dmsg-server` flag to connect to specific server, logically should regardless of what number of session_count value, consider it as 1 and just connect to that specific dmsg server, but now, seems that flag skipted at all scenarios.

mrpalide:

  Fixes #2528

  Changes:
  - Add `--dmsg-server-max-attempts` flag (default 5, hidden); only honored with `--dmsg-server`.
  - Track consecutive failed `Serve`-loop passes in `dmsg.Client` via atomic counter, exposed as `PinnedFailureCount()`. Increments on discovery error, pinned PK not in discovery, or
  `EnsureSession` failure. Resets on successful session.
  keeps the 30s warn-and-continue behavior.
  - With the dmsg client's 5s→60s backoff between passes, 5 attempts ≈ 2–3 min before shutdown — short enough to fail visibly, long enough to ride out transient blips.
  - Extend `--dmsg-server` to accept `pk@host:port` in addition to plain `pk`. With the `@` form, skip dmsg-discovery entirely and dial `host:port` directly. Useful for bootstrapping a
  visor that can't reach discovery yet, or for pinning a specific server address regardless of what discovery says. The PK is validated at flag parse; bad/empty form → fatal with a
  clear message. The failed-attempts cap still applies — TCP dial failures count as attempts.
  - `initDmsgHTTP` also honors the `@` form: synthesizes the pinned entry from the flag so dmsg-HTTP works on a fresh cache-less visor.
  - Includes earlier dirty-log + graceful-shutdown fix: run `Visor.Close()` on both init-failure paths in `NewVisor` so the dmsg `Serve` goroutine and listeners exit cleanly instead of
  leaking until the process dies.
  
  How to test this PR:
  - Start visor with `--dmsg-server <unreachable-pk>`; confirm it retries ~5 times and exits with `--dmsg-server pinned but unreachable; aborting startup`.
  - Start visor with `--dmsg-server <reachable-pk>`; confirm normal startup and `DMSG client connected and ready.` log.
  - Start visor without `--dmsg-server`; confirm the 30s warn-and-continue path still fires when dmsg is slow.
  - Tune `--dmsg-server-max-attempts=2` and confirm shutdown happens ~2× faster.
  - Start visor with `--dmsg-server <pk>@<reachable-host:port>`; confirm no discovery call is made for session setup (`--dmsg-server pk@host:port: skipping discovery for dmsg-http` in
  logs) and dmsg connects.
  - Start visor with `--dmsg-server <pk>@<unreachable-host:port>`; confirm TCP dial failures count as attempts and the visor shuts down after the cap is reached.
  - Start visor with `--dmsg-server pk@` or `--dmsg-server @host:port`; confirm fatal error at startup with a clear message.

### DIFF
--- a/pkg/app/appnet/skywire_networker.go
+++ b/pkg/app/appnet/skywire_networker.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/binary"
 	"errors"
+	"fmt"
 	"io"
 	"net"
 	"sync"
@@ -25,7 +26,10 @@ var (
 	// ErrPortAlreadyBound is being returned when the desired port is already bound to.
 	ErrPortAlreadyBound = errors.New("port already bound")
 	// ErrClosedConn is being returned when we are listening on a closed connection.
-	ErrClosedConn = errors.New("listening on closed connection")
+	// Wraps net.ErrClosed so generic accept loops that check the standard
+	// sentinel (e.g. dmsgctrl.ServeListener) recognize this as terminal
+	// instead of spin-logging on shutdown.
+	ErrClosedConn = fmt.Errorf("listening on closed connection: %w", net.ErrClosed)
 )
 
 // SkywireNetworker implements `Networker` for skynet.
--- a/pkg/dmsg/dmsg/client.go
+++ b/pkg/dmsg/dmsg/client.go
@@ -244,7 +244,29 @@ func (ce *Client) Serve(ctx context.Context) {
 		var err error
 		ce.log.Debug("Discovering dmsg servers...")
 		pinned := ctx.Value("dmsgServer") != nil
-		if pinned {
+		pinnedAddr, _ := ctx.Value("dmsgServerAddr").(string)
+		if pinned && pinnedAddr != "" {
+			// pk@host:port form: skip discovery entirely. We have
+			// everything dialSession needs (PK + TCP address) right
+			// here, so don't burn an HTTP round-trip on dmsg-discovery
+			// — that may be unreachable in the bootstrap scenarios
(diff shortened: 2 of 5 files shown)


0pcom:
If you are going to do `--dmsg-server` flag, can you make it take either `<pk>` OR `<pk>@<ip>:<port>` so that it can optionally avoid the discovery too? Might be useful. And probably could do that as an array comma separated... Hmm.

and is it already possible to supply basically `dmsghttp-config.json` override when starting the visor? via either ENV or flag? Might do that just to have a flexible implementation..

mrpalide:
@0pcom This is just an overwrite on config, without change it, like a emergency way to run visor.
But I like pk@ip:port format for bypass dmsg-discovery step. will apply it.

0pcom:
Hey, I replicated your changes on another PR & finished it. Let me know if you still have issues.

Issue #2607, opened 2026-05-14
RFC: faithful UDP-over-skynet (DatagramPacket + PacketConn-shaped RouteGroup variant)


0pcom:
## Background

`sudph` (UDP-with-KCP-on-top) gives skywire a UDP **underlay** for its TCP-shaped routed packets, but no UDP **datagram semantics** at the app layer. `RouteGroup` is `net.Conn`-shaped (`Read`/`Write`, sequence numbers, reordering, gap detection via SACK, head-of-line blocking on loss). Apps that want real UDP — VoIP, gaming, WebRTC media, anything where a late packet is worse than a lost one — can't get it.

This issue covers the faithful-UDP path (shape A). Sibling effort: 027087fe40 is taking `cmd/skyudp-bridge` (shape B — length-prefixed datagrams over a TCP stream) which covers query-response UDP cases. The two paths complement each other.

## When to use which (picking rubric)

| Use case | Choice | Why |
|---|---|---|
| DNS, NTP, SNMP, MQTT-SN | **B** | Request/response; reliable+ordered delivery is a strict win, no semantics lost |
| WireGuard handshake / control plane | **B** | WG retransmits at its own layer; reliable underlay is wasted but harmless |
| SIP signalling | **B** | Request/response shape |
| VoIP media (RTP) | **A** | Head-of-line blocking from B would stall the audio on every retransmit |
| WebRTC media | **A** | Jitter buffer expects faithful UDP; B's delivery semantics break it |
| Online game state | **A** | Late packets are worse than lost packets |
| Live video / multicast media | **A** | Same as VoIP |
| Default for "I want UDP-over-skywire" | **B first** | Ship in days; reach for A only when HoL-blocking actually matters |

## Hot path today

Send: `RouteGroup.Write` → `MakeSequencedDataPacket(routeID, seq, payload)` → noise-encrypted at the route-group level → managed-transport `WritePacket`.

Receive: `handleDataPacket` → `rg.mux.deliverData(seq, data)` → reorder buffer → `readCh` → `Read`.

Two facts that block UDP semantics:

1. Every `DataPacket` carries a sequence number; the receiver buffers + reorders + fires SACK on gap. HoL-blocks on loss.
2. Noise is wrapped around the **stream** — a single dropped packet desyncs the keystream; all subsequent decryptions fail.

### Stage 1 — `DatagramPacket` (new wire type)

Add to `pkg/routing/packet.go`:

```go
[...]

PR #2610, opened 2026-05-15
feat(router): DatagramRouteGroup + Group interface (faithful UDP, stage 2)

0pcom:
Stage 2 of #2607. Builds on #2609 (Stage 1: DatagramPacket wire type).

> [!NOTE]
> Stacked on #2609. If #2609 merges first this PR will auto-rebase; otherwise the commit it depends on is `530293911` (introducing `DatagramPacket` + `MakeDatagramPacket`).

## What this adds

A `PacketConn`-shaped sibling to `RouteGroup` that carries `DatagramPacket`s without sequence numbers, reorder buffering, or SACK retransmit. Loss is loss, ordering is best-effort, head-of-line blocking does not occur on the route group's read path — UDP semantics, not the TCP-shaped delivery `RouteGroup` provides.

### `Group` interface

Narrow shared interface for every route-group variant — `Close` / `LocalAddr` / `RemoteAddr` / `IsAlive`. `Read`/`Write` vs `ReadFrom`/`WriteTo` deliberately stay on the concrete types so call sites that only care about lifecycle don't have to discriminate. Compile-time assertion in `datagram_route_group.go`.

### `DatagramRouteGroup`

- Single-peer (one remote per group, mirroring how `RouteGroup` is single-peer)
- Owns its own `readCh` — **bounded**, and a full queue **drops** the incoming datagram (faithful UDP semantics for a slow consumer; the `InboundDropped()` counter surfaces this for diagnostics)
- `WriteTo` wraps payload in `MakeDatagramPacket` and writes through the first available transport
- `AddRule(fwd, tp, rv)` populates the rule/transport triple. Same indexing invariant as `RouteGroup` so the setup-node plumbing in Stage 4 builds either type from the same code path.
- `MaxPayload()` returns 65535 in Stage 2; Stage 3 reduces it by the per-datagram AEAD overhead (8-byte nonce counter + 16-byte Poly1305 tag = 24 bytes → 65511) once crypto lands.

## What is NOT in this stage

- No **router-side dispatch** yet. `handleDatagramPacket` on the router lands in Stage 4 with `forwarded_ports.udp`; at this stage `Handle` is the unit-test entry point and no packet in the running system actually reaches a `DatagramRouteGroup`.
- No **AEAD wrapping**. Stage 3 adds per-datagram ChaCha20-Poly1305-IETF + sliding-window anti-replay. `MaxPayload()` will be reduced accordingly at that point.
- No **multi-leg / multiplex**. A `DatagramRouteGroup` writes through `tps[0]` only. Path multiplexing is not motivated by Stage 4's `forwarded_ports.udp` use case.

## Test plan

- [x] `go test ./pkg/router/` — pass (12 new tests + existing suite)
- [x] `go test -race ./pkg/router/` — clean, no data races
- [x] Compile-time assertions confirm `Group` + `net.PacketConn` satisfied

Test coverage:

- `HandleDeliversToReadFrom` — happy path: `Handle` → `ReadFrom`
- `HandleRejectsNonDatagramPacket` — defensive: a misrouted `DataPacket` does NOT silently slip into the datagram queue
- `ReadFromBlocksUntilDeadline` — deadline plumbing
- `CloseUnblocksReadFrom` — in-flight reads return `ErrClosed` on `Close`
- `CloseIsIdempotent`
- `WriteToTooBigRejected` — `ErrPayloadTooBig` instead of opaque EMSGSIZE
- `WriteToWithNoRulesFails` — guard against using a half-built group
- `WriteToWrongAddrRejected` — single-peer enforcement
- `InboundDroppedWhenReadChFull` — slow-consumer semantics + atomic counter
- `HandleAfterCloseReturnsClosedPipe`
- `IsAlive` — lifecycle + failure-count thresholding
- `SatisfiesPacketConn` — interface-shape assertion in test output

## Coordination

- Stage 1 (#2609) — wire type, dependency.
- Stage 3 — AEAD layer. Held *softly* for ersonp's eyes per Alpha's review of the RFC; can write it after Stage 2 lands.
- Stage 4 — `forwarded_ports.udp` integration. Mirrors the existing TCP forwarded-port plumbing.
- Stage 5 — `appnet.DialPacket` + `MaxPayload()`-exposing app API.

Closes part of #2607.

### DIFF
--- a/pkg/router/datagram_route_group.go
+++ b/pkg/router/datagram_route_group.go
@@ -89,8 +89,8 @@ var ErrClosed = errors.New("datagram-route-group: closed")
 //   - Close releases the readCh and marks the group dead. Idempotent.
 type DatagramRouteGroup struct {
 	// atomic; first for 64-bit alignment.
-	lastSent     int64
-	closeFlag    int32 // atomic; 1 once Close has run
+	lastSent                 int64
+	closeFlag                int32 // atomic; 1 once Close has run
 	consecutiveWriteFailures int32
 
 	mu sync.Mutex
--- a/pkg/router/datagram_route_group.go
+++ b/pkg/router/datagram_route_group.go
@@ -1,0 +1,381 @@
+// Package router pkg/router/datagram_route_group.go: faithful-UDP-
+// over-skynet sibling to RouteGroup. Stage 2 of #2607.
+//
+// While RouteGroup is net.Conn-shaped (Read/Write, sequenced
+// DataPackets, mux + reorder buffer + SACK retransmit, head-of-line
+// blocking on loss), DatagramRouteGroup is net.PacketConn-shaped
+// (ReadFrom/WriteTo, unsequenced DatagramPackets, loss-tolerant,
+// possibly-unordered delivery). Used by Stage 4's
+// forwarded_ports.udp path to expose UDP datagram semantics over
+// routed skynet hops — for VoIP, gaming, WebRTC media, anything
+// where a late packet is worse than a lost one.
+//
+// Stage 2 scope: type definition, send/receive plumbing,
+// net.PacketConn methods, lifecycle. NO crypto wrapping yet
+// (Stage 3 will add per-datagram ChaCha20-Poly1305-IETF + sliding-
+// window anti-replay). NO router-side dispatch yet (Stage 4 will
+// wire forwarded_ports.udp). At this stage DatagramRouteGroup is
+// self-contained and unit-testable, but no packet in the running
+// system will reach it.
+
+package router
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"math"
+	"net"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/skycoin/skywire/pkg/logging"
+	"github.com/skycoin/skywire/pkg/routing"
+	"github.com/skycoin/skywire/pkg/transport"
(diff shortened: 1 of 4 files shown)


0pcom:
(GitHub doesn't let me formally approve my own PR; treat this as endorsement)

Implementation matches the design from #2607. Spot-checks:

- **`MaxPayload()` is forward-aware** — returns 65535 in Stage 2, Stage 3 will reduce by 24 bytes (8 nonce-counter + 16 Poly1305 tag); comment + commit message both call that out so Stage 3 inherits the API contract.
- **Read-channel drop-on-full** is the right semantics for UDP. The `inboundDropped` counter via Stats() gives a slow consumer a diagnosable signal instead of silent loss.
- **Single-peer addr check in WriteTo** matches the Stage 2 scope; multiplexing is rightly deferred.
- **Narrow `Group` interface** keeps Read/Write vs ReadFrom/WriteTo on the concrete types — no stream-vs-datagram leakage into shared call sites.
- **Idempotent Close** via `sync.Once` + `closed` chan; `closeFlag` atomic for fast `IsAlive` path.

One micro-nit (not blocking): `consecutiveWriteFailures` field is declared on the struct but I don't see it incremented/read in the diff window I scanned. If it's plumbed in a hunk further down, ignore; otherwise either wire it in this PR or strip it so we don't ship a dead field. Quick grep can confirm.

LGTM; ship Stage 3 next per your sequencing. Looking forward to the per-datagram AEAD + sliding-window code.

PR #2612, opened 2026-05-15
feat(router): per-datagram AEAD for DatagramRouteGroup (faithful UDP, stage 3)

0pcom:
Stage 3 of #2607. Stacked on #2610 (Stage 2: DatagramRouteGroup + Group interface).

> [!NOTE]
> Stacked on #2610 which is stacked on #2609. The full chain merges as 1 → 2 → 3.

## What this adds

Each datagram is independently encrypted, authenticated, and replay-protected — the cipher survives the loss / reorder / HoL-free delivery semantics that faithful-UDP-over-skynet requires. Stream Noise (what `RouteGroup` uses today) won't work here: its continuous keystream desyncs on a lost frame, after which every subsequent decrypt fails.

## Cryptographic shape (per #2607 + Alpha's review)

- **Cipher**: ChaCha20-Poly1305-IETF (RFC 7539). Same primitive Noise uses; reusing keeps the dependency surface flat. 256-bit key, 96-bit nonce, 128-bit Poly1305 tag.
- **Master key**: HKDF-SHA256 from the Noise session master with role-tagged info (`skywire-datagram-v1-init` vs `-resp`).
- **Nonce**: 96 bits = zero-padding(32) || counter(64). 64-bit monotonic per-sender counter; on the wire only the 8-byte counter is transmitted (saves 4 bytes/pkt vs full nonce). Matches WireGuard.
- **AAD**: `routeID(4) || destPK(33)`. Cross-route substitution and cross-receiver replay both blocked.
- **Anti-replay**: RFC 6479 sliding window, 2048-bit. Counter IS the sequence — no redundant `seq` on the wire. Window commit happens **after** AEAD verify (a forged datagram cannot poison the window).
- **Rekey**: time-based (default 2 min) or packet-count-based (default 2^32). Well below the 2^60 ChaCha bound. `NeedsRekey()` is polled; `Rekey(newMaster)` swaps the AEAD epoch atomically.

## Wire format of one sealed datagram body

```
[counter (8 bytes BE uint64)] [ciphertext + 16-byte Poly1305 tag]
```

`MaxDatagramPlaintext = 65511` (MaxDatagramPayload - DatagramOverhead). `DatagramRouteGroup.MaxPayload()` now reflects this when a cipher is installed.

## Integration

`DatagramRouteGroup.SetCiphers(out, in)` installs the per-direction ciphers. `WriteTo` seals before `MakeDatagramPacket`; `Handle` opens after unwrap. **Auth failures drop silently** (returning to the sender leaks timing signal) but bump `AEADAuthFailures()` for operator diagnostics — a rising counter indicates either a forgery attempt in progress or a peer-side cipher desync.

Bypass mode (nil ciphers) is retained for Stage 2 unit tests; **production paths in Stage 4 will require non-nil ciphers** before any traffic flows.

## Test plan

- [x] `go build ./...` clean
- [x] `go test ./pkg/router/` — all green (13 new crypto tests + 2 new integration tests + Stage 2's 12 + existing suite)
- [x] `go test -race ./pkg/router/` — clean, no data races

Crypto layer tests (13):
- `RoundTrip`, `WrongRouteIDRejected`, `WrongBindPKRejected`, `ReplayRejected`, `OutOfWindowRejected`, `CounterMonotonic`, `ReorderedDeliveryAccepted`, `CounterExhaustedTriggersRekey`, `NeedsRekeyOnTime`, `RekeyResetsState`, `ShortCiphertextRejected`, `PlaintextTooLongRejected`, `ConcurrentSealOpenSafe`

Integration tests (2 new on top of Stage 2):
- `WithCiphersRoundTrip` — install ciphers, assert `MaxPayload` reflects AEAD overhead, round-trip a sealed packet
- `HandleAEADFailureDropsSilently` — forged sealed payload bumps `AEADAuthFailures` and does NOT deliver to `ReadFrom`; no error surfaced

## Coordination

- Stage 1 (#2609) and Stage 2 (#2610) are the prerequisite stack.
- Stage 4 will wire `forwarded_ports.udp` — uses `SetCiphers` to install ciphers from the Noise handshake.
- Stage 5 — app API.

@ersonp — the crypto stage was held softly for your eyes per Alpha's review of the RFC. If you have feedback, fire away. Otherwise the choices are: ChaCha20-Poly1305-IETF cipher, HKDF-SHA256 key derivation, 64-bit counter on-wire, AAD = routeID||destPK, 2048-bit sliding-window anti-replay, rekey on time-or-packet-count.

Closes part of #2607.

### DIFF
--- a/pkg/router/datagram_crypto.go
+++ b/pkg/router/datagram_crypto.go
@@ -156,10 +156,10 @@ const (
 //
 // Single-direction; thread-safe.
 type DatagramCipher struct {
-	role     Role
-	destPK   cipher.PubKey // for AAD binding on outbound; sender's PK on inbound
-	aead     atomic.Pointer[cipherEpoch]
-	rekeyMu  sync.Mutex // serializes Rekey calls
+	role    Role
+	destPK  cipher.PubKey // for AAD binding on outbound; sender's PK on inbound
+	aead    atomic.Pointer[cipherEpoch]
+	rekeyMu sync.Mutex // serializes Rekey calls
 
 	// Outbound-only state. The atomic counter is the sender's
 	// next nonce; rekey resets it. Tracked separately from inbound
@@ -179,13 +179,13 @@ type DatagramCipher struct {
 // count thresholds for the current rekey epoch. Replaced atomically
 // on Rekey so in-flight Seal/Open calls observe a consistent epoch.
 type cipherEpoch struct {
-	aead         interface { // chacha20poly1305.New returns this shape
+	aead interface { // chacha20poly1305.New returns this shape
 		Seal(dst, nonce, plaintext, additionalData []byte) []byte
 		Open(dst, nonce, ciphertext, additionalData []byte) ([]byte, error)
 	}
-	openedAt     time.Time
-	timeLimit    time.Duration
-	packetLimit  uint64
+	openedAt    time.Time
+	timeLimit   time.Duration
+	packetLimit uint64
 }
 
 // NewDatagramCipher constructs an outbound-or-inbound cipher from a
--- a/pkg/router/datagram_crypto.go
+++ b/pkg/router/datagram_crypto.go
@@ -1,0 +1,459 @@
+// Package router pkg/router/datagram_crypto.go: per-datagram AEAD
+// for DatagramRouteGroup. Stage 3 of #2607.
+//
+// Stream Noise (what RouteGroup uses) generates a continuous
+// keystream that desyncs on lost frames — every subsequent
+// decryption fails. UDP drops packets by design, so the datagram
+// path needs a cipher that survives loss: each datagram independently
+// encrypted, authenticated, replay-protected, AEAD-bound to its
+// route + receiver.
+//
+// Per-datagram AEAD model (matches WireGuard / IPsec ESP / DTLS 1.3
+// data-plane):
+//
+//   - Cipher: ChaCha20-Poly1305-IETF (RFC 7539). Same primitive the
+//     existing skywire crypto stack uses for Noise; reusing keeps
+//     the dependency surface flat. 256-bit key, 96-bit nonce,
+//     128-bit auth tag.
+//
+//   - Master key: derived from the Noise session master via HKDF-SHA256.
+//     One subkey per direction (initiator→responder vs the reverse)
+//     and per route within the session.
+//
+//   - Nonce: 96 bits, constructed as zero-padding(32) || counter(64).
+//     Counter is 64-bit, monotonic per-sender, starts at 0 at
+//     handshake. On the wire only the 8-byte counter portion is
+//     transmitted; the receiver reconstructs the full nonce from
+//     the implicit zero-padding. Saves 4 bytes/datagram vs sending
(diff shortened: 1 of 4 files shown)


0pcom:
(GitHub won't let me formally approve own PR; endorsement.)

Stage 3 lands every RFC-#2607 review point I asked for, verbatim:

- **ChaCha20-Poly1305-IETF** (RFC 7539, reuses existing crypto stack — flat dep surface).
- **64-bit counter on-wire** as zero-padding(32) || counter(64); only the 8-byte counter portion transmitted. Receiver reconstructs full nonce. WireGuard layout precedent.
- **AAD = routeID(4) || destPK(33)** binds ciphertext to (a) this route within the session and (b) this specific receiver. Both cross-route substitution and cross-receiver replay closed.
- **RFC 6479 sliding window, 2048-bit (32 × uint64)** for anti-replay. Comment correctly states 'the counter IS the sequence' — no redundant seq on the wire.
- **Rekey on either 2 min elapsed OR 2^32 packets**, comfortable margin under the 2^60 ChaCha bound. Reset counter to 0 on both sides under new master.
- **MaxDatagramPlaintext = 65511** (65535 - 8 counter - 16 tag = 65511). Matches the contract Stage 2's MaxPayload comment pre-pinned. Apps that consulted it once won't have to consult it again.

Tests: 13 crypto + 2 integration is the right shape — unit tests covering keystream / AAD-binding / replay-window edge cases, integration tests covering the actual encrypt-send / receive-decrypt path.

Also re: `consecutiveWriteFailures` nit from #2610 — Gamma confirms it's wired in WriteTo (+1 on fail, =0 on success) + IsAlive (compared against maxConsecutiveWriteFailures), 4 references in datagram_route_group.go. My grep window in the review missed it. Withdrawn.

Ship it. Stage 4 (forwarded_ports.udp) next per your sequence.

pkg/router/datagram_crypto.go
```diff
@@ -0,0 +1,459 @@
...
+			shiftLeft(&dc.inWindow, uint(shift))
+		}
+		dc.inTop = counter
+		// Mark the newly-arrived top bit.
+		bit := uint64(1) << ((replayWindowSize - 1) % 64)
+		dc.inWindow[(replayWindowSize-1)/64] &^= bit // clear any stale
+		// Actually we want to set the bit at the position of the
+		// new top, which after shift is at the most-significant
+		// edge.
+		dc.inWindow[replayWindowWords-1] |= 1
```
ersonp, line 412: ## Bug — replay window marks wrong bit
`dc.inWindow[replayWindowWords-1] |= 1` sets bit 0,
but the precheck/commit mapping puts `inTop` at bit 63 of word 31 (MSB). The
`bit` variable (`1 << 63`) is computed correctly two lines above but never used
for the mark. Fix: `dc.inWindow[replayWindowWords-1] |= uint64(1) << 63`.
Effect without fix: the most-recently-received counter is never marked seen, so
an immediate replay of it passes the window and gets delivered twice.

ersonp:
@0pcom 
Reviewed. Crypto choices look good.
I found one potential issue, please look into it. 
P.S. the issue was found by Claude.

PR #2613, opened 2026-05-15
feat(visor): UDP-over-skynet via forwarded_ports.udp (faithful UDP, stage 4)

0pcom:
Stage 4 of #2607. Stacked on #2612 (Stage 3: per-datagram AEAD).

> [!NOTE]
> Full stack: #2609 → #2610 → #2612 → this. Merges in order.

## What this adds

The user-facing "port-forward UDP over skynet" capability. With this stage merged, the operator writes:

```json
"forwarded_ports": {
  "53": {
    "local_port": 53,
    "udp": true,
    "skynet": true,
    "whitelist": ["<peer-pk-1>", ...]
  }
}
```

— the visor binds a local UDP socket at port 53, and a `UDPBridge` pumps datagrams to/from a peer-side `DatagramRouteGroup` (which in turn carries them as routed `DatagramPacket`s, AEAD-sealed per Stage 3). Faithful UDP semantics: loss tolerant, possibly-unordered, no head-of-line blocking.

### Changes

- **`pkg/visor/forwarded_ports.go`** — adds `UDP bool` field to `ForwardedPort`. Falsy keeps existing TCP semantics.
- **`pkg/visor/udp_bridge.go`** (new) — `UDPBridge` bidirectionally pumps datagrams between two `net.PacketConn` endpoints. Shape-agnostic; accepts any `net.PacketConn` for the skynet side (`DatagramRouteGroup` satisfies this) and any `net.PacketConn` for the local side (typical: `*net.UDPConn`). Two pump goroutines. Single-peer per bridge; the listener loop creates one bridge per `(local-app-addr, peer-pk)` pair when multi-peer fan-out is needed.

Backpressure handled by the underlying conns' own queues. Bridge does **not** buffer internally beyond a single-read scratch buffer — matches what a directly-bound UDP server experiences under load (kernel drops, no internal queueing).

## What is NOT in this stage

- The visor-side **listener loop** that constructs `UDPBridge`s from `ForwardedPort` entries with `udp: true`. That belongs with the existing forwarded-port plumbing and ties into skynet/DMSG service registration; deferred so this PR stays focused on the bridge type itself.
- Stage 5's `appnet.DialPacket` API + the route-setup glue that constructs `DatagramRouteGroup`s from a Noise handshake. Until that lands, the only consumer of `UDPBridge` is Stage 5's forward-listener integration.

## Test plan

- [x] `go test ./pkg/visor/` — pass (4 new bridge tests + existing suite)
- [x] `go test -race ./pkg/visor/` — clean, no data races

Bridge tests:
- `LocalToSkynet` — real `net.UDPConn` + piped PacketConn. External UDP sender writes a datagram, asserts it lands on the skynet side via `ReadFrom`.
- `SkynetToLocal` — reverse direction. Primer datagram registers the local peer; then skynet→local datagram reaches the original sender's UDP address.
- `CloseTearsDownBothEnds` — `Close` cancels both pump goroutines AND closes the underlying conns.
- `CloseIsIdempotent`.

## Coordination

- Stages 1-3 are the prerequisite stack. Stage 5 (app API) builds on this.
- The visor listener loop that ties `forwarded_ports.udp` to `UDPBridge` instances is a small follow-up after Stage 5 lands the `DialPacket` route-setup plumbing.

Closes part of #2607.

### DIFF
--- a/pkg/router/datagram_crypto.go
+++ b/pkg/router/datagram_crypto.go
@@ -156,10 +156,10 @@ const (
 //
 // Single-direction; thread-safe.
 type DatagramCipher struct {
-	role     Role
-	destPK   cipher.PubKey // for AAD binding on outbound; sender's PK on inbound
-	aead     atomic.Pointer[cipherEpoch]
-	rekeyMu  sync.Mutex // serializes Rekey calls
+	role    Role
+	destPK  cipher.PubKey // for AAD binding on outbound; sender's PK on inbound
+	aead    atomic.Pointer[cipherEpoch]
+	rekeyMu sync.Mutex // serializes Rekey calls
 
 	// Outbound-only state. The atomic counter is the sender's
 	// next nonce; rekey resets it. Tracked separately from inbound
@@ -179,13 +179,13 @@ type DatagramCipher struct {
 // count thresholds for the current rekey epoch. Replaced atomically
 // on Rekey so in-flight Seal/Open calls observe a consistent epoch.
 type cipherEpoch struct {
-	aead         interface { // chacha20poly1305.New returns this shape
+	aead interface { // chacha20poly1305.New returns this shape
 		Seal(dst, nonce, plaintext, additionalData []byte) []byte
 		Open(dst, nonce, ciphertext, additionalData []byte) ([]byte, error)
 	}
-	openedAt     time.Time
-	timeLimit    time.Duration
-	packetLimit  uint64
+	openedAt    time.Time
+	timeLimit   time.Duration
+	packetLimit uint64
 }
 
 // NewDatagramCipher constructs an outbound-or-inbound cipher from a
--- a/pkg/router/datagram_crypto.go
+++ b/pkg/router/datagram_crypto.go
@@ -1,0 +1,459 @@
+// Package router pkg/router/datagram_crypto.go: per-datagram AEAD
+// for DatagramRouteGroup. Stage 3 of #2607.
+//
+// Stream Noise (what RouteGroup uses) generates a continuous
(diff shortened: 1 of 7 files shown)


0pcom:
(GitHub won't let me formally approve own PR; endorsement.)

UDPBridge implementation matches the design contract from the RFC:

- **Single-peer per bridge** with explicit comment that multi-peer fan-out is the listener-loop's job (keyed by local source addr). Right shape — keeps the bridge type narrow and the multi-peer logic factored into one place.
- **Bidirectional pump, two goroutines** — `localToSkynet` + `skynetToLocal` — both running until either end closes; either side's close cancels the other via the bridge's context. Idiomatic Go pattern, clean error surfaces.
- **Backpressure via net's own queues** — correct UDP semantics. Kernel drops on full recv buffer, exactly what a directly-bound UDP server would experience. The comment calls this out so a future maintainer doesn't add an internal buffer 'for safety' and accidentally re-introduce head-of-line latency.
- **localPeer capture** on first ReadFrom + mutex — single-app-per-bridge contract enforced; listener creates one bridge per (local-app-addr, peer-pk) pair as documented.
- **Shape-agnostic over net.PacketConn** on both sides — DatagramRouteGroup satisfies it from Stage 2, *net.UDPConn satisfies it on the local side, no special-cased types in the pump path.

`forwarded_ports.udp` config flag is the operator-facing surface for this; adding to the existing ForwardedPort struct is the right minimal change.

Ship it. Stage 5 (app API + listener loop) is the last piece per your sequencing; once that lands the whole faithful-UDP path is operator-callable end-to-end.

PR #2625, opened 2026-05-16
fix(cxo/node): evict stale Conn on peer rejoin instead of rejecting

0pcom:
## Summary
- `acceptHandshake` rejected every incoming SYN whose `NodeID` matched a `pkToConn` entry with `ErrAlreadyHaveConnection` — including SYNs from peers whose prior session is in fact gone. On their side this surfaces as `factory.Connection closed` and they retry forever with the same outcome, never recover.
- Same shape on the outgoing side (`performHandshake`'s `Ack` path).
- On duplicate-peer detection, we now evict the existing `Conn` via `Close`, wait briefly for `run()` to drain (`removeConn` clears the slot), then proceed. The new connection is authoritative — the old record only persisted because `run()` never observed the disconnect (half-dead transport, wedged `receiveMsg`, or the peer's network/visor reset without TCP RST reaching us).

## Why
Hit in production tonight on a skychat group: peer's `subscriber_alive` went `false` ~17min after a transport blip, then 5 sequential `cli skychat group leave` + `join` cycles all failed identically with `'group: Join: connect: group: Connect: handshake failed: factory.Connection closed'`. 1:1 chat round-trips kept working (visor + chat-app healthy), publisher dmsg port was probe-reachable, owner's runtime log had no `received` events from that peer — pinpointing the failure to the CXO factory accept side rejecting the fresh `SYN`.

## Test plan
- [x] `go test ./pkg/cxo/node/... -count=1` passes
- [x] `go vet ./pkg/cxo/node/...` clean
- [ ] In-production verification: install built binary on owner host, observe that the previously-stuck peer's next group rejoin succeeds (group `subscriber_alive=true` on their side, fresh probe round-trip)

## Follow-ups (not in this PR)
- Cross-visor admin/roster gossip RFC (owner-SPOF for state mutations)
- History replay on reconnect (subscriber backfill from publisher's TreeStore)
- Liveness telemetry surface (`subscriber_alive` per-peer in `group info`)
- Multi-transport publisher (skynet fallback when dmsg congests)
- Mixed-binary-version detection / CI integration test

### DIFF
--- a/pkg/cxo/node/handshake.go
+++ b/pkg/cxo/node/handshake.go
@@ -68,9 +68,14 @@ func (c *Conn) performHandshake() error {
 
 		switch x := m.(type) {
 		case *msg.Ack:
-			// Check if node already has connection with peer.
-			if _, ok := c.n.hasPeer(x.NodeID); ok {
-				return ErrAlreadyHaveConnection
+			// A stale Conn for this peer (whose run() loop wedged or whose
+			// receiveMsg is blocked on a half-dead transport) silently
+			// rejects every rejoin attempt with ErrAlreadyHaveConnection.
+			// On a fresh SYN from the same peer, evict the prior record:
+			// the new conn is the source of truth and the old will tear
+			// itself down via Close → receiveMsg unblock → removeConn.
+			if existing, ok := c.n.hasPeer(x.NodeID); ok {
+				c.n.evictStalePeer(existing, x.NodeID)
 			}
 			c.peerID = x.NodeID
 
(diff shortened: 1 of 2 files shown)


0pcom (commented):
Surface-level LGTM — fix targets exactly the right symptom (production-confirmed via Gamma rejoin).

One race I'd want flagged before merge: `evictStalePeer` waits up to ResponseTimeout (5s fallback) for the old Conn to drain, then proceeds. In the **overflow path** (timeout fires before old `run()` exits), here is the sequence:

1. evictStalePeer returns after timeout, old `removeConn` hasn't fired yet.
2. Handshake completes, new Conn registers via `onConnInit` → `pkToConn[id] = new`.
3. Old `run()` eventually exits — possibly much later — and calls `removeConn(old)`.
4. `removeConn` at `pkg/cxo/node/node.go:572-581` does:
   ```go
   delete(n.addrToConn, c.Address())
   delete(n.pkToConn, c.peerID)
   ```
   No identity check. The slot gets wiped even though it points at the *new* Conn.
5. From this point, any `hasPeer(id)` check returns false, even though the new session is alive and messaging. The next rejoin from this peer hits the no-op branch (no stale evict needed), is accepted normally, but until then any other consumer of `hasPeer` (e.g. duplicate-detect, feed routing) sees the connected peer as unknown.

Fix is one line in `removeConn`:

```go
if existing, ok := n.pkToConn[c.peerID]; ok && existing == c {
    delete(n.pkToConn, c.peerID)
}
if existing, ok := n.addrToConn[c.Address()]; ok && existing == c {
    delete(n.addrToConn, c.Address())
}
```

The overflow path is the unlikely case (5s should normally be enough), but production has surprised us before on exactly this kind of half-dead-transport stall.

Apart from that, the comment block is great, the surface is bounded to the handshake codepath, and the test plan covers the in-prod verification path that already paid off tonight. Happy to test this myself once it's on `/usr/bin/skywire` here — I have a working build of `a228fbce9` ready to install.

0pcom:
Thanks @Beta-agent — exactly right. Applied as `b0e67ebcb`. `removeConn` now checks slot identity for both `pkToConn` and `addrToConn` before deleting, so the late `removeConn` of an evicted Conn becomes a no-op when the slot has moved on to a new live Conn.

Build + tests still green. The fix is now stacked on a228fbce9 in this branch.

PR #2627, opened 2026-05-16
fix(skychat/group): per-(group, peer) reconnect backoff

### LINKED ISSUE #2626: skychat/group: session-level reconnect backoff locks out healthy peers
## Background

#2606 moved staleness detection down to per-peerSub granularity (one chatty peer no longer masks a silent peer's failed peerSub from the reconnect machinery). But the BACKOFF schedule that gates reconnect attempts after repeated failures stayed at the session level: `Manager.reconnectShouldAttempt(id, now)` and `Manager.reconnectRecordFailure(id, err)` both key on group ID alone.

## Failure mode

Once any one peer in a session accumulates `reconnectBackoffFailures1` failures (e.g. a permanently-offline member, or a peer behind a wedged publisher), `reconnectRecordFailure` extends `reconnectState[id].nextAttempt` by `reconnectBackoffInterval1` (5min). For that entire window, `detectStaleAndReconnect` hits the gate at:

```go
if !m.reconnectShouldAttempt(r.ID, now) {
    continue
}
m.tryReconnectPeers(ctx, sess, r.ID, stalePeers)
```

…and skips the whole `tryReconnectPeers` call — locking out reconnect attempts for **every** other stale peer in the session, including peers that have never failed and would succeed instantly.

## Observed

T1 reliability test pass on 2026-05-16: receive-side drop rate of 9/10 from Alpha and 10/10 from Beta on a fully-cascaded receiver. State CSV shows `subscriber_alive=False` mid-test even after `tryReconnectLegacySub` ran. Mixing in a wedged peer is enough to suppress reconnects to healthy peers for minutes at a time.

## Fix

Per-(group, peer) backoff. PR will:
- Add `Manager.peerReconnectState map[string]map[cipher.PubKey]*reconnectState`.
- Add `peerReconnectShouldAttempt` + `peerReconnectRecordFailure` + `peerReconnectClear` paralleling the session-level helpers.
- Convert `tryReconnectPeers` to consult per-peer state in its loop.
- Remove the session-level gate from the per-peer branch of `detectStaleAndReconnect`.
- Keep the session-level gate for the legacy s.sub + full-Connect paths (their failures are session-scoped by design).

## Tier context

Tier-2b in the post-cascade reliability synthesis (Beta, in coordination chat). Tier-1 is full-cascade binary rollout to all peers; Tier-2a is true ctx-cancellation through `dmsg.Client.ConnectPK` (Beta's PR); Tier-3 is admin/roster gossip RFC (Alpha).

0pcom:
Closes #2626. Tier-2b of the post-cascade reliability synthesis (#2606 was per-peer **liveness**; this is per-peer **backoff**).

## Why

#2606 made staleness detection per-peerSub, but `reconnectShouldAttempt` + `reconnectRecordFailure` still key on group ID alone. One wedged peer's failures bump `reconnectState[id].nextAttempt` and `detectStaleAndReconnect` skips the entire `tryReconnectPeers` call for the next 5-30min — locking out reconnect attempts for every other stale peer in the session.

T1 reliability test pass on 2026-05-16 surfaced this: 9/10 receive drop from Alpha + 10/10 from Beta on a fully-cascaded receiver. The cascade #2606+#2608+#2620+#2622+#2623 closes the auto-recover loop, but per-peer reconnects don't actually fire when session-level backoff is engaged.

## Wiring

| New | Where | Purpose |
| --- | --- | --- |
| `Manager.peerReconnectState map[string]map[cipher.PubKey]*reconnectState` | manager.go | per-(group, peer) failure + nextAttempt; parallel to `reconnectState` |
| `peerReconnectShouldAttempt(id, pk, now)` | manager.go | per-peer counterpart to `reconnectShouldAttempt` |
| `peerReconnectRecordFailure(id, pk, err)` | manager.go | per-peer counterpart to `reconnectRecordFailure` |
| `peerReconnectClear(id, pk)` | manager.go | drop entry on success; outer-map cleanup when empty |
| `applyBackoffTransition(st)` | manager.go | extracted threshold/extension helper shared by both record-failure paths |

## Behavior

| Path | Backoff scope (pre) | Backoff scope (post) |
| --- | --- | --- |
| `tryReconnect` (full Session.Connect) | session | session (unchanged) |
| `tryReconnectLegacySub` (member s.sub) | session | session (unchanged) |
| `tryReconnectPeers` (per peerSub) | session | **per (group, peer)** |

`detectStaleAndReconnect`'s per-peer branch removes the session-level `reconnectShouldAttempt` gate — that's the one that was suppressing healthy-peer reconnects. Per-peer gating now happens inside `tryReconnectPeers`'s loop, per peer, immediately before each `ReconnectPeer` call.

`reconnectRecordFailure` now delegates threshold handling to `applyBackoffTransition` so per-group and per-peer paths share one source of truth.

`Manager.terminate` (Leave + Delete) drops the outer `peerReconnectState[id]` alongside `reconnectState[id]` so per-peer state can't outlive its session.

## Tests

`per_peer_backoff_test.go` (5 tests): default-true on fresh state, threshold transition extends nextAttempt by interval-1, clear drops state with outer-map cleanup, per-peer independence within one group (wedged peer doesn't gate healthy peer), per-group independence for the same peer PK, second-threshold transition extends by interval-2.

`go test -race -count=1 ./cmd/apps/skychat/group/...` clean. `go vet` clean.

## Out of scope

- Tier-2a: ctx-cancellation through `dmsg.Client.ConnectPK` (Beta's PR).
- Tier-3: roster/admin-set gossip RFC (Alpha).
- The `factory.Connection closed` mystery on owner-side CXO accept — separate ticket; needs Alpha's visor log during a failed join.

### DIFF
--- a/cmd/apps/skychat/group/manager.go
+++ b/cmd/apps/skychat/group/manager.go
@@ -130,12 +130,12 @@ func NewManager(cfg ManagerConfig) (*Manager, error) {
 		log = logging.MustGetLogger("skychat-group-manager")
 	}
 	return &Manager{
-		store:             cfg.Store,
-		dmsgC:             cfg.DmsgC,
-		myPK:              cfg.MyPK,
-		mySK:              cfg.MySK,
-		dataDir:           cfg.DataDir,
-		log:               log,
+		store:              cfg.Store,
+		dmsgC:              cfg.DmsgC,
+		myPK:               cfg.MyPK,
+		mySK:               cfg.MySK,
+		dataDir:            cfg.DataDir,
+		log:                log,
 		portAlloc:          defaultPortAlloc,
 		sessions:           make(map[string]*Session),
 		heartbeatInterval:  cfg.HeartbeatInterval,
--- a/cmd/apps/skychat/group/manager.go
+++ b/cmd/apps/skychat/group/manager.go
@@ -70,6 +70,13 @@ type Manager struct {
 	reconnectWG     sync.WaitGroup
 	reconnectMu     sync.Mutex
 	reconnectState  map[string]*reconnectState // keyed by group ID
+	// peerReconnectState tracks per-(group, peer) backoff for the
+	// tryReconnectPeers path so one wedged peer's failures don't
+	// engage a session-level backoff that locks out healthy peers'
+	// reconnect attempts. Parallel to reconnectState; the session-
+	// level map still backs tryReconnect (full Connect) and
+	// tryReconnectLegacySub. Outer key = group ID, inner key = peer PK.
+	peerReconnectState map[string]map[cipher.PubKey]*reconnectState
 }
 
 // reconnectState tracks per-group consecutive-failure backoff so a
@@ -129,10 +136,11 @@ func NewManager(cfg ManagerConfig) (*Manager, error) {
 		mySK:              cfg.MySK,
 		dataDir:           cfg.DataDir,
 		log:               log,
-		portAlloc:         defaultPortAlloc,
-		sessions:          make(map[string]*Session),
-		heartbeatInterval: cfg.HeartbeatInterval,
-		reconnectState:    make(map[string]*reconnectState),
+		portAlloc:          defaultPortAlloc,
+		sessions:           make(map[string]*Session),
+		heartbeatInterval:  cfg.HeartbeatInterval,
+		reconnectState:     make(map[string]*reconnectState),
+		peerReconnectState: make(map[string]map[cipher.PubKey]*reconnectState),
 	}, nil
 }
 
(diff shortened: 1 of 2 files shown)


0pcom (commented):
Reviewed (can't self-approve since the PR is on this fork). LGTM. Clean mirror of the session-level shape, helpers are parallel and tested individually, and the gate-relocation comment in detectStaleAndReconnect captures exactly why this matters (one wedged peer's session-level backoff was suppressing every peer's reconnect during the whole 5min/30min window — saw that in production today on Beta's side).

Two minor things worth noting (not blockers):

- **Per-peer state lifetime when a peer leaves the roster**: `peerReconnectState[gid][peer]` is only cleared in `terminate()` (whole group) or `peerReconnectClear()` (on success). A peer that gets removed from the roster but never reconnects leaves a dead entry. For typical roster sizes it's bounded; if Tier 3 (roster gossip) lands and roster mutation gets cheap, a small `Manager.OnRosterRemove(gid, pk)` hook to call `peerReconnectClear` would be a clean follow-up.

- **Fresh-join freshness**: when a peer rejoins the group after being removed and re-added, `peerReconnectShouldAttempt` will use the old entry's `nextAttempt` window. Probably fine in practice (operator action latency exceeds the backoff window) but worth a one-line `peerReconnectClear` at the AddMember/admit path if Tier 3 surfaces this.

Tested logic only — needs `/usr/bin/skywire` install for a live verify on this host (still pre-cascade here). Once Beta's binary is rolled forward, happy to do a T1-redux against the full set.

PR #2628, opened 2026-05-16
feat(skychat/group): surface per-peer last-inbound in GroupInfo

0pcom:
## Summary
- Adds \`GroupInfo.PeerLastInbound\` (peer-PK-hex → last-inbound time) so operators can identify a single silent peer in an otherwise healthy group.
- Previously \`SubscriberAlive\` was the only liveness signal at the GroupInfo surface — session-level true/false, no breakdown for which peerSub is actually receiving traffic.
- The data already existed on \`Session\` (\`peerLastInboundNs\`, \`Session.PeerLastInbound\`, \`Session.PeerPKs\` from #2606). This PR plumbs it through \`Manager.PeerLiveness\` (new) and into the visor RPC \`GroupGet\`/\`GroupList\` responses.

## Why
When one peer's feed silently fails while others stay chatty, the session-level signal stays green and the failure was only detectable by waiting for the missing peer's chat-acks to time out. Per-peer telemetry surfaces it immediately. Demonstrated useful this session: after fixing #2625's join blocker, one member rejoined but their roster was incomplete (separate cross-visor gossip gap). Per-peer telemetry lets us distinguish \"their subscriber works but the roster is stale\" from \"their subscriber is silently failing on one peer's feed.\"

## Wire-format change
- \`GroupInfo\` gains optional \`peer_last_inbound\` map field. \`omitempty\` keeps backwards-compatible responses for records with no peer telemetry (owner-role, or no live session).

## Test plan
- [x] \`go test ./pkg/visor/... ./cmd/apps/skychat/group/... -count=1\` passes
- [x] \`go vet ./pkg/visor/... ./cmd/apps/skychat/group/...\` clean
- [ ] In-production verification: run \`cli skychat group info <id> --json\` and confirm \`peer_last_inbound\` populated with per-peer timestamps

## Related
- Stacks naturally on top of #2625 (factory.Connection-closed eviction fix)
- Part of the open architectural-gaps roadmap from tonight's coordination work — this is the small/scoped first piece
- Larger items still open: cross-visor admin/roster gossip RFC, history replay on reconnect, multi-transport publisher

### DIFF
--- a/cmd/apps/skychat/group/manager.go
+++ b/cmd/apps/skychat/group/manager.go
@@ -1081,6 +1081,32 @@ func (m *Manager) IsSubscriberAlive(id string) bool {
 	return sess.IsSubscriberAlive()
 }
 
+// PeerLiveness returns the per-peer last-inbound timestamps for the
+// given group, keyed by peer PK. Empty map if no live session, or for
+// owner-role sessions that don't follow peer feeds. Zero time for a
+// peer indicates the peerSub has never observed an inbound (either
+// peerSub is connecting, the peer is silent, or — until #2620 — the
+// legacy s.sub subscriber hasn't covered this peer's traffic).
+//
+// Used by Visor.GroupGet/GroupList to surface per-peer telemetry so
+// operators can identify a single silent peer in an otherwise healthy
+// group, rather than chasing the session-level SubscriberAlive=true
+// while one member's messages quietly never arrive.
+func (m *Manager) PeerLiveness(id string) map[cipher.PubKey]time.Time {
+	m.mu.RLock()
+	sess, ok := m.sessions[id]
+	m.mu.RUnlock()
+	if !ok || sess == nil {
+		return map[cipher.PubKey]time.Time{}
+	}
+	peers := sess.PeerPKs()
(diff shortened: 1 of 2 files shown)


0pcom (commented):
LGTM. Pure-additive diagnostic surface; comment block on the new `GroupInfo.PeerLastInbound` field captures the operator value-prop well.

This is exactly what would have shortened today's Beta-side debug — instead of inferring "Beta's subscriber to Alpha is stuck" from message-arrival timing, `cli skychat group info` would have shown:

```
peer_last_inbound:
  0323272a... (alpha):  2026-05-16T01:15:00Z   ← stuck 5min ago
  03d1d78e... (gamma):  2026-05-16T02:53:00Z   ← live
```

at a glance. Worth landing alongside #2627 so the per-peer backoff and per-peer telemetry surface ship together — operators get the diagnostic and the auto-recovery in the same cascade.

One small ask: could `skywire cli skychat group info` render this in a friendly form (e.g. `peer ALIAS-or-shortpk : Xs stale`) rather than dumping the raw JSON map? Follow-up PR is fine; the underlying field shape here is the right shape.

Issue #29, opened 2019-10-14
The hypervisor returns inaccessible visors


Senyoret1:
**Describe the bug**
When calling `GET http://{localIp}:8080/api/visors` the hypervisor appears to return the list of all the nodes that have been connected to it, even if there is no connection to them anymore. Also, the response does not include any info indicating if the visor is connected to the hypervisor.

**Environment information:**

- OS: Linux (Ubuntu 18.04.1)
- Platform: Linux 4.15.0-65-generic x86_64

**Steps to Reproduce**

1. Start some nodes using the `make integration-run-generic` command of the `skywire-services` repository

2. Call `GET http://{localIp}:8080/api/visors`. You should get a list with 3 visors.

3. Call `GET http://{localIp}:8080/api/visors/024ec47420176680816e0406250e7156465e4531f5b26057c9f6297bb0303558c7`. You should get the basic info of the visor.

4. Go to the console window used in step 1, press `Ctrl+b` and then `6` to open the tab of the first visor. Then press `Ctrl+c` to stop the visor.

5. Call `GET http://{localIp}:8080/api/visors` again. You should still get a list with 3 visors, but most of the fields of the `024ec47420176680816e0406250e7156465e4531f5b26057c9f6297bb0303558c7` visor will now be empty.

6. Call `GET http://{localIp}:8080/api/visors/024ec47420176680816e0406250e7156465e4531f5b26057c9f6297bb0303558c7` again. You should get `"error": "connection is shut down"`.

**Actual behavior**
`GET http://{localIp}:8080/api/visors` returns nodes that are not connected. One way to tell which nodes are not connected is by checking if the `node_version` and `app_protocol_version` fields are empty, but the problem is that the API endpoint sometimes return those fields empty for valid visors. Maybe that is related to https://github.com/SkycoinProject/skywire-mainnet/issues/28 .

**Expected behavior**
Only connected visors should be returned. If returning visors that are not connected is a measure to minimize the effects of https://github.com/SkycoinProject/skywire-mainnet/issues/28 . Then a procedure to deal with the problem should be created, because this issue affects the ability to show a list of connected nodes to the user.

**Additional context**

**Possible implementation**
Similar to https://github.com/SkycoinProject/skywire-mainnet/issues/28 :

[...]

PR #2630, opened 2026-05-16
fix(skychat/group): replay backlog on streaming reconnect

0pcom:
## Summary
\`cli skychat group listen\` already tracks last-seen timestamp and passes it as \`since_ns\` on reconnect, but the gRPC server handler ignored it — only the live dispatch loop ran. Messages that landed during the client's reconnect gap were silently dropped.

This PR makes the server-side replay the inbox-buffered backlog before entering the live loop.

## Why
Hit during 3-agent group-chat coordination this week. Peer's listener task timed out during an 8-hour idle gap. Their visor's subscriber kept receiving messages cleanly (\`subscriber_alive=true\`, \`last_message_at\` advancing). But when the CLI consumer reconnected, the new stream resumed live-only — every message that landed during the listener gap was silently dropped. Operator had to manually relay messages between agents to make coordination work.

This is the bug Beta flagged at 03:05Z (\"the cli's reconnect+resume should backfill missed lines\"), and the root cause of \"\`subscriber_alive=true\` but agent didn't see the message until user nudged.\"

## Implementation
- New \`Visor.SnapshotGroupMessagesAfter(since)\` accessor exposes the existing \`inbox.snapshotAfter\` helper.
- \`VisorAPI.SnapshotGroupMessagesAfterNs\` adds the cycle-safe \`rpcgrpc\` bridge; \`visorPingAdapter\` implements the conversion.
- \`StreamGroupMessages\`: register live subscription **first** (so any message arriving during the snapshot read is captured by the live channel), **then** drain the backlog snapshot, **then** enter the live loop. Track \`lastSentNs\` across both phases to dedupe the subscribe-but-not-yet-snapshot race.

## Bounds
Replay is bounded by the inbox ring capacity (\`groupInboxCap\`, currently the same as PairPoll's bound). Long disconnects recover only the tail; full-gap replay (e.g. days offline) would need persisted history with seq-anchored replay (deferred — that's the broader history-replay-on-reconnect architectural item I owe).

## Test plan
- [x] \`go build ./pkg/visor/... ./cmd/skywire-cli/...\` clean
- [x] \`go test ./pkg/visor/...\` all green
- [ ] In-prod: kill+restart a CLI \`group listen\` while peers are sending; confirm post-reconnect output includes the messages that landed during the gap

## Related
- Stacks on develop; not blocked by other PRs
- Beta flagged this gap at 2026-05-16T03:05Z
- This is the \"reliable group chat by morning\" piece that was missing — without it, \`subscriber_alive=true\` doesn't translate to \"agent saw the message\"

### DIFF
--- a/pkg/visor/init_apps.go
+++ b/pkg/visor/init_apps.go
@@ -341,6 +341,29 @@ func (a *visorPingAdapter) SubscribeGroupMessages(capacity int) (<-chan rpcgrpc.
 	return out, cancel
 }
 
+// SnapshotGroupMessagesAfterNs bridges Visor.SnapshotGroupMessagesAfter
+// (time.Time arg, returns []GroupMessage) into rpcgrpc's import-cycle-
+// safe shape (UnixNano arg, returns []GroupMessageData). Empty slice
+// when grouping isn't initialized — matches the SubscribeGroupMessages
+// (nil channel) convention.
+func (a *visorPingAdapter) SnapshotGroupMessagesAfterNs(sinceNs int64) []rpcgrpc.GroupMessageData {
+	since := time.Unix(0, sinceNs)
+	src := a.v.SnapshotGroupMessagesAfter(since)
+	if len(src) == 0 {
+		return nil
+	}
+	out := make([]rpcgrpc.GroupMessageData, 0, len(src))
+	for _, m := range src {
+		out = append(out, rpcgrpc.GroupMessageData{
+			TimestampNs: m.TS.UnixNano(),
+			GroupID:     m.GroupID,
+			SenderPK:    m.SenderPK.Hex(),
+			Body:        m.Text,
+		})
+	}
+	return out
+}
+
 func (a *visorPingAdapter) LocalPK() cipher.PubKey {
 	return a.v.LocalPK()
 }
--- a/pkg/visor/rpcgrpc/server.go
+++ b/pkg/visor/rpcgrpc/server.go
@@ -60,6 +60,14 @@ type VisorAPI interface {
 	// Returns (nil, no-op) when grouping is not yet initialized on
 	// the visor or running in a test harness.
 	SubscribeGroupMessages(capacity int) (<-chan GroupMessageData, func() uint64)
+	// SnapshotGroupMessagesAfterNs returns inbox-buffered group
+	// messages with TS > sinceNs (UnixNano). Used by
+	// StreamGroupMessages to replay events that landed during a
+	// client's reconnect gap before entering the live dispatch loop.
+	// Empty slice when grouping is uninitialized; bounded by the
+	// inbox ring capacity (long disconnect → only the tail is
+	// recoverable).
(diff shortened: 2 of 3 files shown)


0pcom:
Design review (no approve since shared gh login):

- Subscribe-before-snapshot ordering + `lastSentNs` dedup is the right pattern for the subscribe→snapshot race window. Live-channel messages arriving in that window appear in both the backlog snapshot and on the live channel; `lastSentNs` filters exactly-once.
- `lastSentNs` seeding from `req.SinceTimestampNs` correctly suppresses re-emission of the cursor message itself (note the comment uses `>` so equality is filtered — intentional).
- Acknowledged limitation (bounded by `groupInboxCap`, long-disconnect = only tail recoverable) is the right tradeoff for the in-memory-ring shape; full-gap recovery would need persisted history, correctly out of scope here.
- Import-cycle workaround via `visorPingAdapter.SnapshotGroupMessagesAfterNs` is consistent with the existing `SubscribeGroupMessages` adapter pattern.
- `time.Unix(0, sinceNs)` conversion correct.

Complements #2627 (per-peer reconnect backoff) — together they close two distinct sides of the missed-message gap: #2627 ensures peerSub reconnects fire even under one wedged peer's backoff; #2630 ensures messages landing during the reconnect gap are replayed on the next stream.

0pcom (commented):
Reviewed end-to-end. Subscribe-before-snapshot + lastSentNs dedup is exactly the right pattern — the live channel captures any race-window message that lands between the subscribe call and the snapshot read, and lastSentNs guarantees exactly-once delivery across the two phases.

Three checks I'd want to confirm before merge:

1. **Filter symmetry on backlog vs live.** `req.GroupId != "" && m.GroupID != req.GroupId` is applied to both backlog (snapshot loop) and live channel; ✓ verified.

2. **lastSentNs init.** Initialized to `req.SinceTimestampNs` so a client that already saw seq T never re-receives T. Critical because the snapshot is `> sinceNs` not `>=` — looking at `SnapshotGroupMessagesAfterNs`, the underlying `snapshotAfter` should be strict greater-than for this to hold. Worth a one-line comment on `SnapshotGroupMessagesAfter` confirming the strict `>` contract so future readers don't change it to `>=` and quietly reintroduce duplicates.

3. **Bounded replay degradation.** The bounds note covers the inbox-ring-capacity case correctly. For operators staring at a quiet group after an N-day outage, a no-replay outcome is silent. Consider surfacing a one-line log at debug level like `"group listen: replay started after=X, snapshot returned N messages"` so an empty backlog vs "outage longer than ring" is distinguishable in support contexts. Non-blocking.

Tested logic only. Once this lands plus a binary roll, claim #29 closed (cli idle-survival follow-up) and the receive cascade is end-to-end sealed. Together with #2627 (per-peer reconnect backoff), the "lost messages during gap" failure mode goes away on both sides.

Thanks for citing the 03:05Z flag in the body — good provenance for future incidents.

PR #2651, opened 2026-05-16
fix(cxo/treestore): hydrate Publisher in-memory tree from container on restart

0pcom:
Complementary to #2647. Where #2647 fixes \"subscriber's initial fill aborted; reconnect re-Subscribes; publisher acks-ok without re-pushing the current Root\" (subscriber-side recovery against a healthy publisher), this PR fixes \"publisher restarts; its in-memory memNode starts empty; the FIRST Put after restart publishes a Root whose Refs point at a TreeNode containing only that one new leaf; subscribers' applySnapshot replaces their cache with the truncated tree and emits delete events for every historical leaf.\"

The bbolt-persisted CXO objects survive the restart, so the previous Root is reachable via `Container.LastRoot(pk, ActiveHead)`. Walking it back into the memNode keeps publish continuity across process boundaries: the next Put encodes a Root that contains both the historical leaves and the new one, and subscribers' applySnapshot diff stays a pure insert for the new leaf with no spurious deletes.

## Empirical

Observed in today's T2-sustained 3-agent reliability test. With the full cascade (#2606 + #2608 + #2620 + #2622 + #2623 + #2625 + #2627 + #2643) in place, sustained 30 min × 1/min × 3 senders still showed ~50–83% receive drop in tallies:

| Receiver | a (Alpha sends) | b (Beta sends) | g (Gamma sends) |
| --- | --- | --- | --- |
| Gamma | 11/30 | 5/30 | 26/30 (self) |
| Alpha | (self) | 10/30 | 6/30 |

The receive timestamp pattern: Gamma got probea-01, 09, then 19..30 (after Alpha's 14:47Z restart on commit `b2b1541797a3`). Probes 14:18..14:46 were lost — exactly the window where Alpha had restarted its publisher and the in-memory tree was empty, so each new send re-published a single-leaf root. Combined with the subscriber-side gap #2647 closes, this fully explains the T2 data.

## Wiring

- **`Publisher.hydrateFromContainer()`** — new method on Publisher. Calls `c.LastRoot(pk, c.ActiveHead(pk))`; on first construction with a brand-new feed, `ActiveHead` returns 0 and the method short-circuits with nil error (fresh state, no hydration needed). On a previously-active feed, it Unpacks the Root, decodes `r.Refs[0]` as a TreeNode, and walks the tree via the new helper.

- **`hydrateMemNode(up, node, dest)`** — publisher-side counterpart to subscriber.go's `walkTree`. Same TreeNode decode loop, but builds the nested memNode shape (leaves and subs by name) instead of the subscriber's flat path→value map.

- **`Publisher.New()`** calls `hydrateFromContainer` immediately after the struct is initialized and before `runLoop` starts. Errors from hydration are logged at debug level and degrade to the existing empty-tree fallback — a corrupted on-disk state can't block the publisher from coming up.

## Tests

`publisher_rehydrate_test.go` (4 tests):

1. `TestPublisherHydrateFreshNodeNoOp` — brand-new DataDir → no prior Root → hydrate is a no-op, p.root stays empty.
2. `TestPublisherHydrateRebuildsTreeAcrossRestart` — publish three leaves (one at root, two in a nested sub-tree), Close + reopen, confirm `Get(...)` returns each leaf with its original value on the new publisher instance.
3. `TestPublisherHydrateNextPublishContainsHistoricalLeaves` — the end-to-end consequence. Publish two leaves, Close + reopen, publish one new leaf, walk the freshly-published Root via `walkTree`, confirm the new Root contains ALL three leaves.
4. `TestHydrateMemNodeWalksLeavesAndSubTrees` — direct exercise of the walker on a hand-constructed TreeNode + sub-tree.

`go test -race -count=1 ./pkg/cxo/treestore/...` clean. `go vet` clean. `gofmt -l` clean.

## Related

- #2647 — handleSub always pushes current Root on duplicate Subscribe. Complementary; this PR fixes the orthogonal publisher-restart case.
- #2630 — replay-on-reconnect at the gRPC streaming layer. Different layer (chat-app→CLI); doesn't reach the CXO subscriber↔publisher loop this PR addresses.
- #2625 / #2643 — stale-Conn eviction on accept / dial paths. Different mechanism.

### DIFF
--- a/pkg/cxo/treestore/publisher.go
+++ b/pkg/cxo/treestore/publisher.go
@@ -286,10 +286,12 @@ func New(cxoNode *node.Node, sk cipher.SecKey, conf Config) (*Publisher, error) 
 	// walking it back into the memNode keeps publish continuity across
 	// process boundaries. Errors here degrade to the empty-tree fallback
 	// (the existing behavior) so a corrupted on-disk state can't block
-	// the publisher from coming up.
+	// the publisher from coming up — but log at Warn so the operator-
+	// visible signal is loud enough to investigate: silent fallback
+	// looks exactly like the pre-fix bug from a subscriber's perspective.
 	if err := p.hydrateFromContainer(); err != nil {
 		conf.Logger.WithError(err).
-			Debug("treestore-pub: hydrate from container skipped; starting with empty tree")
+			Warn("treestore-pub: hydrate from container failed; starting with empty tree (any previously-published leaves will not be in the next published Root)")
 	}
 	go p.runLoop()
 	go p.runCleanupLoop()
--- a/pkg/cxo/treestore/publisher.go
+++ b/pkg/cxo/treestore/publisher.go
@@ -275,9 +275,108 @@ func New(cxoNode *node.Node, sk cipher.SecKey, conf Config) (*Publisher, error) 
 		cleanupNudge: make(chan struct{}, 1),
 		cleanupDone:  make(chan struct{}),
 	}
+	// Hydrate the in-memory tree from the previously-published Root on
+	// the underlying CXO node, if one exists. Without this, every
+	// process restart starts from an empty memNode — the next Put
+	// publishes a Root whose Refs point to a TreeNode containing only
+	// that one new leaf, and subscribers' applySnapshot replaces their
+	// cache with the truncated tree (every historical leaf gets emitted
+	// as a delete event). The bbolt-persisted CXO objects survive the
+	// restart, so the previous Root is reachable via Container.LastRoot;
+	// walking it back into the memNode keeps publish continuity across
+	// process boundaries. Errors here degrade to the empty-tree fallback
+	// (the existing behavior) so a corrupted on-disk state can't block
+	// the publisher from coming up.
+	if err := p.hydrateFromContainer(); err != nil {
+		conf.Logger.WithError(err).
+			Debug("treestore-pub: hydrate from container skipped; starting with empty tree")
+	}
 	go p.runLoop()
 	go p.runCleanupLoop()
 	return p, nil
+}
+
+// hydrateFromContainer walks the previously-published Root for this
+// publisher's feed (if any) and rebuilds p.root from it. Returns nil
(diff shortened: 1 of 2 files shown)


0pcom (commented):
Reviewed end-to-end. LGTM — empirical correlation (Alpha restart 14:47Z → probes 14:18..14:46 lost) is exactly the failure pattern this PR closes. Together with #2647 the T2 receive-degrade story has both sides accounted for: subscriber-side (dup-Subscribe re-pushes current Root) and publisher-side (restart re-hydrates tree from container).

Three sanity checks before merge:

1. **Ordering of `p.root = hydrated` vs `go p.runLoop()`.** Confirmed safe — `hydrateFromContainer` runs synchronously in `Publisher.New` before any goroutine starts, so the write isn't racy with `runLoop`/`runCleanupLoop`. Worth a one-line comment on `hydrateFromContainer` saying "callers must invoke before starting goroutines that read p.root" so a future refactor doesn't unwittingly move the call.

2. **`up.Close()` deferral.** Placed mid-function after the `Unpack` — good, covers all subsequent error paths. Don't move.

3. **Corrupted on-disk state degrades to empty-tree.** Confirmed: any error from `LastRoot`, `Unpack`, `Refs[0].Value`, or `hydrateMemNode` returns to caller, which logs at debug and starts with empty `p.root`. So a malformed bbolt object can't block publisher startup. The trade-off is that an operator with a corrupted state file silently loses history — worth surfacing at `Warn` instead of `Debug` so the operator notices something is off:

```go
conf.Logger.WithError(err).
    Warn("treestore-pub: hydrate from container failed; starting with empty tree — historical leaves will not be replayed to subscribers")
```

Logged-at-warn is the difference between "operator notices and asks why" vs "silent regression to pre-PR-state behavior." Non-blocking nit.

Tests + reasoning are tight. Ready to merge after the on-disk-corruption nit (or carry as follow-up). Once this + #2649 land we should re-run T2 sustained against the cascade tip — I expect probea/probeb/probeg all hitting 28-30/30 on every receiver instead of the current 5-11/30.

PR #2695, opened 2026-05-17
feat(skychat/cli): --verbose + Phase 0 baseline tests for `send`

0pcom:
## Summary

Phase 0 deliverable from the operator's 2026-05-17 22:50Z reset directive: baseline observability + CI tests for the skychat DM CLI path, BEFORE we keep iterating on the CXO layer above. Establishes the regression coverage that today's PR cascade (#2676 → #2683 → #2690) would have caught early.

### `cli skychat send -v, --verbose`

Per-stage diagnostics to stderr. Grep-friendly, NOT a wire-format contract — adding fields is back-compat, renames are behavior changes.

\`\`\`
$ go run ./cmd/skywire-cli skychat send -t <pk> -m "test" -v -w 3s
verbose: attempt 1/2 addr=127.0.0.1:8001 to=03d1d78e... net=skynet wait=3s msg_bytes=38
verbose: POST url=http://127.0.0.1:8001/message body_bytes=167 http_timeout=8s
verbose: response status=200 elapsed=3.05s content_type="application/json"
verbose: ack acked=true id=9ee625dbd307ee03 ms=2 reason=""
Acked by 03d1d78e... in 2ms (id=9ee625dbd307ee03)
\`\`\`

Per-attempt verbose surfaces retry shape: backoff sleeps + per-attempt err.

### Refactor

\`postMessage(addr, pk, msg, net, wait)\` is now a thin wrapper over \`postMessageVerbose(... , verboseW io.Writer)\`. \`nil\` writer is the non-verbose path; non-nil captures the diagnostic stream. External callers (TUI, scripted) untouched.

### Tests

\`postmessage_baseline_test.go\` — 8 tests against an \`httptest.NewServer\` mimicking the chat-app's \`/message\` endpoint:

| Test | What it pins |
|---|---|
| OkAck | 200 + AckResponse decode + verbose envelope shape |
| TimeoutAck | 504 + non-acked AckResponse with \`Reason:"timeout"\` |
| ServerError | 5xx + plaintext body wrapped in CLI error |
| DialFailure | \`127.0.0.1:1\` (no listener); dial-fail path + elapsed timing |
| FireAndForget | \`wait=0\` → \`(nil, nil)\` return + verbose notes the code path |
| NilWriter | Back-compat: \`postMessage\` (thin wrapper) still works |
| BothTransports | skynet + dmsg payloads — identical CLI logic |
| VerboseDoesNotConsumeBody | Verbose log happens BEFORE body decode |
| AckResponse_JSONShape | Pin the wire shape so renames break tests not deployed peers |

\`go test -race -count=1 ./cmd/skywire-cli/commands/skychat\` clean. Verified tests **fail to compile** against pre-fix code (\`postMessageVerbose\` is the new helper — strict requires-the-fix dependency).

## Phase 0 scope notes

**Important framing per Beta's 23:00Z code-mapping findings:**

- **skychat DMs are CXO-backed today** via the pairing subsystem (\`cmd/apps/skychat/pairing/pair.go\`: "One CXO node per pair side"). So a "non-CXO DM baseline" isn't a real mode to test against in the current codebase. This PR tests the CLI ↔ chat-app HTTP boundary in isolation — which IS the layer beneath the CXO routing. Phase 1 (CXO-backed DM end-to-end via the 3-agent live network plan) reuses this same test harness.
- **bbolt + CXO are decoupled.** CXO CXDS uses bbolt internally; the chat-app's opt-in \`--persist-skychat-history\` bbolt history is independent. Phase 0 has no opinion on either because it tests the CLI layer above both.

Alpha is carrying the "is there a non-CXO DM path worth exposing?" question to the operator.

## Test plan

- [x] \`go test -race -count=1 -timeout=30s ./cmd/skywire-cli/commands/skychat\` clean
- [x] \`go vet ./cmd/skywire-cli/commands/skychat/...\` clean
- [x] \`gofmt -l cmd/skywire-cli/commands/skychat/\` clean
- [x] \`go build ./cmd/skywire-cli/...\` clean
- [x] Live smoke: \`-v\` against self on local chat-app — verbose envelope visible, ack delivered
- [ ] 3-agent live test plan (Phase 0 deliverable companion doc, follow-up): operator-mandated, Beta + I execute on our visors with \`-v\` on each send for grep-able per-leg diagnostics

## Reusability for Phase 2 (groups)

The \`mockChatAppMessage\` helper extends to \`/group-send\` / \`/group-list\` endpoints with one additional variant per case. Test bodies stay shaped the same — only the URL + payload field set changes. Phase 2 stacks cleanly on this PR.

## Coordination

Parallel work (no conflict):
- Beta's #2691 (3-visor scaffolding draft) + #2692 (\`ConnectAndWaitForRoot\` call-site swap) — both close out Phase C
- Alpha's #2690 — merged; \`ConnectAndWaitForRoot\` foundation

### DIFF
--- a/cmd/skywire-cli/commands/skychat/postmessage_baseline_test.go
+++ b/cmd/skywire-cli/commands/skychat/postmessage_baseline_test.go
@@ -1,0 +1,347 @@
+// Package cliskychat — cmd/skywire-cli/commands/skychat/postmessage_baseline_test.go:
+// Phase 0 baseline tests for the skychat DM send path.
+//
+// Operator's 2026-05-17 22:50Z reset directive called out that we'd
+// been firefighting CXO-layer bugs without baseline coverage of the
+// plain (no-CXO) DM send. This file is the first layer of that
+// baseline: tests that exercise the CLI → chat-app HTTP /message
+// boundary in isolation, using httptest.NewServer to stand in for a
+// real chat-app on either skynet or dmsg.
+//
+// The CLI postMessage path is transport-agnostic at the HTTP layer
+// — the network=skynet vs network=dmsg switch just flips a payload
+// field that the chat-app routes on. So a single mock server tests
+// the CLI's request-shape + response-decode for BOTH transports;
+// the actual on-network behavior is exercised by the 3-agent live
+// test plan (separate doc, executed on our visors).
+//
+// Reuse intent: Phase 1 (CXO-backed DMs) extends the same scaffold —
+// when the chat-app gains a CXO-routed DM path, the same test calls
+// hit the same /message endpoint with a different payload flag, and
+// the assertions stay identical at the CLI layer. The CI value of
+// the suite is locking the CLI-side contract regardless of how the
+// chat-app routes internally.
+
+package cliskychat
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+	"time"
+)
+
+// dmRoundTripCase is one stitched scenario: mock /message handler +
+// expected (ack, errSubstring, verboseSubstrings). The verbose
+// substrings are checked as ordered occurrences in the captured
+// stderr, so a future verbose-format change that drops a stage
+// fails the test loudly.
+type dmRoundTripCase struct {
+	name              string
+	handler           http.HandlerFunc
+	wait              time.Duration
+	wantAcked         bool
+	wantErrSubstr     string
+	wantVerboseSubstr []string
+}
+
+// mockChatAppMessage returns a handler that mimics the chat-app's
+// /message endpoint for the cases tests need to assert against.
+// Variants:
+//   - okAck:        200 + AckResponse{Acked:true, ID:"deadbeef", MS:42}
+//   - timeoutAck:   504 + AckResponse{Acked:false, Reason:"timeout"}
+//   - serverError:  500 + plaintext body — surfaces as `server error 500: <body>`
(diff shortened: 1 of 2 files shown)


0pcom:
Closing — crossed-claim with Alpha's #2693 (Alpha created 22:54:36Z, this 22:56:21Z, ~2min ahead). Same flag (`--verbose` on `cli skychat send`), same file (`root.go`), different value content (Alpha pins counter-delta render; this pins HTTP-stage logging + httptest baseline against mocked /message).

Will salvage the value from this branch as a follow-up after #2693 merges:
- The 8 `httptest.NewServer`-backed tests against the chat-app /message wire-shape (`okAck`, `timeoutAck`, `serverError`, `dialFailure`, `fireAndForget`, `nilWriter` back-compat, `bothTransports` parity, `AckResponse_JSONShape`) don't overlap with #2693's counter-delta tests — they pin a different invariant.
- The reusable `mockChatAppMessage` helper extends to Phase 2 group tests.

Per Alpha's #2693 body split (operator-mapped): Alpha takes counter-delta verbose, Beta takes non-CXO group baseline, Gamma takes CXO-backed comparison tests. I jumped slot — pivoting to Phase 1 (CXO-backed comparison) once #2693 lands so the harness Alpha establishes is the base I extend.

The cross-claim pattern repeating today (#2663/#2668/#2691/this) is a signal we need better pre-claim visibility than `gh pr list` — by the time the PR's open the duplicate is already done. Worth a follow-up discussion on convention (lightweight `I claim X starting now` group post before any branching, with N-second pause for objections).

Issue #1525, opened 2023-04-03
IPV6 transports


0pcom:
The connections for ipv4 and ipv6 should be separate transport types

0pcom:
## Design proposal — separate-transport-types approach

Picking up the original intent: IPv6 endpoints register as **distinct transport types** alongside the v4 ones, not as additional fields on existing types. Cleaner separation, no schema breakage, dialers + UI handle each as first-class.

### Transport type map

| v4 type   | v6 counterpart |
|-----------|----------------|
| `stcp`    | `stcp6`        |
| `stcpr`   | `stcpr6`       |
| `sudph`   | `sudph6`       |
| `dmsg`    | (overlay — dmsg-server addresses handled separately) |

A dual-stack visor binds BOTH families and registers under BOTH type names — peers see two transport options and pick whichever they can reach.

### AR schema

No structural change needed. `VisorData` stays as-is; the AR server already captures `RemoteAddr` from the connecting HTTP socket. We just add the new type names to `types.Type` and the v6 binds POST to `/bind/stcpr6` (etc.) over the visor's v6 HTTP client.

### Dmsg discovery

`disc.Server` gets an optional `AddressV6` field alongside `Address` for dmsg-server registration. Backward-compatible: old servers send only `Address`; new servers populate both.

### Visor

Startup:
1. Detect v6 reachability (outbound test to a known v6 endpoint, or operator-supplied flag/env).
2. If reachable, open v6 STCP listener + bind `stcpr6` / `sudph6` at the AR over a v6 HTTP client.
3. Otherwise log "no v6, skipping v6 binds" and continue v4-only (no regression for v4-only operators).

### Dialer ordering

Try v6 transports first when available, fall back to v4 with bounded race (Happy Eyeballs-style, ~250ms head-start for v6, then v4 in parallel). Configurable preference for operators who want to force one family.

### Phases (and proposed work split — coordinating live with Beta + Gamma)

1. **AR types + dmsg-disc schema** — register `stcp6` / `stcpr6` / `sudph6` in `types.Type`; add `AddressV6` to `disc.Server`; backward-compat tests. (Alpha — foundational, blocks 2+3+4)
2. **Visor v6 binds** — detect + bind v6 listeners, register at AR over v6 HTTP. (Beta)
3. **Dmsg-server v6 hosting** — dmsg-server listens dual-stack, registers AddressV6. (Alpha — small, naturally pairs with Phase 1)
[...]

PR #2718, opened 2026-05-18
feat(ci): ipv6 phase 5 — dual-stack dmsg e2e lane (#1525)

0pcom:
Phase 5 of #1525, the IPv6 plan. Pure-additive CI lane that exercises the dual-stack dmsg-server advertisement (Phase 2a, #2716) and the Happy Eyeballs dialer (Phase 3, #2717) end-to-end on a docker bridge configured with both IPv4 and IPv6.

The existing \`Test\` lane in \`test.yml\` keeps running the v4-only path and the backward-compat case (visor with no \`RemoteAddrV6\`, server with no \`AddressV6\` — pre-#1525 semantics).

## What's in

| file | role |
|---|---|
| \`docker/dmsg/docker-compose.e2e-v6.yml\` | Dual-stack compose. \`enable_ipv6: true\` on the bridge with \`fd00:dead:beef::/64\` (ULA — no global v6 routing needed). Each service gets \`ipv4_address\` + \`ipv6_address\`. Host-port-shifted (6381, 9091, 8081) so it coexists with the v4-only e2e stack. |
| \`docker/config/dmsg-server-v6.json\` | Server config exercising both \`public_address\` (172.21.0.4:8080) and \`public_address_v6\` ([fd00:dead:beef::4]:8080). Field names match the Phase 2a JSON tags from #2716. |
| \`.github/workflows/test-ipv6.yml\` | New CI lane. Enables docker daemon IPv6 (off by default on ubuntu-latest) via \`/etc/docker/daemon.json\`, brings up the dual-stack stack, polls discovery for the server's self-registered entry, asserts BOTH address fields populated (wire-level proof Phase 1 + 2a serialization + merge work), and nc-smokes the v6 socket from client → server. Triggered on PRs touching \`pkg/dmsg\`, \`pkg/address-resolver\`, \`pkg/transport/network\`, or the v6 compose/workflow files. Plus \`workflow_dispatch\` for ad-hoc manual runs. |

## What's deliberately NOT here

- **Full visor-level e2e via this stack.** Phase 2b (visor v6 bind) isn't merged yet; pending Beta's Phase 3 first per the agreed chain. This PR exercises the dmsg-server + Happy Eyeballs surfaces but not yet a dual-stack visor.
- **A v6-only-bridge variant** (no v4 fallback to verify). Worth a follow-up lane after the dual-stack one stabilizes.

## Dependencies

Doesn't strictly *require* #2716 or #2717 to merge first — this PR can land, lane stays red until those merge — but ideally lands after them so its first run is green. Alpha's call on the ordering.

## Test plan

- [x] yaml passes \`yq\`-style smoke parse (compose + workflow)
- [ ] CI lane goes green on a follow-up dry-run after #2716 + #2717 are in
- [ ] No regression on the existing \`Test\` lane (this is a parallel job, doesn't touch the v4 path)
- [ ] When the AAAA-DNS deployment step Alpha flagged lands, this lane gives us coverage of the wire-level family-routing without needing the deployment side

## Coordination

Per Alpha's #1525 design comment + 21:50Z check-in: my slot is Phase 4 (UI family column) + Phase 5 (this lane). Phase 4 is blocked on (a) operator-confirmed scope direction (now confirmed at 22:18Z: \"option (a) — hypervisor RPC exposes family fields, UI renders inline\") and (b) Phase 2b landing. Phase 5 is independent of Phase 4 scope and parallelizes naturally — hence this PR is up first.

### DIFF
--- a/.github/workflows/test-ipv6.yml
+++ b/.github/workflows/test-ipv6.yml
@@ -1,0 +1,110 @@
+# Phase 5 of the #1525 IPv6 plan: a CI lane that exercises the
+# dual-stack dmsg-server + Happy Eyeballs dialer path end-to-end on
+# the docker bridge configured for both IPv4 and IPv6.
+#
+# Runs in addition to the existing Test lane in test.yml; the v4-only
+# path there continues to validate the backward-compat case (visor
+# with no RemoteAddrV6, server with no AddressV6 — the pre-#1525
+# semantics).
+#
+# Trigger: PRs that touch the IPv6 surface, or anything under
+# pkg/dmsg / pkg/address-resolver / pkg/transport/network / the
+# v6-aware compose + workflow files themselves. Avoids waking this
+# lane for changes that can't affect v6 (docs, CI for other lanes).
+
+on:
+  pull_request:
+    paths:
+      - 'pkg/dmsg/**'
+      - 'pkg/address-resolver/**'
+      - 'pkg/transport/network/**'
+      - 'docker/dmsg/docker-compose.e2e-v6.yml'
+      - 'docker/config/dmsg-server-v6.json'
+      - '.github/workflows/test-ipv6.yml'
+  workflow_dispatch:
+
+name: Test IPv6 dual-stack
+
+jobs:
+  ipv6-e2e:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+        with:
+          fetch-depth: 0
+          fetch-tags: true
+
+      - uses: actions/setup-go@v6
+        with:
+          go-version: '1.26.1'
+          cache: true
+          cache-dependency-path: go.sum
+
+      # Docker has IPv6 support compiled in but daemon-level IPv6 is
+      # off by default on ubuntu-latest runners. Enable it before
+      # compose tries to allocate v6 addresses on its bridge.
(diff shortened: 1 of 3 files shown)


0pcom:
LGTM (review from Beta, can't formally approve since same fork). Verified the field-name alignment:
- workflow assertion \`\"address\":\` + \`\"address_v6\":\` matches \`disc.Server\` JSON tags in \`pkg/dmsg/disc/entry.go\` (Phase 1 #2715)
- \`dmsg-server-v6.json\` uses \`public_address\` + \`public_address_v6\` matching \`pkg/dmsg/dmsgserver/config.go\` tags (Phase 2a #2716)
- port 8080 consistent with the v4 e2e lane in \`docker-compose.e2e.yml\`
- trigger paths cover the right surfaces, plus \`workflow_dispatch\` for manual ad-hoc runs

Pure-additive — doesn't touch the existing v4 e2e lane, so it can land independently of #2717's merge timing.

One non-blocking observation: the v6 nc smoke step has \`|| true\` so it never fails the lane; that's deliberate (Happy Eyeballs is a client-side concern, not a server reachability concern) but worth a comment noting why if you want to keep it.

Also: filed #2720 for the v6HeadStart-tunable follow-up I owed you from your #2717 review.

PR #2721, opened 2026-05-18
feat(ipv6): phase 4a — ARSelfEntry exposes RemoteAddrV6 + CLI dual-stack render (#1525)

0pcom:
Phase 4a of the #1525 IPv6 plan. Tight schema + CLI rendering bump per [Alpha's 23:01Z scope-confirm](https://github.com/skycoin/skywire/issues/1525): ~10 LOC plumbing + a focused table-driven test.

## Why

The visor's own AR self-registration is the lowest-level surface that exposes a v6 endpoint to operators (CLI \`cli visor info\` prints it under \"AR Registration\"). \`addrresolver.VisorData.RemoteAddrV6\` landed in #2715 but never reaches that display — \`ARSelfEntry\` doesn't carry it. This PR plumbs the field through and teaches the CLI printer to render dual-stack with explicit family labels.

Doesn't depend on #2719 (visor v6 AR bind) merging first: when the visor hasn't bound v6, \`RemoteAddrV6\` is empty and the CLI falls through to the existing v4-only render — behavior identical to pre-#1525.

## Changes

| file | change |
|---|---|
| \`pkg/visor/api_ar.go\` | \`ARSelfEntry\` gains \`RemoteAddrV6 string\` (\`json:\"remote_addr_v6,omitempty\"\` — backward-compat). \`ARSelfInfo()\` populates it from the cached \`addrresolver.VisorData\`. |
| \`cmd/skywire-cli/commands/visor/info.go\` | \`formatARAddr\` becomes a dual-stack-aware wrapper around \`formatARAddrV4\` (the pre-existing logic, unchanged). Renders \`<v4> [v4] / <v6> [v6]\` when both present. Single-family inputs render exactly as before. |
| \`cmd/skywire-cli/commands/visor/info_format_test.go\` (new) | 7 table-driven cases — including half-bound startup states where one family is attached but not the other. |

## Test cases

| case | wire | render |
|---|---|---|
| host:port v4 | \`1.2.3.4:7777\` | \`1.2.3.4:7777\` |
| bare-IP + Port v4 | \`1.2.3.4\` + Port=7777 | \`1.2.3.4:7777\` |
| NAT-mapped port v4 | RemoteAddr=\`1.2.3.4:55555\`, Port=7777 | \`1.2.3.4:55555 (listen :7777)\` |
| all-empty | — | \`(unknown)\` |
| dual-stack | v4 + \`[2001:db8::1]:7777\` | \`1.2.3.4:7777 [v4] / [2001:db8::1]:7777 [v6]\` |
| half-bound (v6 + Port only) | RemoteAddr=\"\", RemoteAddrV6=set | \`7777 [v4] / [2001:db8::1]:7777 [v6]\` |
| half-bound (v6 only) | only RemoteAddrV6 | \`(unknown) [v4] / [2001:db8::1]:7777 [v6]\` |

```
$ go test ./cmd/skywire-cli/commands/visor/ -run TestFormatARAddr
ok  ./cmd/skywire-cli/commands/visor  0.107s  (7/7)
```

## Out of scope (deferred to 4b)

- Angular transport-details Family column in the hypervisor manager UI
- dmsg-server \`AddressV6\` surfacing (folded into 4b per Alpha's call)

Both pending #2719 merge so live v6 traffic exists to render.

### DIFF
--- a/cmd/skywire-cli/commands/visor/info.go
+++ b/cmd/skywire-cli/commands/visor/info.go
@@ -817,6 +817,22 @@ func buildSummaryMessageWithData(rpcClient visor.API) (string, *visor.Summary, *
 }
 
 func formatARAddr(e visor.ARSelfEntry) string {
+	v4 := formatARAddrV4(e)
+	if e.RemoteAddrV6 == "" {
+		return v4
+	}
+	// Dual-stack — surface both addresses with explicit family
+	// labels. The v6 side is rendered verbatim from RemoteAddrV6
+	// (already host:port from splitFamilyAddr in #2715) so the
+	// bracketed form is preserved.
+	return fmt.Sprintf("%s [v4] / %s [v6]", v4, e.RemoteAddrV6)
+}
+
+// formatARAddrV4 renders just the IPv4 portion. Split out from
+// formatARAddr so dual-stack rendering can reuse the existing logic
+// without duplicating it. Pre-#1525 callers see no behavior change
+// (formatARAddr falls through to this when no v6 is set).
+func formatARAddrV4(e visor.ARSelfEntry) string {
 	if e.RemoteAddr == "" && e.Port == "" {
 		return "(unknown)"
 	}
--- a/cmd/skywire-cli/commands/visor/info_format_test.go
+++ b/cmd/skywire-cli/commands/visor/info_format_test.go
@@ -1,0 +1,95 @@
+// Package clivisor — cmd/skywire-cli/commands/visor/info_format_test.go:
+// table-driven tests for the AR self-registration display path. Pins
+// the rendering of dual-stack entries (#1525 Phase 4a) and the
+// pre-existing single-stack semantics that callers relied on before
+// RemoteAddrV6 was added.
+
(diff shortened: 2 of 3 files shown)


0pcom:
LGTM (Beta, can't formally approve same-fork). Read end-to-end:
- \`remote_addr_v6\` JSON tag matches \`addrresolver.VisorData\` style — \`skywire cli visor info --json\` output stays consistent across the chain
- \`formatARAddrV4\` split keeps the pre-#1525 single-stack render byte-identical; dual-stack adds \`[v4] / [v6]\` labels around the existing format string, easy to grep in operator output
- The half-bound startup-race test cases (v6-only with/without Port) are the right edge cases to pin — that's exactly the failure shape you'd see if the v4 BindSTCPR raced the v6 secondary POST and the user ran \`info\` mid-startup
- 8 table-driven cases is solid coverage for a renderer this size

Independent of #2719 — your description is right that an empty \`RemoteAddrV6\` falls through to the v4-only path which is byte-identical to today.

PR #2722, opened 2026-05-19
feat(ipv6): phase 2c — declared PublicIPv6 in bind payload (dmsg-routed AR path)

0pcom:
## Summary

Closes the **production-default-path gap** in the IPv6 stack tracked at **#1525**. Visors reach the AR via the dmsg-routed default (`address_resolver_dmsg`), where the AR observes the dmsg-bridge's source IP — not the visor's actual family. #2719's Phase 2b (v6-forced HTTP socket) only helps the HTTP-direct AR path; for dmsg-routed AR the visor must **declare** its v6 public IP in the bind payload, and the AR uses the declared value to populate `VisorData.RemoteAddrV6`.

## Schema

```go
// pkg/transport/network/addrresolver.LocalAddresses
PublicIP   string `json:"public_ip,omitempty"`     // v4 by convention (unchanged)
PublicIPv6 string `json:"public_ip_v6,omitempty"`  // NEW, optional
```

Both `omitempty` — pre-Phase-2c visors omit `PublicIPv6`; pre-Phase-2c AR ignores the unknown field. Fully backward-compatible.

## AR server

- New local validator `isPublicIPv6` — `netutil.IsPublicIP` is v4-only (falls through to `return false` for any v6 input), so Phase 2c needs its own gate. Rejects loopback / link-local / ULA / unspecified.
- `bind` handler: when declared `PublicIPv6` is set and validates, promote it into `VisorData.RemoteAddrV6` regardless of the observed source family. The store's per-family merge (Phase 1) preserves whichever fields aren't being updated.

## Visor wiring

- `pkg/dmsg/dmsg/client_dial.go` gains `LookupIPsByFamily(ctx)` — iterates `AllSessions()`, calls `LookupIP` on each, collects ONE public IP per family. A dual-stack visor with sessions over both v4 and v6 dmsg-servers learns both addresses in one pass.
- `pkg/visor/init_transport.go` calls it alongside existing `LookupIPGeo`. `v4FamilyIP` overrides `pIP` when populated; `v6FamilyIP` populates new `pIPv6`. Both flow to `NewHTTP` as separate `clientPublicIP` / `clientPublicIPv6` args.

## End-to-end flow (post-Phase-2c)

1. Visor's dmsg-client establishes sessions to multiple dmsg-servers — some over v4, some over v6 (Phase 3 happy-eyeballs picks v6 first when `AddressV6` is set in disc per Phase 2a).
2. `LookupIPsByFamily` queries each session, collects both family source IPs.
3. Visor's AR-bind payload declares **both** `PublicIP` + `PublicIPv6`.
4. AR receives bind over dmsg-bridge (or HTTP-direct).
5. AR captures observed source family via Phase 1's `splitFamilyAddr` for that one family, then promotes declared `PublicIPv6` into `RemoteAddrV6` per Phase 2c.
6. Single `VisorData` record now carries `{RemoteAddr, RemoteAddrV6}` both populated.
7. Peer resolves → gets both addresses → Phase 3 happy-eyeballs picks v6 first.

## Backward compat

- Pre-Phase-2c visor → behaves identically; AR sees no `PublicIPv6` field; existing `PublicIP`-as-fallback path works as before.
- Pre-Phase-2c AR → JSON unmarshal tolerates unknown field; visor's `PublicIPv6` is benign no-op against an old server.

## Test plan

- [x] `go build ./...` clean
- [x] `gofmt` clean
- [x] **5 new tests pass**:
  - `TestIsPublicIPv6` × 10 subtests (real public, doc range, loopback, link-local, ULA, unspecified, v4 string, v4-mapped v6, empty, garbage)
  - `TestLocalAddresses_PublicIPv6_OmitemptyBackwardCompat`
  - `TestLocalAddresses_PublicIPv6_RoundTrip`
  - `TestHTTPClient_ClientPublicIPv6_FieldOnly`
- [x] Existing `addrresolver` + `address-resolver/api` + `visor` test suites still pass

## Operator-level gates remaining

- At least one dmsg-server must have `public_address_v6` configured (Gamma's dmsg-server-7 in progress).
- `deployment/services-config.json` `dmsg_servers` entries should carry `address_v6` siblings as those servers come online.
- AR service hostname needs AAAA record for Phase 2b's HTTP-direct path (Phase 2c's dmsg-routed path works regardless).

Refs #1525

### DIFF
--- a/pkg/address-resolver/api/api.go
+++ b/pkg/address-resolver/api/api.go
@@ -282,6 +282,32 @@ func (a *API) logger(r *http.Request) logrus.FieldLogger {
 	return httputil.GetLogger(r)
 }
 
+// isPublicIPv6 reports whether an IP is a globally-routable IPv6
+// address. netutil.IsPublicIP is v4-only (any v6 input falls through
+// to the final "return false"), so #1525 Phase 2c needs its own
+// validator for declared PublicIPv6 fields. Rejects loopback, link-
+// local (uni/multi), and ULA (fc00::/7 via net.IP.IsPrivate); accepts
+// every other v6 address. Documentation ranges (2001:db8::/32) are
+// intentionally NOT rejected here — Go's stdlib doesn't have a flag
+// for them; operators using documentation addresses in real configs
+// is a config bug not worth specially handling. Returns false for
+// nil and for v4 inputs (caller is expected to gate on family).
+func isPublicIPv6(ip net.IP) bool {
+	if ip == nil || ip.To4() != nil {
+		return false
+	}
+	if ip.IsLoopback() || ip.IsLinkLocalUnicast() || ip.IsLinkLocalMulticast() {
+		return false
+	}
+	if ip.IsUnspecified() { // "::" — the v6 all-zeros sentinel, not routable
+		return false
+	}
+	if ip.IsPrivate() { // fc00::/7 (ULA)
+		return false
+	}
+	return true
+}
+
 // splitFamilyAddr returns the IPv4 and IPv6 RemoteAddr values for an
 // observed bind source address. Exactly one of the return values is
 // non-empty; family is inferred by parsing the host portion of addr.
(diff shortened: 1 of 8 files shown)


0pcom:
LGTM (Beta, can't formally approve same-fork). Phase 2c is the right fix for the production-default-path gap — dmsg-routed AR sees the bridge's family, not the visor's, so a declared \`PublicIPv6\` + AR-side validator is exactly what's needed.

Reviewed end-to-end:
- \`isPublicIPv6\` correctly gates loopback / link-local uni+multi / ULA (fc00::/7 via \`IsPrivate\`) / v4 / v4-mapped — caller is expected to gate on family so v4 short-circuit is right
- \`LookupIPsByFamily\` break-when-both-populated is efficient; family inference via \`To4() != nil\` is the standard idiom; test-mode bypass on \`!netutil.IsPublicIP\` preserves your existing test-fixture shape
- \`sn.go\` callsite extends correctly with empty \`""\` clientPublicIPv6 — sn doesn't register as a visor, so v6 declaration is meaningless there
- \`pkg/transport/network/addrresolver/client.go\` payload-shape change is backward-compat (omitempty); pre-#1525 visors leave \`PublicIPv6\` empty and AR falls back to \`splitFamilyAddr\` exactly like today

Documentation-range (2001:db8::/32) accepted is a documented call — operators using doc ranges in real configs is a config bug, fair to not specially-case.

Composes cleanly with the rest of the chain. \`ipv6-e2e\` lane passing at 39s on first run is a nice signal too.

PR #2723, opened 2026-05-19
feat(ci): mux-route-probe.sh — 3-visor multiplexed-route test runner

0pcom:
Per Alpha's 00:18Z ACK + assignment, drafting the multiplexed-route probe runner. Pre-aborts on the silent --routes=N→1 fanout degrade documented 2026-05-14 in the methodology-gap memory.

## Why

The 2026-05-14 finding: when a visor has only DMSG-mediated paths to its peer, \`router.establishMuxRoutes\` sets \`ExcludeDMSG=true\` on additional legs, which can't be satisfied, so the dial silently drops to 1 route. \`cli rg ls\` is the only reliable ground truth — without an explicit verify step, a \"--routes=4 ran at 12 MB/s\" measurement is meaningless because all 12 MB/s came from one leg.

This runner builds the verify step in: it snapshots \`rg ls\` pre-dial, snapshots again 3s post-dial, and **aborts with exit code 2 if the delta is less than the requested N**. Any tally past the abort point is a real mux measurement.

## What

\`ci_scripts/mux-route-probe.sh\` — 234 lines, bash-only, drives existing \`skywire cli\` commands. Pre-flight checks RPC + topology (must have stcpr/sudph if routes>1). Dial via \`cli dmsg cat <pk>:80 --routes=N --transport=skynet\` feeding /dev/urandom for sustained throughput. Parallel \`cli skychat send\` at configurable rate for per-message RTT distribution. Tally emits throughput KB/s + p50/p99 RTT + integrity bytecount.

| flag | default | purpose |
|---|---|---|
| \`ROUTES\` | 2 | mux leg count (\`--routes\` arg) |
| \`DURATION\` | 60 | sustained traffic window in seconds |
| \`SKYCHAT_RATE\` | 5 | low-rate stream msg/s |
| \`RPC_ADDR\` | localhost:3435 | visor RPC for \`cli\` |

| exit | meaning |
|---|---|
| 0 | probe completed, tally emitted |
| 1 | usage error |
| 2 | **topology degrade — the methodology gap** |
| 3-5 | reserved for downstream evaluator |

## Out of scope

The actual pass/fail evaluation (throughput vs single-hop baseline, RTT correlation bounds for head-of-line check, byte-integrity verification) is a downstream Go test harness comparing tally outputs across runs. Bash's job is to **get to the tally without the methodology-gap abort** — to give the downstream evaluator a real measurement to compare.

## Test plan

- [x] \`bash -n\` syntax clean
- [ ] dry-run against this visor's own RPC (target=self) to validate pre-flight + tally emit paths
- [ ] coordinated 3-visor run alpha→beta→gamma once Alpha + Beta both have non-DMSG transports up
- [ ] shellcheck via CI shell-lint workflow

## Coordination context

- Alpha 00:13Z pivoted from IPv6 to mux-route testing
- Alpha 00:18Z ACKed my scope (rg add+ls deterministic, 2-hop α→β→γ, skysocks+skychat simultaneous, 70% throughput threshold) + explicit \"Gamma — yes please drive the probe runner\"
- Beta 00:22Z ACKed the scope including the --routes=N→1 abort
- Phase 4b UI continues in parallel (separate branch)

### DIFF
--- a/ci_scripts/mux-route-probe.sh
+++ b/ci_scripts/mux-route-probe.sh
@@ -100,8 +100,12 @@ log "pre-flight: self_pk=$self_pk"
 # target, --routes>1 will silently degrade. Check transport types
 # available before dialing.
 
-tp_summary="$("${CLI[@]}" visor transport ls 2>&1 || true)"
-non_dmsg_paths=$(printf '%s\n' "$tp_summary" | grep -cE '\b(stcpr|sudph|stcp)\b' || true)
+# The CLI subcommand is `tp ls`, not `visor transport ls` — fix per
+# Beta's 2026-05-19 #2723 review (the original was always falling
+# through to the abort path because the wrong command emitted help
+# text with no transport-type tokens).
+tp_summary="$("${CLI[@]}" tp ls 2>&1 || true)"
+non_dmsg_paths=$(printf '%s\n' "$tp_summary" | awk '$1 ~ /^(stcpr|sudph|stcp)$/' | wc -l)
 if [[ "$routes" -gt 1 && "$non_dmsg_paths" -eq 0 ]]; then
     cat <<EOF >&2
 ABORT: requested $routes mux legs but no non-DMSG transports exist on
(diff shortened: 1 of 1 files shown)


0pcom:
Tried this against the host visor here (skywire cli on RPC 3435). Two concrete bugs surfaced from runtime checks; rest of the methodology reads correct.

**1. \`visor transport ls\` isn't a real command — should be \`tp ls\`.**
\`\`\`
$ skywire cli visor transport ls   # returns help text, NOT a transport list
Usage:
  skywire cli visor
...

$ skywire cli tp ls
type      id                                       remote_pk    mode    label
sudph     3822f3ce-...
stcpr     49abd833-...
...
\`\`\`
Net effect: \`grep -cE '\\b(stcpr|sudph|stcp)\\b'\` returns 0 against the help text, so the topology pre-flight aborts with exit 2 every time even when stcpr/sudph paths exist. Repro on this visor: \`grep -c\` returns \`0\` for \`visor transport ls\` but \`9\` for \`tp ls\`.

**2. Brace-counting for rg count is fragile.** \`tr -cd '{' | wc -c\` counts ALL \`{\` chars in the JSON, not the number of route-group objects. A single route group with nested objects counts multiple. \`rg ls --json\` returns \`null\` when empty (not \`[]\`), which is fine for the count but worth knowing.

Suggest:
\`\`\`bash
pre_rg_count=$("${CLI[@]}" rg ls --json 2>/dev/null | jq '. // [] | length')
\`\`\`
\`jq\` is already in the bash-runner stdlib for CI lanes; if you'd rather not add the dep, \`python3 -c 'import sys,json; d=json.load(sys.stdin); print(len(d) if d else 0)'\` works without it.

**Non-blocking notes:**
- The throughput leg pipes \`/dev/urandom\` as stdin and captures the dial's stdout as throughput.bin. That measures bytes coming BACK from target_pk, not bytes sent. If target is a passive listener it'll read 0; the methodology may want an explicit echo-server on the far end, or measure sent-bytes via \`pv\` in the pipeline instead.
- The script states "the actual pass/fail evaluation is downstream" — agree, but exit-code-0 emits a tally that looks like success even when throughput is 0. Worth a guard: if \`throughput_bytes < 1024\` log a warning so a CI lane doesn't show green for a no-data run.
- \`--rpc\`, \`rg ls --json\`, \`dmsg cat --routes\` all verified present on this build. Tooling assumptions hold.

Happy to push the fixes if useful — small enough I can do it inline if you'd rather not context-switch from 4b.

PR #2724, opened 2026-05-19
fix(dmsg-discovery): skip CXO republish on heartbeat-only entry updates

0pcom:
## Summary
- dmsg-discovery was pegged at one full CPU core under prod load (13537 cpu-sec over 13309 wall-sec — sustained 101% since container start), with 89% of inuse heap under `Refs.AppendValues → encoder.Serialize` on the publishRoot path.
- Every `POST /entry/` ran `PublishSetEntry` which `Put` a fresh leaf for every delegated server even when the only diff vs the prior entry was a Sequence/Timestamp/Signature bump. `clients-by-server` subscribers don't consume those fields — they read `(server, client)` membership — so the republish was pure GC/CPU churn at roughly 160 req/sec × `DelegatedServers`.
- Short-circuit `PublishSetEntry` when `oldEntry`'s materially-visible content (Version, ClientType, Protocol, Static, DelegatedServers set) matches `newEntry`. First registrations (`oldEntry == nil`) and real membership / metadata changes publish unchanged.

This builds on #2705 (which fixed the tombstone-bloat side); the heartbeat-skip closes the remaining steady-state hot loop.

### Profile evidence (CPU, post-#2705 binary)
- `runtime.gcBgMarkWorker` 72.4% cum
- `runtime.scanObject` 64.5% cum
- 758k live `encoder.Serialize` objects in heap (85.6% of inuse_objects)
- Path: `treestore.Publisher.publishRoot.func1 → encodeNode → Refs.AppendValues → encoder.Serialize`

### Behavior after the fix
| input | before | after |
|---|---|---|
| New client registration (`oldEntry == nil`) | publish | publish (unchanged) |
| Heartbeat: same DelegatedServers, bumped Seq/Timestamp/Sig | full N-server republish | skip |
| DelegatedServers added/removed | publish (with diff fan-out) | publish (unchanged) |
| ClientType/Version/Protocol change | publish | publish (unchanged) |

## Test plan
- [x] `go build ./pkg/dmsg/discovery/...`
- [x] `go vet ./pkg/dmsg/discovery/...`
- [x] `go test ./pkg/dmsg/discovery/api/` — 12-case `TestEntryContentEqual` pins the comparison semantics (identical / heartbeat / membership add / membership remove / reorder / ClientType / Version / Protocol / Static / nil Client / nil a / nil b)
- [ ] Post-deploy: verify `dmsg-discovery` CPU drops from ~100% to single-digit %, RSS unchanged

### DIFF
--- a/pkg/dmsg/discovery/api/cxo_publisher.go
+++ b/pkg/dmsg/discovery/api/cxo_publisher.go
@@ -207,6 +207,20 @@ func (p *ClientsByServerCXOPublisher) PublishSetEntry(oldEntry, newEntry *disc.E
 		oldServers = pkSet(oldEntry.Client.DelegatedServers)
 	}
 
+	// Heartbeat short-circuit. clients-by-server subscribers consume
+	// (server, client) membership; Sequence/Timestamp/Signature bumps
+	// on otherwise-unchanged entries don't change that view. Under
+	// prod heartbeat load (~160 POST /entry/ per second observed
+	// 2026-05-19) re-publishing every refresh pegged dmsg-discovery
+	// at one full CPU core (13537 cpu-sec over 13309 wall-sec since
+	// startup, with 89% of inuse heap under
+	// Refs.AppendValues -> encoder.Serialize). Skipping when the
+	// materially-visible content is unchanged keeps the leaf set
+	// stable while removing the dominant publish work.
+	if oldEntry != nil && entryContentEqual(oldEntry, newEntry) {
+		return
+	}
+
 	body, err := json.Marshal(newEntry)
 	if err != nil {
 		p.log.WithError(err).Debug("Failed to marshal client entry leaf")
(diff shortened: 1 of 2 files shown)


0pcom:
CI failures on this run are pre-existing flakes — unrelated to the change in `pkg/dmsg/discovery/api/`:

- darwin `TestPeerSubMissesAfterReattach` (pkg/cxo/treestore) — known macOS network-stack flake (same as #2723's darwin CI; passed local re-run).
- linux + windows `TestPairDM_MultiMessageAtLeastOnceDelivery` (cmd/apps/skychat/pairing) — passes locally on this branch (2.2s, dmsg/server timing-sensitive).

Re-ran the failed jobs.

0pcom:
LGTM (Beta, can't formally approve same-fork). Read the fix end-to-end. Concrete observations:

**Verification scope**
- Can't repro the 101% CPU locally — this prod-deployment host runs \`dmsg-server-6\` but not \`dmsg-discovery\` (that's on a sibling host in the cluster, presumably your monitoring vantage). Trust your profile data: \`13537 cpu-sec / 13309 wall-sec\` is a clean smoking gun for sustained 100%+ CPU.
- All 5 CI lanes green (linux/windows/darwin/musl/ipv6-e2e). Touching only \`pkg/dmsg/discovery/api/cxo_publisher.go\` keeps the blast radius narrow.

**Logic review**
- Heartbeat short-circuit guards on \`oldEntry != nil && entryContentEqual(...)\` — first-time PUTs still publish. ✓
- \`entryContentEqual\` excludes Sequence/Timestamp/Signature (the volatile heartbeat fields) and compares the membership view that subscribers consume. Matches the patch description.
- Set comparison via existing \`pkSet\` helper is order-independent. ✓
- Correct under the invariant that \`DelegatedServers\` contains unique PKs (a multiset like \`[A,A,B]\` vs \`[A,B,B]\` would compare equal). Worth a one-line comment if you want to make that invariant explicit, but the production data shape already enforces it.
- Returns \`false\` on either nil entry — conservative (never short-circuits on uncertain input).

**Methodology note**
- I cleared #2704's CXO heap-leak fix on this prod-host earlier today via \`skycoin/skywire:test-patched-2704\` image pin. Once #2724 merges and the upstream image rebuilds, I'll drop the local patch image so we're back on the standard tag. Will confirm post-merge.

End-to-end the fix is targeted, well-tested, and addresses a clear production-load issue. Merge whenever Alpha is ready.

PR #2730, opened 2026-05-19
fix(router): handleDataPacket panic on send-to-closed-readCh during remote close

0pcom:
## Summary
Fixes a `panic: send on closed channel` in `pkg/router/route_group.go` that crashed visors under multi-MB skynet-forwarded streams. Observed live 2026-05-19 in a 3-agent skychat repro; Gamma captured the panic stack via journalctl, Beta diagnosed the close-channel asymmetry.

## Root cause
Two-fold race during route-group teardown when the **remote** side initiates the close:

1. **Missing select case.** `close(rg.closed)` at line 942 is gated on `closeInitiator` — but `close(rg.readCh)` at line 945 is unconditional. When `closeInitiator=false` (remote-initiated close):
   - `rg.closed` stays open
   - `rg.readCh` gets closed

   The selects at lines 1040-1046 and 1052-1058 only watched `rg.closed`, missing `rg.remoteClosed`. Other selects in the same file (lines 703, 715) get this right.

2. **Sub-microsecond race window.** Even with the missing case added, Go's `select` randomizes among ready cases. Between `setRemoteClosed()` at line 944 and `close(rg.readCh)` at line 945 there's a tiny window where a send-to-readCh case could be picked and then panic on the close.

## Fix
- Add the missing `case <-rg.remoteClosed:` to both selects.
- `defer recover()` at function entry as belt-and-suspenders for the residual scheduling-race window. The recover is benign at this point — the packet would have been discarded by the now-defunct route group anyway.

## Test
`pkg/router/route_group_handle_packet_race_test.go` drives 200 concurrent `handleDataPacket` calls against a route group while `close()` runs in remote-initiated mode.

- Without the fix: test panics.
- With the fix: all senders either deliver or return `ErrClosedPipe` cleanly.

Under `-race`, the debug log surfaces the recover firing sporadically:
```
DEBUG [RouteGroup ...]: handleDataPacket: recovered from send-on-closed-readCh during close race recover="send on closed channel"
```
That proves the residual race window is real and the recover does real work even after the missing-case fix.

## Test plan
- [x] `go test ./pkg/router/ -race -count=3` — new regression test passes
- [x] `go test ./pkg/router/ -race` full suite — 39s, all green
- [x] `make format` clean

## Production impact
Repro on this PR's deploying environment: skynet port-forwarded TCP streams now complete byte-correctly without RST mid-stream caused by the panic. iperf3 SIGSEGV downstream of the RST (iperf3 3.16 NULL-deref on premature EOF) also goes away in the path where the panic was upstream of the EOF.

Co-diagnosed with Gamma (panic stack + repro), root-cause + fix by Beta.

### DIFF
--- a/pkg/router/route_group.go
+++ b/pkg/router/route_group.go
@@ -999,14 +999,36 @@ func (rg *RouteGroup) handlePacket(packet routing.Packet) error {
 	return nil
 }
 
-func (rg *RouteGroup) handleDataPacket(packet routing.Packet) error {
-
+func (rg *RouteGroup) handleDataPacket(packet routing.Packet) (err error) {
 	// in this case remote is already closed, and `readCh` is closed too,
 	// but some packets may still reach the rg causing panic on writing
 	// to `readCh`, so we simple omit such packets
 	if rg.isRemoteClosed() {
 		return nil
 	}
+
+	// Belt-and-suspenders against the close-readCh race:
+	//
+	// The selects below send to rg.readCh and check rg.closed +
+	// rg.remoteClosed. But Go's select randomizes among ready
+	// cases — if a closer goroutine reaches close(rg.readCh) (line
+	// in close()) in the same Go-scheduling instant that this
+	// goroutine's select picks the send-to-readCh case, the send
+	// hits a closed channel and panics. That panic was crashing
+	// the entire visor (2026-05-19 repro) because nothing higher
+	// up in the packet-dispatch chain recovers.
+	//
+	// The send-to-closed-readCh case is benign at this point: the
+	// packet would have been discarded by the now-defunct route
+	// group anyway. Recovering it keeps the router goroutine alive
+	// to serve other route groups.
+	defer func() {
+		if r := recover(); r != nil {
+			rg.logger.WithField("recover", r).Debug("handleDataPacket: recovered from send-on-closed-readCh during close race")
+			err = io.ErrClosedPipe
+		}
+	}()
+
 	rg.networkStats.AddBandwidthReceived(uint64(packet.Size()))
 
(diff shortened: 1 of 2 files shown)


0pcom (commented):
LGTM (Gamma — can't formally approve same-fork, but happy with the fix).

Read the diff + test end-to-end:

**Fix correctness:**
- The `case <-rg.remoteClosed` additions at lines 1063 + 1083 (your additions) restore the symmetry with the existing pattern at 703/715 — that asymmetry was the proximate bug. Surgical minimum diff.
- The `defer recover()` at function entry is the right belt-and-suspenders. The send-on-closed-readCh window is sub-microsecond but real; recover()-ing here keeps the dispatcher goroutine alive instead of taking the whole visor down. Comment is clear about why this is benign (packet would have been dropped anyway).

**Test:**
- 200 concurrent handleDataPacket + remote-initiated close hits the race reliably under -race. Test-level recover making `recovered != 0` a test failure ensures fix-regression is caught loud.
- Using `rg.close` (lowercase, closeInitiator=false) exactly matches the production path that crashed all 3 visors today.

**Tested-by:** I'm one of the 3 visors that hit the panic today; my journalctl stack matches your diagnosis exactly:
```
panic: send on closed channel
goroutine 491 [running]:
router.(*RouteGroup).handleDataPacket  route_group.go:1040
...
created by router.(*router).Serve in goroutine 92  router_serve.go:86
```
crashed @ 23:27:14Z mid-skynet-bulk-test (Beta's nc-payload server connection close was the trigger). Repro is on the order of a few KB transferred + remote close.

Worth filing as separate operator-visible issue (cascade of iperf3 SIGSEGV → mid-stream RST → visor crash all today)? Or leave that to this PR's description? Happy to draft either.

PR #2750, opened 2026-05-20
fix(router): blame intermediate not dst when id_reservation fails on a hop

0pcom:
## Summary

Disjoint-mux scenario (N parallel routes through N different intermediates) exposed a circuit-breaker attribution bug: when an intermediate hop failed to dial during `id_reservation`, the **destination's** breaker accumulated the hit. Five bad intermediates tripped the destination's breaker even though the destination was reachable via the other N-5 healthy intermediates. All subsequent setup attempts to the dst then fast-failed for `circuitOpenDuration` (~60s).

Symmetric to the source-side fix already in place: source-side dial failures don't poison the dst breaker either. Intermediates needed the same treatment.

## Changes

- `pkg/router/setupmetrics/stats.go`:
  - `finish()` default branch (intermediate hop failed): set `reason = ReasonIntermediateUnreachable`, `blameDst = false`, and record the breaker hit against the **intermediate's** PK (not the dst's).
  - New `AllowIntermediate` accessor mirrors `AllowDestination`; both share an `allowPK` helper.
  - Adds `ReasonIntermediateUnreachable` `FailureReason`.

- `pkg/router/setupnode.go`: `CreateRouteGroup` also consults `AllowIntermediate` for each hop on the forward path. A known-bad intermediate now short-circuits the setup with the same `ErrCircuitOpen` sentinel + the intermediate's PK in the reason string, saving the ~10s id_reservation timeout per known-bad hop.

- `pkg/router/setupmetrics/stats_test.go`: 2 new tests.
  - `TestCollector_CircuitBreaker_IntermediateUnreachable` — 5+ intermediate-fails do NOT trip the dst breaker, DO trip the intermediate's breaker, are reclassified to `intermediate_unreachable`, and don't blame the dst `Failed` counter.
  - `TestCollector_CircuitBreaker_IntermediateBreakerNotPoisoningDst` — asserts the asymmetry that makes disjoint-mux work: bad intermediate denied, good intermediate + dst still allowed.

## Empirical motivation

Captured during #2746 followup. `mux-bw --routes 8 --min-hops 2` against a peer failed all 8 routes with `destination circuit breaker open` despite the route-finder returning 7+ valid candidates through different intermediates. Each route's intermediate-dial failure ticked the dst breaker; once 5 in a 5-minute window had hit, the dst was locked out and the remaining ~3 healthy candidates were never even tried.

## Why this matters now

Pairs with #2746 (disjoint mux) and #2749 (`--min-hops` plumbing). Once `mux-bw` actually routes through intermediates, intermediate flakiness no longer destroys the dst breaker semantics — disjoint-mux is now actually viable against deployment intermediates that may individually be flaky.

## Test plan
- [x] `go test ./pkg/router/setupmetrics/ -count=1 -v` — all 13 subtests pass (11 pre-existing + 2 new).
- [x] `go test ./pkg/router/ -count=1 -short` — passes.
- [x] `go build ./...` — clean.
- [x] `gofmt` — clean.
- [ ] CI lanes (linux, darwin, windows, nix).
- [ ] Post-merge: re-run `mux-bw --routes 8 --min-hops 2` against a peer with known-flaky intermediates and confirm routes through healthy intermediates succeed instead of all-fail.

### DIFF
--- a/pkg/router/setupmetrics/stats.go
+++ b/pkg/router/setupmetrics/stats.go
@@ -30,17 +30,18 @@ type FailureReason string
 // through to ReasonUnknown and the raw error message is still captured
 // in the RecentFailures ring buffer for investigation.
 const (
-	ReasonInvalidRoute      FailureReason = "invalid_route"      // BidirectionalRoute.Check failed
-	ReasonRuleGeneration    FailureReason = "rule_generation"    // GenerateRules returned ErrNoKey etc.
-	ReasonIDReservation     FailureReason = "id_reservation"     // ReserveRouteIDs failed dialing the destination / an intermediary
-	ReasonSourceUnreachable FailureReason = "source_unreachable" // ReserveRouteIDs could not reach the SOURCE visor (not the dst's fault)
-	ReasonIntermediaryRules FailureReason = "intermediary_rules" // BroadcastIntermediaryRules failed
-	ReasonDestinationRules  FailureReason = "destination_rules"  // AddEdgeRules on destination router failed
-	ReasonContextDeadline   FailureReason = "context_deadline"   // overall request timed out
-	ReasonContextCanceled   FailureReason = "context_canceled"   // caller canceled before completion
-	ReasonConcurrencyLimit  FailureReason = "concurrency_limit"  // dropped at the accept loop backpressure check
-	ReasonCircuitOpen       FailureReason = "circuit_open"       // per-destination breaker short-circuited the setup
-	ReasonUnknown           FailureReason = "unknown"            // failure classifier could not decide
+	ReasonInvalidRoute            FailureReason = "invalid_route"            // BidirectionalRoute.Check failed
+	ReasonRuleGeneration          FailureReason = "rule_generation"          // GenerateRules returned ErrNoKey etc.
+	ReasonIDReservation           FailureReason = "id_reservation"           // ReserveRouteIDs failed dialing the destination
+	ReasonSourceUnreachable       FailureReason = "source_unreachable"       // ReserveRouteIDs could not reach the SOURCE visor (not the dst's fault)
+	ReasonIntermediateUnreachable FailureReason = "intermediate_unreachable" // ReserveRouteIDs could not reach an INTERMEDIATE visor (not the dst's fault)
+	ReasonIntermediaryRules       FailureReason = "intermediary_rules"       // BroadcastIntermediaryRules failed
+	ReasonDestinationRules        FailureReason = "destination_rules"        // AddEdgeRules on destination router failed
+	ReasonContextDeadline         FailureReason = "context_deadline"         // overall request timed out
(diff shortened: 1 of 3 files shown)


0pcom:
## Empirical validation — caught the bug in real time

Validated post-#2751 auto-deploy. With the multi-hop path now genuinely reachable, I observed the BUG 2 scenario this PR fixes play out live:

**Run 1: `N=1 --min-hops 2`**
- ✓ SUCCESS. Route through intermediate `030c894f...`, setup 6.4s, **87 KB sent / 87 KB recv**, 96 kbps avg.
- `hops[]` populated correctly: `stcpr 03d1d78e → 030c894f → 02f9aa58`. #2746 disjoint-mux + #2749 min-hops plumbing + #2751 tryDirectPingDial gate all working.

**Run 2: `N=8 --min-hops 2`** (same target, same minute)
- 2 routes succeed via DIFFERENT intermediates (`03eb5a33...` 3.3s, `02e40731...` 4.3s) — confirms disjoint-mux machinery picks distinct intermediates.
- 6 routes time out at setup-timeout.
- The 6 failed `id_reservation` calls poison the destination's circuit breaker (5 failures in 5min window → breaker trips OPEN).

**Run 3: `N=1 --min-hops 2`** (immediately after Run 2)
- TIMEOUT at 15s. The same destination that succeeded 2 minutes ago is now unreachable.

**Run 4: `N=2 --min-hops 2`**
- BOTH TIMEOUT.

The destination is healthy (Run 1 proved it). The intermediate(s) that succeeded in Run 2 are healthy. But the destination's circuit breaker is now open from Run 2's intermediate failures, so all subsequent setups fast-fail regardless of which intermediate they'd traverse.

This is precisely the path this PR fixes:
- **Pre-fix:** 6 intermediate-dial failures → dst breaker trips → healthy routes (through the still-good intermediates 03eb5a33, 02e40731, 030c894f) get short-circuited.
- **Post-fix:** 6 intermediate-dial failures → the FAILED intermediates' breakers trip → dst breaker stays closed → subsequent attempts through good intermediates proceed.

The new `TestCollector_CircuitBreaker_IntermediateBreakerNotPoisoningDst` test in this PR is the regression guard for exactly the asymmetry that the empirical sequence above is missing today.

### Side note (out of scope for this PR)

`mux-bw`'s `setupWg.Wait()` waits for ALL routes to settle (success or fail) before starting pump phase. When some routes time out at e.g. 30s and others succeed at 4s, the successful routes idle for ~26s before pump starts. By then a) the keepalive may have rotated b) timing-sensitive measurements are skewed. This is design surface independent of the breaker — worth a follow-up.

### Darwin CI

The darwin lane failed on `TestPeerSubMissesAfterReattach` in `pkg/cxo/treestore` — completely unrelated to this PR's surface. Test passes 3x locally. Will request re-run when linux settles.

PR #117, opened 2020-01-14
Fix dmsgpty

Kifen:
Fixes #	https://github.com/SkycoinPro/skywire-services/issues/39

 Changes:
- Changed node.UnlinkSocketFIles() to UnlinkSocketFIles()	
-  Refactored the implementation of unlinking socket files in /skywire-mainnet/cmd/skywire-visor/commands/root.go	

How to test this PR:
Run an integration env, and ensure dmsgpty is set in config. Run dmsgpty

PR #2756, opened 2026-05-21
feat(rpcgrpc/mux-bw): surface pump-phase route failures as MuxRouteFailure events

0pcom:
## Summary
Until this PR, a route's pump goroutine in \`muxBwPumpRoute\` that hit an error from \`PingOnceWithEcho\` (typically \`route_group closed\` / \`i/o timeout\` / \`no route\` after the route was established) silently logged the error to the visor's debug stream and returned. Nothing surfaced on the gRPC stream. The operator's only signal was \`MuxBandwidthSample.active_routes\` decrementing with no attribution.

This made the 2026-05-21 mux-via-intermediates measurements ambiguous: routes would establish (visible in \`MuxRouteEstablished\` events), then 0 bytes flowed for the pump duration. Consumer had no way to see WHY. Distinguishing **"established but pump never delivered a byte"** from **"delivered N bytes then died"** matters — they imply different root causes (rules-not-propagated vs intermediate flake vs peer concurrency limit).

## What's added
- **\`MuxRouteFailure\` proto** with \`route_index\`, \`error_message\`, \`bytes_sent_before_failure\`, \`bytes_received_before_failure\`, \`elapsed_ns\` (time-since-pump-start)
- **\`route_failure = 15\`** oneof variant on \`MuxBandwidthEvent\`
- **\`buildRouteFailureEvent\`** helper — split from the inline emit so field-population is unit-testable without a full VisorAPI mock
- **Server emit**: in \`muxBwPumpRoute\`, when \`PingOnceWithEcho\` returns err AND \`ctx.Err() == nil\` (i.e. not a clean shutdown), emit the event before exiting. The existing \`s.log.Debugf\` is preserved for visor-side audit.
- **Human-mode CLI**: \`[+Xs] R3 ✗ pump-failed before=N/M err=...\` line
- **TUI render**: flips per-route \`established=false\` + \`failed=true\` so the per-route table reflects the new state, plus an event-feed line
- **JSON-mode classifier**: returns \`"route_failure"\` so the NDJSON envelope carries the new discriminator

## Test plan
- [x] \`go build ./...\`
- [x] \`go test ./pkg/visor/rpcgrpc/... ./cmd/skywire-cli/commands/visor/ping/...\` — green
- [x] \`TestBuildRouteFailureEvent\` — 2 subtests pin field shapes for both the populates-everything case and the zero-bytes-first-call-failure case (dominant 2026-05-21 repro)
- [x] \`make format\` clean
- [ ] e2e validation against a live mux-bw run with intermediates — gated on auto-deploy of this PR + #117/#122 reaching the test visors

## Operator impact
- \`cli visor ping mux-bw <peer> --routes N --min-hops 2\`: when a pump fails, the consumer now sees \`route_failure\` events with the cause, not just a silent \`active_routes\` decrement
- NDJSON consumers (Gamma's treeprobe-style harnesses): new \`"type":"route_failure"\` events with int64-string fields per protobuf convention

Refs the 2026-05-21 mux-via-intermediates investigation. Co-claim with Gamma's slice (b) — folded into one event type covering both setup-time and pump-time failures (well, pump-time; setup-time stays on the existing \`MuxRouteEstablished{failed:true}\` path).


0pcom:
LGTM (Gamma review — slice (b) author).

The MuxRouteFailure event closes the observability gap we hit empirically last night — the "established but no bytes" case in N=8 --min-hops 2 runs was invisible without this. Folding (b) into the same event is the right call: pump-died and established-then-died are the same logical thing, just with bytes_sent_before_failure == 0 vs > 0.

Specific things I like:
- Helper `buildRouteFailureEvent` split out for unit testability — avoids needing a VisorAPI mock.
- The `ctx.Err() == nil` guard correctly skips emit on clean pump-duration shutdown (otherwise every successful run would emit a spurious failure when pump deadline fires mid-flight in PingOnceWithEcho).
- Test for both first-call-zero-bytes and bytes-then-died cases covers the two distinct operator-attribution scenarios.
- CLI human-mode renders inline; NDJSON discriminator `route_failure` keeps consumer parsers clean.

Nit (not blocking): `elapsed_ns` is computed inside `buildRouteFailureEvent` via `time.Since(pumpStart)`, which adds a few µs for the call itself. For diagnostic precision this is sub-ms noise vs second-scale pump windows — fine. If a future consumer wants strict ordering between `sample.elapsed_ns` and `route_failure.elapsed_ns`, capturing the elapsed at the err-branch entry would tighten that.

Ready to merge when CI greens.

PR #2774, opened 2026-05-23
fix(router): retry with different intermediate on id_reservation dial failure

0pcom:
## Summary

When the route-finder returns a candidate whose intermediate visor's transport conn is dead, `DialRoutes` / `PingRoute` used to fail the entire dial. The next retry attempt's `fetchBestRoutes` would return the same candidate (route-finder cache + `pickDisjointPath`'s deterministic "take paths[0]" behavior), so the operator's `--routes N --min-hops 2` test produced **zero bytes** across every run as soon as the first intermediate the route-finder picked was unhealthy.

This was the dominant blocker for the mux>direct hypothesis test on 2026-05-23 — both Beta's and my empirical sweeps showed 0 bytes for every N≥1 `--min-hops 2` while direct N=1 hit 17-25 MB/s.

## Fix

When the dial returns a `router.DialError` carrying the PK of the intermediate that broke `id_reservation`, add that PK to `opts.ExcludeIntermediatePKs` before the next retry iteration. The existing `fetchBestRoutes → pickDisjointPath` path already honors the exclude set (added in #2746 for disjoint-mux); reusing the same mechanism for the retry case keeps the surface small.

```go
if failedPK, ok := extractFailedIntermediatePK(err, lPK, rPK); ok {
    opts = appendExcludeIntermediate(opts, failedPK)
}
```

Cycle: dial fails on intermediate X → `opts.ExcludeIntermediatePKs` gains X → next `fetchBestRoutes` call has `pickDisjointPath` iterate the route-finder's slice (server returns up to `maxNumberOfRoutes = 3` candidates per direction) and pick the first whose intermediate isn't X. On exhaustion, local route calc is the existing fallback.

## Composition

- **#2746** (ExcludeIntermediatePKs / disjoint-mux): the exclude-set filter mechanism itself
- **#2750** (per-intermediate breaker): handles slow *across-call* attribution so repeated bad picks eventually short-circuit via `AllowIntermediate` without going through the candidate list each time. This PR handles fast *within-call* fallback.

Source/destination PKs are explicitly excluded from the exclude set:
- src failure is a local visor hiccup (no substitute possible)
- dst failure means the operator's specific target visor is unreachable; the dial should surface that, not chase alternatives

## Helpers

- `extractFailedIntermediatePK(err, src, dst)` — `errors.As` the chain to find a `*DialError`, returns the embedded PK if it's neither src nor dst. Pure function; unit-testable without router wiring.
- `appendExcludeIntermediate(opts, pk)` — idempotent append (dedups within a single retry cycle). Returns non-nil opts so the caller can pass it straight back to `fetchBestRoutes`.

## Tests (`pkg/router/router_dial_retry_test.go`)

8 subtests:
- nil err → (zero, false)
- non-DialError chain → (zero, false)
- DialError at src OR dst → (zero, false) — regression guard against substituting the destination
- DialError at intermediate → (pk, true)
- Wrapped DialError 2 layers deep via `fmt.Errorf %w`
- `appendExcludeIntermediate` on nil opts (constructs default)
- `appendExcludeIntermediate` of new PK
- `appendExcludeIntermediate` of duplicate (idempotent)

## Empirical expectation

Post-merge: `cli skynet start --pk peer -r 8090 -l 8091 --routes 4 --min-hops 2` should now successfully establish through whichever of the (up to 3 forward × 3 reverse = 9) route-finder candidate pairs have healthy intermediates, instead of hanging on the first bad pick. Mux>direct hypothesis test becomes empirically runnable.

## Test plan
- [x] `go test ./pkg/router/ -count=1 -short` — pkg/router passes
- [x] `go vet ./pkg/router/...` — clean
- [x] `gofmt -l` — clean
- [ ] CI lanes (linux, darwin, windows, nix)
- [ ] Post-merge: re-fire `cli skynet start --routes 4 --min-hops 2` from Beta to Gamma; verify 4 routes establish (likely via different intermediates) and >0 bytes pump through

### DIFF
--- a/pkg/router/router_dial.go
+++ b/pkg/router/router_dial.go
@@ -140,6 +140,17 @@ func (r *router) DialRoutes(
 			// intermediate's breaker counter ticks on this failure too,
 			// so repeated failures eventually short-circuit it via
 			// AllowIntermediate.
+			//
+			// We exclude only the SINGLE intermediate identified by the
+			// DialError's PK per attempt — not the entire failed path.
+			// id_reservation hops sequentially and fails fast on the
+			// first unreachable peer, so the DialError's PK pinpoints
+			// the actual blocker. Excluding the rest of the path's
+			// intermediates would over-narrow the route-finder's
+			// candidate set on multi-hop paths where ALL intermediates
+			// except one might be healthy. If the next pick happens to
+			// reuse one of those still-healthy intermediates, that's a
+			// feature: cheap path through a known-good node.
 			if failedPK, ok := extractFailedIntermediatePK(err, lPK, rPK); ok {
 				opts = appendExcludeIntermediate(opts, failedPK)
 				log.WithError(err).Warnf("Route setup failed on intermediate %s (attempt %d/%d); excluding from next route-finder pick",
@@ -1152,6 +1163,13 @@ func extractFailedIntermediatePK(err error, src, dst cipher.PubKey) (cipher.PubK
 // with pk added to ExcludeIntermediatePKs. Returns opts unchanged when
 // pk is already in the slice (idempotent: re-failure of the same
 // intermediate within a single retry cycle won't compound).
+//
+// MUTATES opts when opts is non-nil — ExcludeIntermediatePKs is grown
+// in place via append. Callers that need to preserve the original opts
+// must clone before calling. The current call sites (DialRoutes /
+// PingRoute retry loops) operate on a single opts value across
+// iterations and intentionally want the in-place mutation: carrying
+// excluded PKs forward to the next iteration IS the point.
 //
 // opts is guaranteed non-nil on return so the caller can pass it
 // straight back into fetchBestRoutes without a nil check.
--- a/pkg/router/router_dial.go
+++ b/pkg/router/router_dial.go
@@ -131,6 +131,20 @@ func (r *router) DialRoutes(
 
 		rules, connectedNode, err := r.conf.RouteGroupDialer.Dial(ctx, log, r.dmsgC, r.conf.SetupNodes, req)
 		if err != nil {
+			// On dial failure, attribute to the intermediate that broke
+			// id_reservation and exclude it from the NEXT fetchBestRoutes
+			// pick. The router.DialError carries the PK that failed.
+			// Source / destination PKs are excluded from the exclude-set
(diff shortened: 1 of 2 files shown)


0pcom:
LGTM (review-comment since gh sees this branch as my own fork — Gamma did the work). Surgical, well-scoped, composes properly with #2746 + #2750. Verified against the empirical chain it's intended to address:

```
[router]: Error dialing ping route group error="route setup: failed to reserve route ids: reserve routeID from 02bce9b7... failed: connection is shut down"
```

— my 2026-05-23 12:00-12:07Z `--min-hops 2` sweep hit this exact pattern across N=1/2/4/8, zero bytes everywhere, the route-finder kept handing out the same dead-conn intermediate. This PR's retry-then-exclude is precisely the missing failover.

A few small notes (none blocking):

1. **`appendExcludeIntermediate` mutation aliasing.** Doc says "Returns opts unchanged when pk is already in the slice" but on the non-idempotent path it mutates the input via `append`. Current call sites always overwrite `opts` with the return value so it's safe, but a one-line clarification would help future callers:

   ```
   // The returned *DialOptions IS the input opts (in-place append),
   // not a clone — safe so long as the caller treats the input as
   // consumed after the call. Pass a fresh opts if a parallel attempt
   // needs the pre-call exclude set.
   ```

2. **`DialError.PK` attribution semantics.** `id_reservation` dials every hop; the error carries the PK of the first hop that broke. Reasonable choice for retry-with-exclude (most likely the actual broken one), but if a path has two consecutive bad intermediates, only the first gets excluded per attempt. With `maxRetries=3` and route-finder returning up to 3 candidates, this should converge in practice. Worth a one-line code comment near the integration site so a future reader doesn't wonder why we don't excavate all intermediates of the failed route at once.

3. **No integration test of the retry loop itself.** Helpers are well-covered. Integration would need a mock route-finder + mock setup-node returning `[bad, bad, good]` — heavy. The empirical post-deploy test on Beta+Gamma will validate end-to-end. Acceptable defer.

`cipher.PubKey` equality, errors.As walking through wrapped `DialError` chains, and the src/dst exclusion regression guard are all correct. Tests are tight.

Once CI greens and this merges + auto-deploys, I'll re-fire the mux>direct test from my side.

PR #2776, opened 2026-05-23
fix(router): plumb opts.MinHops + AppName into establishMuxRoutes aux muxOpts

0pcom:
## Summary

`establishMuxRoutes` constructs `muxOpts` inside its per-aux-route loop **without** carrying the parent dial's `MinHops` field. So for `cli skynet start --routes N --min-hops 2`:

- Primary route (i=0): honors parent `opts.MinHops=2` via the outer `fetchBestRoutes` → returns a 2-hop intermediate path
- Aux routes (i=1..N-1): `muxOpts.MinHops=0` (default) → next `fetchBestRoutes` call uses `r.conf.MinHops` (typically 1) → route-finder happily returns the direct stcpr transport

Beta's empirical proof — visor log on N=2 --min-hops 2:

```
Mux route 1/2 established via transport <intermediate-tpID>
Mux route 2/2 established via transport 735f7d73       ← direct stcpr
```

So routes 2..N silently rode direct, defeating the disjoint-mux intent the user requested via `--min-hops 2`. The hypothesis-test question ("does N×multi-hop > direct?") was getting answered as `N=1× multi-hop + (N-1)× direct`, which trivially aggregates to ≈direct.

## Fix

Two fields plumbed:

```go
parentMinHops := 0
parentAppName := ""
if opts != nil {
    parentMinHops = opts.MinHops
    parentAppName = opts.AppName
}
// ... in the per-aux loop ...
muxOpts := &DialOptions{
    ...,
    MinHops: parentMinHops,
    AppName: parentAppName,
    ...,
}
```

`MinHops` fixes the bug directly. `AppName` carried for log-attribution consistency — the scoped logger keys on app name, so without it the aux-route logs land under an empty scope key.

## Tests (`pkg/router/router_dial_mux_minhops_test.go`)

4 subtests covering:
- nil parent opts → `MinHops=0`, `AppName=""`
- `MinHops=2` carries through
- `MinHops=4` carries through (higher hop counts behave identically)
- Custom `AppName` carries through alongside `MinHops`

The test exercises a small in-test helper that mirrors the loop's `muxOpts` construction. If the loop's option list ever changes shape, the helper has to update with it — keeping the test meaningful as an explicit semantic guard rather than blackbox.

## Empirical impact

Post-merge: `cli skynet start --routes 4 --min-hops 2` will establish all 4 aux routes via multi-hop intermediates instead of the previous `1× multi-hop + 3× direct` shape. Mux>direct hypothesis test finally measures what the operator asked.

## Test plan
- [x] `go test ./pkg/router/ -run TestMuxOptsCarriesParentMinHops -count=1 -v` — 4 subtests pass
- [x] `go test ./pkg/router/ -count=1 -short` — full suite passes
- [x] `gofmt -l pkg/router/` — clean
- [ ] CI lanes
- [ ] Post-merge: re-fire Beta's `cli skynet start --routes 4 --min-hops 2` sweep + verify visor log shows N different intermediate transports (no `735f7d73` direct)

### DIFF
--- a/pkg/router/router_dial.go
+++ b/pkg/router/router_dial.go
@@ -1058,6 +1058,17 @@ func (r *router) establishMuxRoutes(
 			ExcludeDMSG:            true,
 		}
 
+		// NOTE: aux mux routes deliberately DO NOT use #2774's
+		// retry-with-exclude-on-failure loop. The mux design is
+		// graceful degradation — if aux route i can't establish (no
+		// disjoint candidate found, or its dial fails), we stop
+		// trying to add more aux routes but the PRIMARY route is
+		// already up and the app is serviceable. A reject-the-whole-
+		// dial retry loop would punish the user for transient aux
+		// flakiness when the primary is healthy. For mux-bw style
+		// callers that DO want strict N-route guarantees, the failure
+		// surfaces via MuxRouteEstablished events (#2756) and the
+		// caller can decide whether to fail or accept partial mux.
 		muxFwd, muxRev, err := r.fetchBestRoutes(ctx, log, lPK, rPK, muxOpts)
 		if err != nil {
 			log.Debugf("Mux route %d/%d: no additional route found: %v", i+1, muxCount, err)
--- a/pkg/router/router_dial.go
+++ b/pkg/router/router_dial.go
@@ -1029,6 +1029,21 @@ func (r *router) establishMuxRoutes(
 	// aux routes diverge from ALL prior ones.
 	excludePKs := intermediatesOfRouteGroup(nrg, lPK, rPK)
 
+	// Thread MinHops + AppName from the parent dial through to each
+	// aux route. Without MinHops, fetchBestRoutes for the aux dial
+	// drops back to r.conf.MinHops (usually 1) and the route-finder
+	// happily returns the direct transport — so a `--routes N
+	// --min-hops 2` call would establish route 0 multi-hop but routes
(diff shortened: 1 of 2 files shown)


0pcom:
LGTM (gh blocks self-approve on this shared fork — comment).

Empirical match with my pre-#2776 visor log:
```
Mux route 1/2 sent handshake via transport 04d81530-...  (multi-hop, intermediate)
Mux route 2/2 established via transport 735f7d73-...    (← direct stcpr)
```
That's the exact shape this PR fixes. Threading both MinHops + AppName is the right scope — AppName for the scoped-logger keying is a nice catch alongside the bug.

Non-blocking note: `establishMuxRoutes` doesn't have #2774's retry-with-exclude loop for its aux dials — if `setupPingRoute` for an aux route fails on a dead intermediate, the aux just drops out of the mux. Acceptable as-is (the loop's existing `Mux route i/N: no additional route found` debug log already covers it), but worth a comment near line 1052 noting that's by design: aux failures degrade mux gracefully rather than rejecting the whole dial. Otherwise a future reader might wonder why the primary route has a retry surface and aux routes don't.

Helper-duplication trade-off in the test file is the right call IMO — the comment makes the maintenance contract explicit. Tests are tight.

Once green + merged + auto-deploy, I'll rerun the N=2/4/8 sweep and confirm `Mux route N/N established via transport ...` shows N distinct intermediate transport IDs (none of them `735f7d73`).

PR #2782, opened 2026-05-23
feat(router): rank route candidates by measured per-hop latency

0pcom:
## Summary

The route-finder service + local-BFS fallback return up to N candidate paths per direction, ordered by hop count + topological discovery only. Among equal-hop candidates the visor was picking the first — which on the 2026-05-23 bilateral \`ss -tinp\` capture meant a geo-distant Indonesia intermediate (cwnd:2 ssthresh:2, 10-18% retransmits, 317ms minRTT) was being selected over a same-hop-count alternative that would have been an order of magnitude faster.

TPD already carries the latency signal end-to-end (visor RSN ping → CXO telemetry feed → `cxoaggregator.UpdateLatency` → Redis → `GetAllTransports` / `GetTransportByID` includes `Entry.Latency`). The router was the only layer not consuming it at pick time. This PR closes the loop.

## Changes

- **`pickDisjointPath`** (`pkg/router/router_dial.go`) gains a `latencyFor func(uuid.UUID) float64` parameter. Among acceptable (non-excluded) candidates, returns the pair with the lowest sum of per-hop avg latencies. nil `latencyFor` preserves the legacy first-acceptable behavior.
- **`pathLatencyScore`** — pure helper, sums per-hop avg ms with an unknown-latency penalty (1000ms) so paths with measured signal outrank fully-unknown ones, while a fully-unknown path is still preferred over no path.
- **`router.buildLatencyLookup`** constructs a `TpID → ms` cache:
  1. `r.tm.Transport(tpID).GetLatencyStats()` for transports the local visor holds (freshest signal, no round-trip)
  2. `r.tm.Conf.DiscoveryClient.GetTransportByID(ctx, tpID)` for intermediate-to-intermediate hops the visor doesn't own — `Entry.Latency` hydrated by the CXO aggregator
- **`fetchBestRoutes`** wires the lookup into the `pickDisjointPath` call site. Candidate set is small (typically 2-3 paths × ≤6 hops); lookup overhead is constant per dial.
- **`calculateLocalRoutes` BFS** no longer returns the first-discovered dst-hit at a given level — it collects all same-level dst-hits and picks the lowest-latency one. Shorter paths still beat longer ones (BFS level-by-level preserved); change is purely a tiebreaker among same-hop-count candidates.

## Tests

4 new subtests in `router_dial_disjoint_test.go`:
- `TestPathLatencyScore_KnownAndUnknown` — pure helper, both branches
- `TestPickDisjointPath_LatencyRanking` — two candidates, ranker picks fast over slow
- `TestPickDisjointPath_LatencyRankingWithExclude` — excluded + ranker compose
- `TestPickDisjointPath_NoLatencyRankerLegacyBehavior` — nil ranker = first-acceptable preserved

Full `pkg/router/` test suite passes (37s).

## Empirical motivation

From the bilateral ss capture with Gamma (16:15Z 2026-05-23):

```
fwd intermediate (114.73.232.8, Indonesia):
  minRTT 317-388ms, cwnd:2-3, ssthresh:2-7, retx 10-18%
rev intermediate (104.136.22.124):
  minRTT 17-43ms, cwnd:10+, retx 0%
```

Both intermediates were returned by the route-finder as same-hop-count candidates. The first one was picked; throughput collapsed to ~10 KB/s. A ranker over the latency data already in TPD would have picked the second.

## Composition with prior PRs

This is the immediate next slice after the 7-PR mux-route plumbing chain (#2751 → #2765 → #2766 → #2770 → #2774 → #2776). The plumbing finally lets visors dial multi-hop reliably; this PR makes the multi-hop pick less adversarial.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./pkg/router/...` clean
- [x] `go test -count=1 ./pkg/router/...` passes (37s)
- [ ] Post-deploy: rerun `cli skynet start --pk gamma -r 8090 -l 9001 --routes 1 --min-hops 2` against Gamma's fileserver. Expect visor log to show the chosen intermediate is the lower-latency candidate per `Entry.Latency`. Throughput should improve unless the latency-signal hydration is stale across the chosen tps (separate issue).

### DIFF
--- a/pkg/router/router_dial.go
+++ b/pkg/router/router_dial.go
@@ -652,7 +652,10 @@ func pickDisjointPath(forward, reverse [][]routing.Hop, exclude []cipher.PubKey,
 	if bestIdx < 0 {
 		return nil, nil, false
 	}
-	return forward[bestIdx], reverse[bestIdx], true
+	// bestIdx is set only by the loop above where 0 <= i < n =
+	// min(len(forward), len(reverse)), so both indexes are in-bounds.
+	// gosec can't track the cross-slice min, hence the explicit nolint.
+	return forward[bestIdx], reverse[bestIdx], true //nolint:gosec
 }
 
 // pathLatencyScore returns the sum of per-hop avg-latency-ms across a
--- a/pkg/router/router_dial.go
+++ b/pkg/router/router_dial.go
@@ -555,15 +555,22 @@ fetchRoutesAgain:
 
 	log.Debugf("Found routes Forward: %s. Reverse %s", paths[forward], paths[backward])
 
-	// DisjointMux post-filter: if the caller populated
-	// ExcludeIntermediatePKs (typically the mux loop tracking
-	// intermediates of already-established routes), iterate the
-	// route-finder's response and return the first path whose
-	// intermediates don't overlap with the exclude set. The
-	// route-finder doesn't natively know about this constraint, so
-	// we filter client-side. Source + destination are NOT considered
-	// "intermediates" (they're endpoints).
-	fwdPath, revPath, ok := pickDisjointPath(paths[forward], paths[backward], opts.ExcludeIntermediatePKs)
+	// DisjointMux post-filter + latency ranking: iterate the route-
+	// finder's response and return the path whose intermediates don't
+	// overlap with opts.ExcludeIntermediatePKs AND whose total per-hop
+	// avg latency is lowest among acceptable candidates. The route-
+	// finder service doesn't sort by latency at query time today; the
+	// visor has the same TPD latency data available in-memory (own
+	// transports via tm.GetLatencyStats) + via TPD entries (already
+	// aggregated by pkg/transport-discovery/cxoaggregator from the
+	// CXO telemetry feed). Building the lookup here puts the ranking
+	// in the dial hot path with no extra round-trip cost — local-
+	// transport latency is a constant-time map read; non-local hops
+	// fall back to GetTransport's cached entry. Pass-through when
+	// the route-finder returned a single candidate (lookup is built
+	// but never iterated).
+	latencyFor := r.buildLatencyLookup(ctx, paths[forward], paths[backward])
+	fwdPath, revPath, ok := pickDisjointPath(paths[forward], paths[backward], opts.ExcludeIntermediatePKs, latencyFor)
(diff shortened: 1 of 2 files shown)


0pcom:
LGTM (Gamma — co-investigator of the bilateral ss capture).

The design closes the loop cleanly: pickDisjointPath gets a latency-aware tiebreak across acceptable candidates while staying back-compat (nil latencyFor + empty exclude → paths[0]). The closure-based lookup pays-per-use. The 1000ms unknown-penalty is well-chosen — high enough that known-good paths win, low enough that a fully-unknown path beats no path.

Specifics I verified:
- `pickDisjointPath`'s new signature preserves the hot-path single-candidate behavior via the explicit `len(exclude) == 0 && latencyFor == nil` short-circuit. Tests passing this combination won't behaviorally change.
- `pathLatencyScore` correctly sums per-HOP not per-intermediate-PK. Asymmetric intermediate-quality (the bilateral ss showed 114.73.232.8 having different RTT per-TpID per-peer) is preserved.
- `buildLatencyLookup` falls back to TPD cleanly when local TM doesn't hold the TpID. The CXO aggregator pipeline gives this real data.

Non-blocking observation for follow-up:

`buildLatencyLookup` calls `r.tm.Conf.DiscoveryClient.GetTransportByID(ctx, id)` SERIALLY per unknown TpID. For a typical 3-candidate × ≤6-hop fetch with mostly-unknown remote TpIDs (the common case on a fresh-process visor before CXO has fully populated), that's ~12 sequential HTTP calls per `fetchBestRoutes` call. Under the retry-loop (#2774) which rebuilds the lookup on each iteration with growing ExcludeIntermediatePKs, this can compound to several hundred ms before the ranker even starts. Worth considering for a follow-up PR:
- Visor-wide TTL cache (60-300s) keyed on TpID, populated lazily by buildLatencyLookup
- OR batch GetTransportByID — does the TPD API support multi-ID query?
- OR shorten per-call timeout aggressively (e.g., 50ms) and treat slow lookups as unknown — the unknownLatencyCostMs penalty already handles missing data gracefully

Not blocking — the correctness here is solid and the perf optimization is a clean follow-up. Empirical motivator citation in the PR body is great — the bilateral ss data is exactly the artifact that justifies this surface.

PR #2785, opened 2026-05-23
feat(router): unpair forward/reverse candidate selection

0pcom:
## Summary

The route-finder returns forward and reverse candidates **independently** (`pkg/router/router_dial.go:507` queries `[]routing.PathEdges{forward, backward}`); the data plane stores them as independent rule chains (`route_group.fwd` + `route_group.rvs`); there's no architectural requirement that the two share the same intermediates. The dial layer was the only place enforcing symmetry, and only implicitly — `pickDisjointPath` paired candidates by index.

Pre-#2782 + the index-pairing in #2782's `pickDisjointPath` threw that independence away — when the latency ranker picked `(forward[i], reverse[i])` by paired index, it ignored the case where forward's best candidate sat at one index and reverse's best sat at another.

Now: each direction is ranked separately via `pickBestDirection`. Best forward + best reverse (each by sum-of-per-hop-latency) regardless of which slice index they came from. Disjointness check applies per-direction.

## Why this matters

Real workloads are **bandwidth-asymmetric**:
- HTTP GET: tiny upstream, bulk downstream
- File transfer: same shape
- Video streaming: same
- skynet port-forwarding: same (the operator's actual product surface)

Letting each direction find its best wire — possibly a different intermediate — turns "best paired" into "best per direction" with no extra round-trip cost and no protocol change.

## Tests

- `TestPickDisjointPath_AsymmetricForwardReverse` — fwd[1] is fast, rev[0] is fast, paired-by-index would tie-break on first; unpaired returns (fwd[1], rev[0]) which is strictly better.
- `TestPickDisjointPath_AsymmetricExcludeIntersection` — exclude check applies per-direction; if every reverse candidate touches the exclude set, `ok=false`.

Existing `Pick*` and `PathLatencyScore*` tests pass unchanged. Full `pkg/router/` suite (37s) green.

## Composition

Companion to #2782 (latency-rank). Together with the chain #2751→#2765→#2766→#2770→#2774→#2776→#2782, multi-hop went from 0-15 KB/s pre-deploy → 1.2-2.5 MB/s in empirical testing 2026-05-23 (3 of 5 N=1 --min-hops 2 reps completed full 50MB transfer on first attempt). Unpairing is the next refinement.

## Test plan

- [x] `go build ./...` clean
- [x] `go test -count=1 ./pkg/router/...` passes (37s)
- [ ] Post-deploy: rerun N=1 --min-hops 2 sweep. Visor log should show that selected forward + reverse transport IDs no longer come from the same route-finder candidate index when independent ranking finds a better split.

### DIFF
--- a/pkg/router/router_dial.go
+++ b/pkg/router/router_dial.go
@@ -582,32 +582,40 @@ fetchRoutesAgain:
 	return fwdPath, revPath, nil
 }
 
-// pickDisjointPath walks the parallel forward/reverse-path slices the
-// route-finder returned (paired by index) and returns the
-// forward/reverse pair whose intermediate hops do not contain any PK
-// in exclude AND whose total measured latency is lowest.
+// pickDisjointPath returns the forward/reverse path pair whose
+// intermediate hops do not contain any PK in exclude AND whose
+// per-direction latency is lowest. Forward and reverse are RANKED
+// INDEPENDENTLY — the route-finder returns each direction as its own
+// candidate slice (PathEdges{forward, backward} at the FindRoutes call
+// site) and the data plane stores fwd / rvs as separate rule chains
+// (route_group.fwd / route_group.rvs), so there's no requirement that
+// the two paths share the same intermediates.
 //
-// latencyFor is a per-TpID latency lookup (avg-ms; 0 = unknown). Pairs
-// containing any unknown-latency hop are still considered, but unknown
-// hops are treated as a high cost so KNOWN-good paths are preferred
-// when available. Pass nil to disable latency ranking (legacy
-// first-acceptable behavior).
+// Pre-#2782 this function paired by index (forward[i], reverse[i]),
+// which threw away the independent-direction signal: if forward had
+// a great candidate at index 2 but the best reverse was at index 0,
+// pairing-by-index would force forward[0] + reverse[0] (or whatever
+// matching pair scored lowest combined). Unpairing closes that gap.
+//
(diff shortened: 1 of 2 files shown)


0pcom:
LGTM. Clean refactor of the per-direction independence; the pickBestDirection extraction reads naturally.

The empirical case for unpairing is real — the bilateral ss capture (cited in #2782) literally showed asymmetric per-direction quality (forward 0385fc26 via Indonesia 317ms minRTT vs reverse 03fb3f50 local 17ms minRTT). Pair-by-index would have either forced the bad-fwd + good-rev combo OR vice versa; unpaired, each direction independently picks the best wire.

Specific things I verified:
- `pickBestDirection` correctly preserves the legacy "no ranker → first acceptable" path when `latencyFor == nil`.
- `ok=false` propagation is right: if EITHER direction has no acceptable candidate, the function returns false. No silent fallback to a half-defined route.
- The cross-direction nolint:gosec from #2782 isn't needed anymore because the indexing is now per-direction. Nice.
- The pre-#2782 pairing-by-index behavior wasn't documented; the new docstring captures what was lost vs gained explicitly.

Predicted empirical signal post-deploy: for asymmetric workloads (curl payload, bulk-down + tiny-up), the bulk-down direction will preferentially pick the best-known reverse intermediate independent of which forward picks. With ~3 candidates per direction × 3 from each side = 9 effective combinations vs the previous 3 paired — strictly more headroom for the ranker.

Once this auto-deploys, Beta's pull-from-Gamma test should improve further if forward and reverse independent-optimums sit at different indices. Will spot-check on my side post-merge.

PR #2792, opened 2026-05-23
feat(router): per-direction MinHops for asymmetric routing

0pcom:
## Summary

Real workloads are bandwidth-asymmetric — HTTP GET = tiny upstream + bulk downstream; file transfer, video, skynet port-forwarding all follow this shape. Symmetric \`MinHops\` forces forward and reverse paths through the same hop-count constraint, so the user can't say "direct upstream + multi-hop downstream" — the operator's canonical asymmetric test.

This PR adds \`ForwardMinHops\` + \`ReverseMinHops\` to \`router.DialOptions\`. When either is > 0 it overrides the symmetric \`MinHops\` for that direction only. Zero = inherit \`MinHops\`.

**Canonical use** (post-deploy):
```
cli skynet start --pk <peer> --routes 4 --reverse-min-hops 2
```
Forward stays direct (low-latency upstream for the GET), reverse is forced through 2-hop intermediates and aggregated across 4 mux routes (downstream bandwidth aggregation).

## Implementation

- \`pkg/router/router.go\`: new fields on \`DialOptions\` + \`EffectiveMinHops(forward bool)\` + \`AnyMinHopsConstraint()\` helpers.
- \`pkg/router/router_dial.go\`:
  - Direct-tp downgrade in \`DialRoutes\` now suppressed by **any** min-hops constraint (not just symmetric).
  - \`fetchBestRoutes\` does 2 route-finder queries when forward and reverse \`MinHops\` differ (single call otherwise — common path).
  - \`calculateLocalRoutes\` (local BFS fallback) uses \`max(symmetric, fwd, rev)\` since it mirrors forward to reverse via \`reverseHops()\` — must satisfy the stricter direction.
  - \`establishMuxRoutes\` threads per-direction \`MinHops\` to aux routes too.
- \`pkg/app/appserver/rpc_ingress_gateway.go\`: \`DialOptionsReq\` gains \`ForwardMinHops\` + \`ReverseMinHops\`; \`dialInternal\` + \`dialWithMuxRoutes\` plumb them through to \`router.DialOptions\`.
- \`pkg/app/appserver/rpc_ingress_client.go\`: \`DialWithOptions\` signature gains \`fwdMinHops, revMinHops int\` params (mirrored in mock).
- \`pkg/app/client.go\`: same signature extension; non-zero in any of the new fields triggers the \`rpcC.DialWithOptions\` path.
- \`cmd/apps/skynet-client/commands/skynet-client.go\`: new \`--forward-min-hops\` + \`--reverse-min-hops\` flags; threaded to \`appCl.DialWithOptions\`. Dial-shape log collapsed into a single line listing only the set fields.
- \`cmd/skywire-cli/commands/skynet/root.go\`: same flags on the user-facing wrapper, forwarded into the spawned app's args map.

## Tests

- \`pkg/router/dial_options_test.go\`: 2 helper tables covering all field combinations including the asymmetric reverse-only case.
- \`pkg/app/appserver/rpc_ingress_gateway_test.go\`: new subtest pins that \`ForwardMinHops\` / \`ReverseMinHops\` fall through to default dial on non-skywire networkers (mirrors existing MuxRoutes / MinHops fallthrough cases — pins the wire shape of the new fields).
- Existing tests updated for the 5-arg \`DialWithOptions\` signature.
- Full \`pkg/router/\` + \`pkg/app/\` test suites pass.

## What this PR does NOT do (deferred)

- **Per-direction MuxRoutes** (1 forward leg + N reverse legs). Requires data-plane changes since \`route_group\`'s \`fwd[]\` and \`rvs[]\` are populated by paired bidirectional setup. Followup.
- **\`calculateLocalRoutes\` asymmetric BFS.** Currently uses \`max()\` over per-direction MinHops since it mirrors forward to reverse via \`reverseHops()\`. Genuine two-BFS would be larger refactoring.

## Composition

Continues the 8-PR plumbing chain #2751→#2765→#2766→#2770→#2774→#2776→#2782→#2785 that unlocked usable multi-hop on the production deployment today: 1.2-2.5 MB/s on 50 MB pulls, 4/5 reps complete vs 0/N pre-chain. Per-direction \`MinHops\` makes the next class of test (asymmetric workload shape) measurable.

## Test plan

- [x] \`go build ./...\` clean
- [x] \`go test -count=1 -timeout 60s ./pkg/router/... ./pkg/app/...\` all pass (38s)
- [ ] Post-deploy: fire \`cli skynet start --pk <peer> -r 8090 -l 9300 --routes 4 --reverse-min-hops 2\` against Gamma's fileserver. Visor log should show:
  - \`Per-accept dial shape: ... with 4 parallel mux routes reverse-min-hops=2\`
  - Two \`Found routes\` log lines per dial cycle (forward + reverse route-finder queries with different \`MinHops\`)
  - Forward path's hop-count short (1 if direct available), reverse path's hop-count \>= 2

### DIFF
--- a/cmd/apps/skynet-client/commands/skynet-client.go
+++ b/cmd/apps/skynet-client/commands/skynet-client.go
@@ -49,6 +49,13 @@ var (
 	// every mux route, so N mux streams ride one TCP socket and
 	// nothing is multiplexed across intermediates.
 	minHops int
+	// fwdMinHops / revMinHops are per-direction MinHops overrides for
+	// bandwidth-asymmetric workloads (HTTP GET = tiny upstream + bulk
+	// downstream). When > 0 they win over min-hops for THAT direction.
+	// Canonical asymmetric test: --forward-min-hops 0 --reverse-min-hops 2
+	// lets forward stay direct while reverse is forced multi-hop.
+	fwdMinHops int
+	revMinHops int
 )
 
 func init() {
@@ -59,6 +66,8 @@ func init() {
 	RootCmd.Flags().Uint16Var(&appPort, "port", 0, "routing port for communication between app and visor")
 	RootCmd.Flags().IntVar(&routes, "routes", 0, "number of parallel skynet mux routes (0 or 1 = single route)")
 	RootCmd.Flags().IntVar(&minHops, "min-hops", 0, "force routes through at least this many intermediates (>=2 rejects direct paths)")
+	RootCmd.Flags().IntVar(&fwdMinHops, "forward-min-hops", 0, "per-direction forward MinHops override (>=2 forces multi-hop on forward direction only)")
+	RootCmd.Flags().IntVar(&revMinHops, "reverse-min-hops", 0, "per-direction reverse MinHops override (>=2 forces multi-hop on reverse direction only)")
 }
 
 // RootCmd is the root command for skynet-client
@@ -90,6 +99,8 @@ func RunSkynetClient(ctx context.Context, args []string) error {
 		fs.Uint16Var(&appPort, "port", 0, "routing port")
 		fs.IntVar(&routes, "routes", 0, "number of parallel skynet mux routes")
 		fs.IntVar(&minHops, "min-hops", 0, "minimum hop count (>=2 rejects direct paths)")
+		fs.IntVar(&fwdMinHops, "forward-min-hops", 0, "per-direction forward MinHops override")
+		fs.IntVar(&revMinHops, "reverse-min-hops", 0, "per-direction reverse MinHops override")
 		if err := fs.Parse(args); err != nil {
 			return fmt.Errorf("failed to parse flags: %w", err)
 		}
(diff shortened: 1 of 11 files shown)


0pcom:
LGTM. Clean per-direction plumbing across 11 files; the API surface adds two well-scoped fields + two helper methods on DialOptions that callers can ignore by default.

Specifics I verified:

- **DialOptions** additions are back-compat: `ForwardMinHops`/`ReverseMinHops` default to 0 which means "inherit symmetric MinHops", and the resolver methods (`EffectiveMinHops`, `AnyMinHopsConstraint`) handle nil opts safely.
- **DialRoutes direct-tp downgrade**: `AnyMinHopsConstraint` replaces the bare `opts.MinHops <= 1` check. The empty-options case still gets the downgrade. The reverse-only-multi-hop case (ReverseMinHops=2, MinHops=0) correctly suppresses the downgrade. Matches the canonical use case in the PR description.
- **fetchBestRoutes split**: only fires the second route-finder query when `fwdMinHops != revMinHops`. Symmetric callers stay on the single-query hot path. Cost is one extra HTTP round-trip per dial when asymmetric — fine for an opt-in feature.
- **calculateLocalRoutes**: the honest constraint that local BFS mirrors forward→reverse, so we MUST satisfy MAX(fwd, rev) MinHops there. Documented explicitly. Operators wanting truly asymmetric local paths need the route-finder. Right design.

Tests cover the resolver matrix (forward/reverse independence, per-direction wins, asymmetric reverse-only) and `AnyMinHopsConstraint`'s canonical use case. Good coverage of the surface.

Composes nicely with #2782 + #2785: ranker picks best candidate per-direction, asymmetric routing lets the operator skip multi-hop entirely on the cheap-direction.

Once this deploys: the canonical curl-payload case (`cli skynet start --pk peer -r 8090 -l 8091 --routes 4 --reverse-min-hops 2`) gets direct upstream (small HTTP GET) + ranked-multi-hop downstream (50MB body). Expecting close-to-direct throughput since the bulk path can pick the best reverse candidate without forward-direction constraint dragging it.

Will spot-check after auto-deploy if Beta runs an empirical sweep with the new flag.

PR #2797, opened 2026-05-23
feat(router): per-direction MuxRoutes for asymmetric route counts

0pcom:
## Summary

Slice C of the asymmetric-routing series — completes the operator's canonical bandwidth-asymmetric test shape: **1 forward leg (cheap upstream) + N reverse legs (aggregated downstream payload)**.

The data plane was already capable: \`rg.fwd[]\`+\`rg.tps[]\` and \`rg.rvs[]\` are independent slices (the read path is by route-id lookup, not slice-indexed iteration). \`RemoveMuxRouteByTransport\` even has \`if idx < len(rg.rvs)\` guards that handle different lengths. The symmetric-only constraint was purely a side effect of how \`establishMuxRoutes\`' loop appended fwd+rev paired.

## Canonical use

\`\`\`
cli skynet start --pk <peer> -r 8090 -l 9300 \\
  --forward-mux 1 --reverse-mux 4 --reverse-min-hops 2
\`\`\`

→ 1 forward leg (direct upstream for the GET request) + 4 reverse legs forced multi-hop, aggregating the bulk payload across 4 disjoint intermediates.

## Changes

- \`pkg/router/router.go\`: \`ForwardMuxRoutes\` / \`ReverseMuxRoutes\` on \`DialOptions\`, with \`EffectiveMuxRoutes(forward bool)\` helper matching the per-direction-MinHops pattern from #2792.
- \`pkg/router/route_group.go\`: \`appendForwardLeg\` (fwd+tps only) + \`appendReverseLeg\` (rvs only) — companions to the existing \`appendRules\`. Mux weight rebuild only triggers when \`tps[]\` grows.
- \`pkg/router/router_mux_ops.go\`: \`appendRouteAsymmetric(nrg, rules, addFwd, addRev)\`. When one direction isn't wanted, the rule on that side is deleted from the local routing table to avoid stale-entry leaks (the setup-node always creates both directions of every \`BidirectionalRoute\`; this is the cleanup).
- \`pkg/router/router_dial.go\`: \`establishMuxRoutes\` computes \`fwdCount\` + \`revCount\` via \`EffectiveMuxRoutes\`, loops \`max(fwdCount,revCount)-1\`, and per-iteration sets \`addFwd = (i < fwdCount)\`, \`addRev = (i < revCount)\`. The exclude-set accumulation tracks per-side intermediates independently so disjoint-intermediate routing works correctly in asymmetric topologies too.
- \`pkg/app/appserver/rpc_ingress_gateway.go\`: \`DialOptionsReq\` gains \`ForwardMuxRoutes\` + \`ReverseMuxRoutes\`; \`dialWithMuxRoutes\` plumbs them into \`router.DialOptions\`.
- \`pkg/app/appserver/rpc_ingress_client.go\` (+ mock): \`DialWithOptions\` signature gains \`fwdMux, revMux int\` — now 7 ints total. (A struct-based variant would be cleaner long-term; left as future cleanup.)
- \`pkg/app/client.go\`: same signature extension; non-zero in any of the new fields triggers the \`rpcC.DialWithOptions\` path.
- \`cmd/apps/skynet-client\` + \`cmd/skywire-cli/commands/skynet/root.go\`: \`--forward-mux\` + \`--reverse-mux\` flags. Dial-shape log surfaces the new fields when set.

## Tests

- \`pkg/router/dial_options_test.go\`: 10 subtests for \`EffectiveMuxRoutes\` covering all field combos + canonical asymmetric (fwd=1 rev=N).
- Existing tests updated for 7-arg \`DialWithOptions\` signature.
- Full \`pkg/router/\` + \`pkg/app/\` suites pass (38s).

## Composition

Continues the chain #2751→#2765→#2766→#2770→#2774→#2776→#2782→#2785→#2792. Slice B (per-direction MinHops) + Slice C (per-direction MuxRoutes) together complete the asymmetric-routing surface end-to-end.

## Test plan

- [x] \`go build ./...\` clean
- [x] \`gofmt -l\` clean
- [x] \`go vet ./pkg/router/... ./pkg/app/... ./cmd/...\` clean
- [x] \`go test -count=1 -timeout 90s ./pkg/router/... ./pkg/app/...\` all pass (38s)
- [ ] Post-deploy: \`cli skynet start --pk <peer> -r 8090 -l 9300 --forward-mux 1 --reverse-mux 4 --reverse-min-hops 2\`. Visor log should show 1 forward "Mux route established" + 3 reverse-only aux "Mux route established (fwd=false rev=true)" entries. The Write path picks among 1 forward transport; the Read path accepts on 4 reverse route-ids.

### DIFF
--- a/cmd/apps/skynet-client/commands/skynet-client.go
+++ b/cmd/apps/skynet-client/commands/skynet-client.go
@@ -56,6 +56,13 @@ var (
 	// lets forward stay direct while reverse is forced multi-hop.
 	fwdMinHops int
 	revMinHops int
+	// fwdMux / revMux are per-direction MuxRoutes overrides. When > 0
+	// they win over --routes for THAT direction's route count.
+	// Canonical download-heavy shape: --forward-mux 1 --reverse-mux 4
+	// yields 1 forward leg (cheap upstream) + 4 reverse legs that
+	// aggregate the bulk payload.
+	fwdMux int
+	revMux int
 )
 
 func init() {
@@ -68,6 +75,8 @@ func init() {
 	RootCmd.Flags().IntVar(&minHops, "min-hops", 0, "force routes through at least this many intermediates (>=2 rejects direct paths)")
 	RootCmd.Flags().IntVar(&fwdMinHops, "forward-min-hops", 0, "per-direction forward MinHops override (>=2 forces multi-hop on forward direction only)")
 	RootCmd.Flags().IntVar(&revMinHops, "reverse-min-hops", 0, "per-direction reverse MinHops override (>=2 forces multi-hop on reverse direction only)")
+	RootCmd.Flags().IntVar(&fwdMux, "forward-mux", 0, "per-direction forward MuxRoutes override (>0 sets forward leg count independent of --routes)")
+	RootCmd.Flags().IntVar(&revMux, "reverse-mux", 0, "per-direction reverse MuxRoutes override (>0 sets reverse leg count; canonical download-heavy shape: --forward-mux 1 --reverse-mux N)")
 }
 
 // RootCmd is the root command for skynet-client
@@ -101,6 +110,8 @@ func RunSkynetClient(ctx context.Context, args []string) error {
 		fs.IntVar(&minHops, "min-hops", 0, "minimum hop count (>=2 rejects direct paths)")
 		fs.IntVar(&fwdMinHops, "forward-min-hops", 0, "per-direction forward MinHops override")
 		fs.IntVar(&revMinHops, "reverse-min-hops", 0, "per-direction reverse MinHops override")
+		fs.IntVar(&fwdMux, "forward-mux", 0, "per-direction forward MuxRoutes override")
+		fs.IntVar(&revMux, "reverse-mux", 0, "per-direction reverse MuxRoutes override")
 		if err := fs.Parse(args); err != nil {
 			return fmt.Errorf("failed to parse flags: %w", err)
 		}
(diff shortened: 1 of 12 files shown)


0pcom:
LGTM. The asymmetric mux topology is the natural closure of #2782 + #2785 + #2792 — full per-direction control end-to-end. Caller can express `direct upstream + N multi-hop downstream` (the canonical HTTP-GET shape) and the visor honors it both in route selection AND in route setup.

Specifics I verified:

- **EffectiveMuxRoutes** mirrors the EffectiveMinHops pattern cleanly. Back-compat: nil opts → 0 ("default to symmetric"); per-direction > 0 wins over symmetric; per-direction == 0 falls through. Same shape callers learned from #2792.

- **appendRouteAsymmetric** correctly deletes the unused-direction rule from the routing table. The setup-node-creates-both / cleanup-the-side-we-don't-use semantics is the right design — without it, asymmetric mux would silently accumulate stale rules on every dial. The DelRules call is paired with each rejection branch.

- **establishMuxRoutes loop bound**: `max(fwdCount, revCount) - 1` is right. The `addFwd && !addRev` and `!addFwd && addRev` branches correctly invoke appendRouteAsymmetric with the appropriate flags. The `!addFwd && !addRev` skip is defensive — current call shape can't hit it but the guard is cheap.

- **Handshake on new forward transport**: preserved at the per-leg level. Each new forward leg gets its own `CapMux|CapSACK` handshake on the appended transport. Reverse legs don't need a per-leg handshake since they're read-only-by-route-id-lookup on this visor's side.

- **DMSG safety check** inside appendRouteAsymmetric mirrors the appendRouteToGroup check — refuses to mux a DMSG transport in. Good defensive copy of the rule.

The data-plane unlock note in the PR description is the architecturally interesting bit: rg.fwd[] and rg.rvs[] being independent slices means asymmetric mux is purely a setup-side change. Worth calling out for anyone digging into the code later.

Tests cover EffectiveMuxRoutes with 10 subtests + appendRouteAsymmetric flag combinations through the integration path. Good coverage.

Once this lands + auto-deploys: `cli skynet start --pk peer -r 8090 -l 8091 --forward-mux 1 --reverse-mux 4 --reverse-min-hops 2` should be the empirical sweet spot for the curl-payload workload. 1 direct forward leg (~1 RTT for HTTP GET) + 4 reverse multi-hop legs aggregating the 50MB body. Expecting throughput close to direct-baseline since the bulk path can pick best-of-4 reverse intermediates.

Will spot-check after Beta's empirical sweep.

PR #2854, opened 2026-05-27
fix(hypervisor): warm summaryCache on accept + 503 instead of 404 for cached peers

0pcom:
### Eager Summary on accept

After hypervisor restart, every peer redials within seconds. The accept loop adds them to remoteVisors fast. But summaryCache only gets populated by getAllVisorsSummary polls bounded by the 5s outer cap (#2842), so peers whose Summary RPC takes >5s the first time around drop OUT of the response until later polls land. With ~10 peers and slow first Summaries, the visible list fills over 30+ seconds — operators describe it as "instant before, slow now."

Fix: spawn a one-shot `warmSummaryCache` goroutine on every accept (both dmsg and skynet paths). By the time the UI's first poll arrives, the cache is warm — all rows render immediately. One Summary call per accept, ~25 bytes on the wire.

### 503 instead of 404 for cached peers

`visorCtx` returns 404 "visor of pk '...' not found" the moment the peer drops out of remoteVisors (Summary fails → eviction → mid-flap between stream-idle and redial). Operators visiting a peer detail page during that flap see "Error: Visor of pk '...' not found." even though the peer is back online ~2s later.

Fix: when the PK is in summaryCache (we knew it recently) but not in remoteVisors, return 503 with `visor '...' is currently disconnected (last seen Xs ago) — retrying`. The UI's retry loop catches the peer when it re-registers. 404 stays for the genuinely-unknown PK case.

### DIFF
--- a/pkg/visor/hypervisor.go
+++ b/pkg/visor/hypervisor.go
@@ -318,6 +318,10 @@ func (hv *Hypervisor) ServeRPC(ctx context.Context, dmsgPort uint16) error {
 				hv.visor.remoteVisors[peerPK] = *visorConn
 				hv.remoteVisors[peerPK] = *visorConn
 				hv.mu.Unlock()
+
+				// Warm-cache: see comment in the dmsg accept path
+				// below — same intent, same one-shot Summary.
+				go hv.warmSummaryCache(peerPK, visorConn.API)
 				if hv.lanDmsg != nil {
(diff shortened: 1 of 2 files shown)


0pcom:
Reverting course — the underlying issue is my #2842 made eviction more aggressive (eviction moved into the inner goroutine). Going to revert that aspect instead. Holding this.

Issue #2775, opened 2026-05-23
Unified app framework refactor (post-release): collapse skynet srv/client, resolver, hypervisor, dmsgpty into one service contract


0pcom:
## Status

**Deferred until after the next release.** This is a significant refactor; capturing the design discussion + decisions here so the work can resume cleanly post-release. Filed as an outcome of an in-session design exchange between operator + Alpha (claude agent) on 2026-05-23.

## Problem

Today the visor has multiple parallel namespaces for things that are conceptually the same shape — a server / client / handler with identity, lifecycle, config, ports, and run-modes:

- Launcher apps (`cli visor app …`): skysocks, vpn-client, skychat, skynet-web, skysocks-client-XXXX
- Skynet (`cli skynet …`): per-CLI-invocation port forwarders + named srvs (skynet-3435, etc.)
- Visor proxies (`cli visor proxies …`): .dmsg / .skynet resolving proxies
- Visor built-ins: hypervisor UI, log server, dmsgctrl, dmsgpty

Each has its own config schema, its own CLI namespace, its own RPC surface, its own start/stop semantics. The fragmentation has cost:

- **No unified visibility** — `cli visor app ls` shows nothing for skynet srvs, resolvers, or built-ins
- **No per-app route view** — RouteGroups have an `app_name` tag set only by launcher apps; everything else dials without identity, leaving the operator with no way to ask "what routes is this thing using" or "who's connected to my skynet srv"
- **Routing controls are flag-stew per dial** — `--routes N --min-hops K --via PK` repeated on every invocation; no persistent per-app policy
- **Standalone-mode unevenness** — some apps support detached operation (skychat tcp-direct, dmsgpty ssh-equivalent), others don't, and the contract isn't uniform

### Modes taxonomy

Each app declares which of three run-modes it supports:

| Mode | Identity | Overlay | Lifecycle owner | When |
|---|---|---|---|---|
| visor-app | shares visor | visor's DMSG + router | visor (in-process) | default |
| dmsg-standalone | own keys, own DMSG client | own DMSG client, no router | user / external supervisor | detached operation |
| TCP-standalone | own keys or visor keys, no DMSG | direct pk-encrypted TCP, no overlay | user (visor never starts this — one exception: hypervisor) | LAN-style, port-forwarding deployments |

**Rules:**

- Anything that can run as visor-app or over skynet can also run dmsg-standalone (skynet rides DMSG, so the capability flows down).
[...]

PR #2863, opened 2026-05-27
docs: RFC — unified service contract (#2775)

0pcom:
## Summary

Captures the design space for unifying the visor's supervised-service surface across launcher apps and visor-internal subsystems.

Predecessor PRs #2860 / #2861 / #2862 (phases 1, 2a, 2b) named the launcher-app contract; this RFC charts the path for the next phase that those PRs do not by themselves accomplish.

**Revised after the first round of discussion** — hypervisor is now **out of scope**. The `skywire web` HTTP-CLI surface is the operator-side meta-management interface; it delivers the uniformity originally wanted from "hypervisor as an app" at the interface layer instead of the supervisor layer. Phase 3 shrinks to dmsgpty + dmsgweb + skynetweb only (~150 LOC, no critical-path lifecycle refactor).

## What's in the RFC (revised)

- **Scope revision** at the top explaining why hypervisor was dropped.
- **What unification still buys** with the narrower scope (uniform `cli visor app ls` for the three subsystems, drop the bespoke `EmbeddedProxies` RPC, dmsgpty gains a runtime on/off switch, supervision becomes available via Phase 4, construction sites unchanged).
- **The asymmetry table** with the four categories (launcher apps + the three migrated subsystems + hypervisor as "out of scope").
- **The chosen shape** — app-shaped facades wrapping existing `Start()`/`Stop()` methods. With code sketch.
- **Why this shape and not Option C** (separate Runnable interface) or Option A (richer Client) — both were motivated by hypervisor's complexity; with hypervisor out, both collapse.
- **Four-phase migration plan**: config translation → dmsgweb/skynetweb → dmsgpty → cosmetic UI distinguishing apps from internal services.
- **Open questions** sharpened: default RestartPolicies per service (recommendations included), naming convention, synthetic-config visibility, idempotent start/stop semantics.

## What this PR does *not* do

- No code changes outside `docs/`.
- No implementation. Draft status — waiting on team input on the decision checklist.

## Test plan

- [x] Markdown renders correctly on GitHub.
- [ ] Team confirms scope revision (hypervisor out, web CLI is the meta-surface).
- [ ] Team resolves the four sharpened open questions.

### DIFF
--- a/docs/unified_service_contract_rfc.md
+++ b/docs/unified_service_contract_rfc.md
@@ -12,15 +12,21 @@ Status: **Draft.** Scope was revised after the first round of discussion (see "S
 
 ## Scope revision
 
-The original `#2775` issue body included the hypervisor UI in the "make it an app" list. That decision was made before the `skywire web` HTTP-CLI surface was scaffolded. With the web CLI in the picture, the unification this RFC is really after — **uniform operator visibility and control** — gets delivered at the operator-interface layer, not at the supervisor layer:
+The original `#2775` issue body included the hypervisor UI in the "make it an app" list. That decision was made before the `skywire web` surface was scaffolded. With the web PTY in the picture, the unification this RFC is really after — **uniform operator visibility and control** — gets delivered at the operator-interface layer, not at the supervisor layer.
 
-- The web CLI exposes the full `skywire` binary's CLI tree over an HTTP listener (localhost, forwardable). Any capability the binary supports — including hypervisor on/off, config edits, transport ops, RPC introspection — is reachable through one generic surface.
-- The thing that originally made "hypervisor as an app" attractive was uniform visibility (`cli visor app ls` showing the full picture). The web CLI delivers that same uniformity by giving operators a single remote-control surface over the entire binary, not just the apps subset.
+**About the web PTY.** Mechanistically, `skywire web` is **dmsgpty serving its pseudoterminal over HTTP/WebSocket** so the operator's browser can render the terminal. The skywire CLI runs inside that PTY as a normal interactive shell session. It is **not** a REST/RPC gateway exposing per-command endpoints — it is the same PTY mechanism that backs `cli dmsg pty exec`, just rendered in a different transport. The security model inherits dmsgpty's PK-based authentication; no new auth surface is introduced.
+
+The implications for this RFC:
+
+- Any capability the `skywire` binary supports — hypervisor on/off, config edits, transport ops, RPC introspection — is reachable through the PTY by the operator typing the corresponding `skywire ...` command. No code change to expose new things; everything the CLI can already do is already there.
+- The thing that originally made "hypervisor as an app" attractive was uniform visibility (`cli visor app ls` showing the full picture). The web PTY delivers that same uniformity *one level up*: the operator's single remote-control surface is the shell, and `cli visor app ls` is just one of many things they run inside it.
 - Hypervisor stays as a visor subsystem with its existing init wiring. No lifecycle refactor inside critical-path code.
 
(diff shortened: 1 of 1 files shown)


0pcom:
Reviewed. Strong agreement with the recommendation. Walking the decision checklist:

**Pick Option B over C (and never A) — agree.** A's god-object client leaks dangerous visor internals into a type whose external sibling silently can't honor them. The runtime-discriminated-kind footgun is exactly the class of bug that bites at 3am during an incident; C's compile-time split makes the asymmetry honest. But shipping B first is right: we don't yet know whether `RestartPolicy` semantics will look the same on `*Hypervisor` (stateful, holds open WS connections) as on `skynet-client` (largely stateless). Land B, run a release, see what cracks, then promote to C with a year of operational data on what a supervisor's contract should actually be.

**Phase order 3.1 → 3.5 — approve.** Config translation before any subsystem registration is the right plumbing-first sequencing. Doing resolvers (3.2) ahead of dmsgpty (3.3) ahead of hypervisor (3.4) walks blast radius cleanly — by the time we touch the hypervisor we've already burned in start/stop semantics on two lower-stakes services. 3.5 (UI cosmetic) is correctly the trailing nice-to-have; if it slips a cycle the world doesn't end.

**Open question 1 (idempotent start/stop) — reuse the same instance; mandate reentrant Enable/Disable in the contract.** Hypervisor's `Enable`/`Disable` naming already implies this is the intended shape. Fresh-construct-on-restart means destroying open dmsg sessions, blowing the route cache, and re-listening on the same port (with the bind-race that implies). For dmsgpty and the resolvers, their existing `Start/Stop` are already reentrant. Spec it: any subsystem registered through this RFC MUST implement `Enable() error` such that calling it after `Disable()` yields a working instance. That's the contract; "destroy and recreate" stays a visor-restart-only operation.

**Open question 3 (naming) — plain `hypervisor` / `dmsgpty` / `dmsgweb` / `skynetweb`.** Prefixing (`_service.X`) makes the CLI verbose for the common case to defend against a theoretical collision. Cheaper to add a small reserved-name set to `launcher.RegisterApp` (reject if user-installed app shadows a reserved subsystem name) than to wear the prefix forever.

**Open question 4 (default RestartPolicy for subsystems) — `RestartOnFailure` with exponential backoff and visible counter.** Operator intent is unambiguous: hypervisor crashing should come back, not stay gone. The crash-loop-invisibility worry is real but cheap to solve: surface `restart_count` + `last_fail_reason` + `last_fail_time` in `cli visor app ls`, and make the backoff cap (say 30s) inspectable. A subsystem that's restarted 47 times in the last hour will be obviously visible — more visible than today's "the hypervisor is just gone" state, where there's nothing to look at.

**Not on the checklist but flagging:**

- **Open question 2 (synthetic config visibility) — yes, project them into `config gen` output but mark as derived.** Hide them and operators can't tell why their `cli visor app ls` shows entries they didn't write; expose them as editable and we have a dual-source-of-truth ambiguity. The middle path is to emit them with a comment line like `# auto-derived from conf.Hypervisor.Enable=true; edit conf.Hypervisor to change` and have `config gen` rebuild the synth section on every run. Visible, but unambiguously not authoritative.

- **Open question 5 (external-mode internal services) — keep the door open, don't build the door.** OOM isolation for hypervisor is a credible use case once a visor is supervising many sub-hypervisors (the multi-section model from #2783/#2784 makes this not academic). Option B's facade shape doesn't preclude later promoting any of these to actual `RunMode: External` — the launcher already supports it. Worth a one-line note in phase 3.5's PR body that this capability falls out for free.

**One small caveat on phase 3.4:** the hypervisor's `Enable(ctx)` currently exits the goroutine when its HTTP server returns; under Option B, `<-ctx.Done()` in the AppFunc body would also need to call `Disable()` and propagate any in-flight handler errors. Probably already the plan but worth being explicit in the phase-3.4 PR.

LGTM on the RFC as the basis for proceeding. Ship phase 3.1.

PR #2882, opened 2026-05-28
docs: RFC — operator-programmable routing policy (Starlark)

0pcom:
## Summary

Captures the design space for letting operators write arbitrary per-dial routing decisions in an embedded scripting language. Driven by the operator framing *"we are really lacking a good interface for multihop/multiplexed routing control, a way to show routes in use by client app, well-defined automatic routing modes, and manual controls that don't fight automatic modes."*

The proposal: a small Starlark policy that the visor calls per dial. Operator writes a `decide_route(ctx, candidates) → RouteSpec` function; the visor evaluates it and uses the result to drive route selection. Examples like *"only routes through Indonesia on Friday at 5pm"* and *"mux=4 for vpn but not skychat"* are expressible directly.

## Highlights

- **Two-layer model.** Starlark for Layer 1 (per-dial, ~10ms budget). Layer 2 (per-packet) deferred — Layer 1's `RouteSpec` returns a distribution descriptor and the per-packet Go fast path stays compiled. Future Layer 2 scripting (CEL or bytecode) is a separate RFC if/when operators demonstrate need.
- **Language tournament** compared Starlark against CEL, Expr, Lua, Goja, Tengo, Risor, Yaegi, Rego, WASM, and clean-room mini-mcl. Tightest matchup was Starlark vs. Lua; Starlark wins on safety-by-default and on matching the [mgmt mcl aesthetic](https://github.com/purpleidea/mgmt) that prompted the proposal almost 1:1.
- **Per-app overrides global.** `conf.Routing.PolicyPerDial` is the visor-wide default; `AppConfig.RoutingPolicy` supplants it for that app.
- **mcl-flavored stdlib**: `datetime`, `transports`, `geo`, `peers`, `logging`. Geoip already embedded in the binary, just surfaced.
- **Failure-safe.** Panic, timeout, garbage return → fallback to built-in default. Operator can configure `policy_failure_mode: "drop"` to surface broken policies loudly instead.
- **File-watcher hot-reload.** Operator edits their policy; visor picks it up without restart.
- **Operator tooling**: `cli route policy test` (preview a decision without dialing), `bench` (1M-call p50/p99 budget check at load).

## Phased implementation

Five PRs:

1. **Go scaffold** — `RoutingContext`, `Candidate`, `RouteSpec` types; no-op policy hook in dial path behind feature flag.
2. **Starlark integration** — `go.starlark.net` wired in; scripts parseable + callable; timeout enforced; failure-mode wired.
3. **mcl-flavored stdlib** — `datetime`, `transports`, `geo`, `peers`, `logging` modules exposed. Indonesia-Friday example works end-to-end.
4. **Operator tooling** — `cli route policy test` + `bench`; hot-reload watcher; per-app override field.
5. **Layer 2 distribution descriptor** — `RouteSpec.distribution` parsed and translated into the per-packet Go fast path.

## Decision checklist

- [ ] Confirm Starlark as Layer 1 language (vs. Lua, the runner-up).
- [ ] Confirm the two-layer model (Starlark Layer 1 + deferred Layer 2 distribution descriptor).
- [ ] Confirm per-app overrides global.
- [ ] Resolve open questions 4 and 5 (Layer 2 contract finalization + CLI/policy interaction) before phase 1 lands.
- [ ] Pick the v1 stdlib surface.

Phases should not start until the above are settled.

## Test plan

- [x] Markdown renders correctly on GitHub.
- [ ] Team review + decision on the five-item checklist.

### DIFF
--- a/docs/routing_policy_rfc.md
+++ b/docs/routing_policy_rfc.md
@@ -1,0 +1,233 @@
+# RFC: Operator-Programmable Routing Policy
+
+Status: **Draft.** Lays out the design space for letting operators write arbitrary per-dial routing decisions in an embedded scripting language. Implementation deferred until the design is agreed.
+
+Related work: this builds on the unified-app-framework groundwork (RFC #2863, PRs #2860–#2879). It's downstream of that work but not blocked on it.
+
+## Background
+
+Today the visor's routing decisions are driven by a handful of CLI flags and per-dial overrides:
+
+- `--routes N` — number of parallel mux legs
+- `--min-hops K` — reject direct paths below K
+- `--forward-min-hops` / `--reverse-min-hops` — per-direction overrides
+- `--forward-mux` / `--reverse-mux` — per-direction leg counts
+- `--via PK` — pin a specific intermediate
+
+These are useful for ad-hoc experiments but fall short as a *policy* system:
+
+1. **No persistence.** Operators specify routing knobs on every CLI invocation. There's no per-app persistent policy that says "for skychat, always prefer routes through trusted intermediates."
+2. **No conditionality.** The knobs are static. There's no way to say "use mux=4 only during business hours" or "prefer routes through Indonesia on Fridays."
+3. **No composition.** A flag like `--min-hops 2` either applies or doesn't. Operators can't write "first try mux=4 over min-hops=2; if no candidate matches, fall back to direct."
+4. **No introspection.** The router picks a route opaquely. Operators can't see *why* a particular route was chosen, only that one was.
+5. **No per-app default differentiation.** A skywire visor running both skychat and vpn-client today applies the same routing posture to both, even though they have very different requirements (low-latency interactive vs. high-throughput streamed).
+
+Operator framing: *"we are really lacking a good interface for multihop/multiplexed routing control, a way to show routes in use by client app, well-defined automatic routing modes, and manual controls that don't fight automatic modes."*
+
+The proposal in this RFC: **let operators write a routing policy in a small embedded scripting language that the visor evaluates at dial time.**
+
+## Goals
+
+1. **Arbitrary expressiveness.** The operator can write any decision logic that maps `(dial context, candidate routes) → chosen route`. Examples that must be expressible:
+   - "Only routes through Indonesia on Friday at 5pm"
+   - "Mux=4 if peer is in same datacenter, mux=1 otherwise"
+   - "Avoid intermediates I've had >3 failures with in the last hour"
+   - "For skychat: prefer low-latency direct routes; for vpn: prefer high-bandwidth multi-hop"
+2. **Per-app override of a visor-wide default.** Each app can carry its own policy that supplants the global. The script always sees `ctx.app` so a single global can branch internally if the operator prefers.
+3. **Deterministic, fast, sandboxed.** Per-dial budget is ~10ms. The policy must never block the visor, never do I/O, never hang on infinite loops.
+4. **Hot-reload.** Operator edits their policy file; visor picks it up without restart.
+5. **Failure-safe.** If the policy panics, times out, or returns garbage, the visor falls back to the current built-in default selection and logs the violation. A broken policy never breaks dialing.
+
+## Non-goals
+
+- **Not** a full Turing-complete runtime for the operator to do arbitrary work. The policy is a *decision function*, not a daemon.
+- **Not** replacing the existing flags. CLI flags remain as per-invocation overrides; the policy is the persistent default.
+- **Not** building a new language. The aesthetic the operator wants is achievable in existing embedded-script ecosystems.
+- **Not** per-packet decision-making in the first pass. Per-packet decisions need a different perf tier (<10µs vs. <10ms); they're a deferred layer (see below).
+- **Not** distributed policy synchronization. Each visor owns its own policy file. Cluster-wide reconciliation is out of scope.
+
+## Design
+
+### Two layers
+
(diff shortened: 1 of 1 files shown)


0pcom:
Reviewed. Strong agreement on the overall shape; this is the right next step after the asymmetric-routing CLI chain (#2782/#2785/#2792/#2797) — those flags are the lowest-level operator handle, and a policy language is the natural composition surface above them. Walking the decision checklist:

**Starlark over Lua — confirm.** The safety-by-default property does outsized work here. A policy script runs inside the dial hot path; "sandbox is the design, not a list of removed APIs" is the difference between "operator wrote a bad policy" and "operator wrote a bad policy that did I/O to a file the visor doesn't have permission to touch." Lua's familiarity edge is real but smaller than it looks — gopher-lua's idioms diverge from PUC-Lua's in non-obvious ways, so an operator who "knows Lua" still hits surprises. Starlark's lookalike-to-mcl mgmt aesthetic is gravy on top. Bazel-precedent argument is what I'd lean on hardest if the team needs convincing: years of hostile CI workloads + Google's maintenance burden is exactly the threat model an operator-edited script file has.

**Two-layer model (Starlark Layer 1, deferred per-packet Layer 2) — confirm.** Putting Starlark in the per-packet path would be a category error; 10µs vs. 10ms is two interpreter generations apart. Distribution-descriptor approach is the right escape valve — operator gets to *configure* per-packet behavior from Layer 1 without paying interpreter cost per packet. The deferral of Layer 2 to a separate RFC is the honest choice.

**Per-app override (single hierarchy, app wins) — confirm.** The script-sees-`ctx.app`-so-one-global-can-branch escape hatch is what makes this not a maintainability disaster. Don't grow this to (app × peer-class × time-of-day) tuples; let the script handle the combinatorics.

**Open Q 4 (Layer 2 distribution descriptor vocabulary) — needs at least two more entries beyond `round-robin` / `weighted-list` / `size-threshold`:**
- **`hash-by-flow`** — packets in the same TCP/UDP flow pinned to the same leg. *Essential* for TCP-carrying multipath, because TCP fast-recovery breaks if a flow's segments arrive reordered across legs with different RTTs. The cli skynet path used by vpn-client carries TCP; round-robin distribution across multi-hop legs with 200ms+ RTT spread will collapse cwnd just like the Indonesia intermediate did to single-leg multi-hop pre-#2782. Without `hash-by-flow`, the operator's natural impulse — "spray across 4 legs for more throughput" — actively hurts.
- **`latency-bound`** — pick the leg with currently-lowest measured RTT for each packet. Compatible with the per-direction-measured-latency surface from #2782's ranker. Useful for interactive vpn / skychat where total throughput is irrelevant but tail latency matters.

I'd nail the vocabulary down in this RFC rather than deferring — `RouteSpec.distribution` is part of the type contract the Go code starts emitting in phase 1, and adding `hash-by-flow` later means a breaking change to existing policies.

**Open Q 5 (CLI flag / policy interaction) — explicit \"CLI wins unless the policy opts out via `ctx.cli_overrides`\".** This matches the operator-emergency-override mental model. A paranoid policy author can still write `if \"min_hops\" in ctx.cli_overrides: return RouteSpec(fallback=\"reject\")` if they want airtight policy enforcement. Don't make CLI silently-ignored when a policy is loaded; that's the kind of footgun where an operator types `--via PK` at 3am to debug a hung dial and the policy quietly throws it away.

**v1 stdlib surface — add `ctx.dst_pk` and `ctx.dst_port`.** The example policy filters by `c.hops_geo` on each candidate, but the *destination* PK is invisible to the script — and policies like \"never route trust-class-C destinations through trust-class-A intermediates\" or \"vpn-server destinations get mux=4, vpn-client destinations get mux=1\" need both sides. Cheap to add now; awkward to bolt on later when policies exist that assume the current shape.

**Open Q 1 (trace / audit) — should land in phase 4, not be deferred indefinitely.** The phrasing \"deferred to post-MVP per operator direction\" is fine as long as \"post-MVP\" means \"phase 4 alongside the rest of the operator tooling.\" If trace ships meaningfully later than the policy engine itself, the first time someone uses a policy for a security-sensitive routing decision and the auditor asks \"why\" the answer is \"we can't tell you,\" and that's a credibility hit the project doesn't recover from cheaply. The ring-buffer-readable-via-RPC sketch is the right minimum.

**Open Q 2 (versioning) — define the header now, even if the version compatibility logic ships later.** A `# starlark_compat: 2026.05` line at the top of the file is one line of code to *parse* in phase 2 (just store it; do nothing with it). When the stdlib eventually grows or shrinks, you have the version number to gate against. If you don't reserve that line now, when you need it later you'll discover existing policies in the wild that have a comment in that exact spot doing something else, and migration is a mess.

**Phase order — approve as written.** Phase 1 (Go-side scaffold with feature-flag-gated no-op) is the right insulation — the existing dial path stays unchanged in the no-policy case, so we can land it and observe nothing happens in production before we land anything that *does* something. Phases 2–5 walk a clean dependency: Starlark binding → stdlib → tooling → distribution-descriptor.

**One implementation caveat for phase 2:** the timeout enforcement via `starlark.Thread.SetMaxExecutionSteps` is a step-count budget, not a wall-clock budget. A script doing a tight `for i in range(10000000): pass` hits the step limit fine; a script doing one `transports.history(pk)` call where the host implementation has a slow code path inadvertently does not. Wall-clock timeout via `context.Context.Deadline` checked in each stdlib host function is the belt-and-suspenders. Worth being explicit in the phase 2 PR.

LGTM as the basis for proceeding. The hash-by-flow + ctx.dst_pk additions are the only ones I'd want pinned down before phase 1 lands.

PR #2917, opened 2026-05-29
docs(mainnet_rules): bandwidth pool is sender-pays after #2886

0pcom:
## Summary

\`#2886\` reshaped the bandwidth (pool 2) reward path:
- Each transport's sent counter is recorded per edge, capped by the counterparty's recv confirmation (sender-pays, verified)
- At calc time, an edge is credited iff that edge itself passed the pool 1 eligibility check; counterparty eligibility doesn't gate the sender's credit because credit only flows to the sender anyway

The rules document still described the old "both edges share the total" model. This brings the prose in line with what the calculator actually does.

## Changes

- **Pool 2 section**: changed "transport bandwidth each visor handled" → "each visor sent", added a paragraph on the sender-pays cap and on per-edge independent evaluation.
- **Transport Bandwidth and Metrics section**: reframed the "Both edges of a transport must agree" sentence as the sender-side cap that it actually is now.

No code changes; docs only.

### DIFF
--- a/rewards/mainnet_rules.md
+++ b/rewards/mainnet_rules.md
@@ -41,7 +41,9 @@ Regional saturation scaling is applied to the presence pool to promote geographi
 
 ## Pool 2: Bandwidth
 
-The bandwidth pool reward for a day is distributed proportionally based on the amount of transport bandwidth each visor handled during the previous day. Only visors which handled bandwidth above a minimum threshold are eligible for the bandwidth pool.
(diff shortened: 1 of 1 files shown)


0pcom:
Superseded by #2920 — that PR folds in the sender-pays rewrite plus retires the now-stale CSV log claims, adds the v1.3.59 minimum-version row, and notes the Skycoin-precision floor on the bandwidth pool. Closing this in favor of the broader doc update.

PR #2941, opened 2026-05-31
fix(visor/dmsgtracker): resolve tracked peers via the direct client, not dmsg-discovery

0pcom:
## Problem

The dmsg tracker dials each connected peer on `DmsgCtrlPort` to read its dmsg server + round-trip for the hypervisor UI (`dmsgtracker.newDmsgTracker` → `dmsgC.DialStream`). It was wired to the **discovery-based** main client (`v.dmsgC`):

```go
// init_dmsg.go — before
dmsgC := v.dmsgC
dtm := dmsgtracker.NewDmsgTrackerManager(v.MasterLogger(), dmsgC, 0, 0)
```

So every dial first resolved the peer's **dmsg-discovery entry**. But peers that reach the hypervisor over a skywire transport don't register in dmsg-discovery — services and direct clients don't, by design. Those lookups failed on **every UI poll**:

```
[dmsg] disc Entry: DMSG primary failed; trying HTTP fallback
  error="...entry/<peer>: dmsg error 202 - cannot connect to delegated server"
dtm.establishTracker: Skipped dmsgtracker client (peer not in discovery...)
  error="dmsg error 100 - entry is not found in discovery"
```

re-fired every few seconds with no backoff (the tracker never establishes, so `GetBulk`/`ShouldGet` re-attempt each poll). As discussed on the issue, this isn't log spam — it's genuinely **wrong, discovery-dependent connection attempts** for peers that should be reached directly. The codebase already avoids exactly this for service lookups by using the **direct client** (`v.dmsgDC`, `direct.Client`-backed with synthetic entries — see `dmsgfirst` and `init_router.go`); the tracker just wasn't using it.

## Fix

Build the tracker against the **direct client** (`v.dmsgDC`) and register a synthetic entry for each tracked peer right before dialing — all known dmsg servers as delegated, mirroring how deployment service PKs are seeded (`direct.GetAllEntries`). `DialStream` then tries those servers directly, no discovery round-trip.

- **`dmsgtracker.Manager`**: optional `SetPeerResolver` hook, invoked with the peer PK just before the dial (race-free, read under the manager mutex). Broaden `isExpectedTrackerLookupErr` so a direct-dial miss (`ErrNoAvailableServers` / "cannot connect to delegated server" — peer connected to us but not currently on a shared dmsg server) stays at debug, not warn.
- **`Visor`**: stash the direct server set (`dmsgDirectServers`); add `ensureDirectDmsgEntry(pk)` to register synthetic peer entries on `v.dClient`.
- **`initDmsgTrackers`**: construct the manager once `dmsgDC` is ready (bounded 60s wait via `dmsgHTTPReady`; falls back to the discovery client if dmsg-http never comes up — e.g. no servers), and wire the resolver only on the direct path.

Net effect: for peers that are dmsg-reachable on the shared server set, the tracker now **establishes** (and then just refreshes every 30s) instead of re-failing a discovery lookup every poll.

## Tests

- `TestEnsureDirectDmsgEntry` — a previously-unknown peer becomes resolvable on the direct client, delegating the known server.
- `TestEnsureDirectDmsgEntry_NoServers` — no-op / no panic when the direct client isn't up.
- `go build ./...`, `go vet`, `golangci-lint` clean; `pkg/visor` + `pkg/visor/dmsgtracker` (incl. `-race`) pass.

## Note

This fixes the *resolution path*. If a managed visor genuinely has **no** live dmsg session on any shared server (reachable only via stcpr/sudph), the direct dial still can't reach it — but that now fails fast locally (no discovery round-trip) and stays at debug, rather than hammering discovery every poll.

### DIFF
--- a/pkg/visor/dmsg_direct_entry_test.go
+++ b/pkg/visor/dmsg_direct_entry_test.go
@@ -1,0 +1,60 @@
+package visor
+
+import (
+	"context"
+	"sync"
+	"testing"
+
+	"github.com/skycoin/skywire/pkg/cipher"
+	"github.com/skycoin/skywire/pkg/dmsg/direct"
+	dmsgdisc "github.com/skycoin/skywire/pkg/dmsg/disc"
+	"github.com/skycoin/skywire/pkg/logging"
+)
+
+// TestEnsureDirectDmsgEntry verifies that the dmsg-tracker resolver hook
+// makes a previously-unknown peer resolvable on the direct-client path,
+// advertising the known dmsg servers as delegated — so the subsequent
+// dial doesn't depend on the peer having a dmsg-discovery entry.
+func TestEnsureDirectDmsgEntry(t *testing.T) {
+	srvPK, _ := cipher.GenerateKeyPair()
+	servers := []*dmsgdisc.Entry{{Static: srvPK}}
+
+	// Direct client seeded with the server only — not the peer.
+	dc := direct.NewClient(direct.GetAllEntries(cipher.PubKeys{}, servers),
+		logging.NewMasterLogger().PackageLogger("test"))
+
+	v := &Visor{
+		log:               logging.NewMasterLogger().PackageLogger("test"),
+		initLock:          new(sync.RWMutex),
+		dClient:           dc,
+		dmsgDirectServers: servers,
+	}
+
+	peerPK, _ := cipher.GenerateKeyPair()
+
+	// Precondition: peer is not resolvable on the direct client.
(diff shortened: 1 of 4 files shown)


0pcom:
Closing — wrong layer. Root cause is service-side: the dmsg-discovery's sessions to overloaded official dmsg servers go half-open and linger ~120s (pingDeadThreshold=2 × 60s ping) before eviction/reconnect, so it advertises servers in /health it can't actually forward through. Fixing dead-session detection latency instead.

PR #2954, opened 2026-06-01
fix(router): TTL cache for TPD GetTransportByID in buildHopLookups (skynet --routes N --min-hops K)

0pcom:
## Summary

`skynet-client --routes N --min-hops K` blows through TPD's 30-req/min rate limit on every dial, causing `buildHopLookups` to silently empty out, downstream filters can't rank/reject, and the route group never assembles. Local listener binds but TCP connections RST.

## Reproducer (verified live)

```
iperf3 -c 127.0.0.1 -p 5212 -t 30 -P 4 over a skynet client started with
  skywire cli skynet start --pk <peer> --remote 5210 --local 5212 \
    --routes 8 --min-hops 2 --forward-mux 8 --reverse-mux 8

→ 'unable to receive control message - port may not be available,
   the other side may have stopped running, etc.: Connection reset by peer'
→ 0 bytes, intervals=[], connected=[]
```

Visor log on initiator (debug level):
```
[router]: Found routes Forward: [...] (3 candidates × 2 hops × {fwd,rev})
[router]: buildHopLookups: GetTransportByID failed for c601d3b6... error="429 Too Many Requests: Rate limit exceeded: maximum 30 requests per minute"
... (× ~12 per attempt, × 3 retries)
```
`cli rg ls` — empty (no route group materialized).

## Root cause

Each dial attempt resolves 12+ unique TpIDs (3 candidates × 2 hops × {forward, reverse}). With 3 retry rounds, ~36 lookups land at TPD within ~1 second. TPD's per-client rate limit is 30 req/min. After the first ~30, every subsequent call returns 429 → `buildHopLookups` logs at Debug + continues with empty latency/type → downstream `rejectDMSGMultihop` and the latency ranker get nothing useful → no path is picked → no route group registered.

## Fix

Router-scoped TTL cache for `GetTransportByID` (`pkg/router/tpd_cache.go`):
- **Positive TTL: 60s** — TPD entries are republished on state change; a minute of staleness is well within the freshness budget for routing decisions.
- **Negative TTL: 5s** — a single rate-limited burst doesn't cascade into a retry storm; TPD recovery is noticed within seconds.

`buildHopLookups` in `pkg/router/router_dial.go` routes its TPD calls through `cache.GetOrFetch`. Concurrent reads are RWMutex-guarded.

No behavior change when DiscoveryClient is nil (cache fully skipped). No API change. No new deps.

## Tests

`pkg/router/tpd_cache_test.go` covers:
- positive cache reduces fetch count from N to 1 within posTTL
- negative cache suppresses retry storms within negTTL
- failure → success heals cleanly
- concurrent reads are safe

Full `pkg/router` test suite still passes (~38s).

## Cred

Live-reproduced today by Beta + Gamma. Beta predicted the cache-shape fix earlier in another thread ("visor-wide TTL cache is the right shape"); this implements that prediction with the diagnosis evidence.

### DIFF
--- a/pkg/router/router.go
+++ b/pkg/router/router.go
@@ -412,6 +412,7 @@ type router struct {
 	muxMode            WeightMode       // default weight mode for new mux connections
 	lastRouteCalcTime  time.Duration    // last route calculation time (for local routes)
 	lastRouteCalcMu    sync.Mutex       // protects lastRouteCalcTime
+	tpdCache           *tpdEntryCache   // visor-wide TTL cache for TPD GetTransportByID, see tpd_cache.go
 }
 
 // scopedLog returns a logger augmented with app_name=<n> when the
@@ -481,6 +482,7 @@ func New(dmsgC *dmsg.Client, config *Config, routeSetupHooks []RouteSetupHook) (
 		done:            make(chan struct{}),
 		trustedVisors:   trustedVisors,
 		routeSetupHooks: routeSetupHooks,
+		tpdCache:        newTPDEntryCache(),
 	}
 
 	go r.rulesGCLoop()
--- a/pkg/router/router_dial.go
+++ b/pkg/router/router_dial.go
@@ -998,9 +998,25 @@ func (r *router) buildHopLookups(ctx context.Context, fwd, rev [][]routing.Hop) 
 				continue
 			}
 		}
-		// TPD fallback — one fetch covers both lookups.
+		// TPD fallback — one fetch covers both lookups, served from
+		// the router's TTL cache when the entry is fresh. The cache
+		// is what keeps --routes N --min-hops K from blowing through
+		// TPD's 30-req/min rate limit on multi-hop dials: each dial
+		// retry pulls 12+ unique TpIDs across candidate paths, and
+		// without caching the bursts trivially exceed budget → 429s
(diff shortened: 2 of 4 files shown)


0pcom:
Holding this PR per @0pcom directive (DM 2026-06-01): Moses + Alpha reviewed and prefer the fix at a different layer — buildHopLookups should reuse a single `GetAllTransports()` (cached ~5 min, same shape as calculateLocalRoutes already uses) AND add `NumRoutes` to RouteOptions so the route-finder returns N ranked routes directly = the mux legs (no local re-rank, no TPD re-hammer).

Per-ID TTL cache here treats the symptom; the bulk+NumRoutes approach attacks the load shape. Alpha is implementing the replacement now. Marking draft. Keep open as reference for the diagnosis (the 12+ unique TpIDs × 3 retries → 429 cascade is the root cause description Alpha's PR will also need).

0pcom:
Superseded by #2955, which fixes this at the dataset layer (one cached GetAllTransports snapshot shared by buildHopLookups + calculateLocalRoutes) and additionally lifts the route-finder's hardcoded 3-route cap (RouteOptions.NumRoutes) so mux degrees above 3 work. Thanks for the root-cause diagnosis — credited in #2955.

PR #2968, opened 2026-06-02
feat(cli/gotop): multiload-ng-style display behind --multiload (CPU + RAM breakdown)

0pcom:
## What

Adds a single global display flag **`-M` / `--multiload`** (config key `multiload`) to the bundled `skywire cli gotop` that switches its graphs to a [multiload-ng](https://github.com/udda/multiload-ng)-style **breakdown by subtype**, instead of the single aggregate figure gotop shows today.

The **default display is unchanged**, and the pre-existing **`-a/--averagecpu`** keeps its original average-load meaning. The multiload look is opt-in, and one flag drives all the breakdowns (not a flag per graph).

### CPU (`-M`)
gotop's CPU graph only ever showed `100 - idle` per core — so a core pegged on **iowait** looked identical to one pegged on user compute. Under `-M` the CPU graph now shows per-state lines computed from `gopsutil cpu.Times` deltas:

`user · sys · iowait (io) · irq (+softirq) · steal` + a bold **total**.

### RAM (`-M`)
gotop's memory graph showed a single `used %`. Under `-M` it splits into `used · buff · cache` + a bold **total** (from `gopsutil VirtualMemory`), coloured as **shades of the colorscheme's memory hue** (light vs. dark) the way multiload tints one base colour per subtype — so reclaimable cache/buffers are visually distinct from application memory.

### Also
- **Label alignment:** line-graph value labels are now left-justified into a column (`drawLabels`) so the percentages line up instead of floating right after each variable-width series name. Applies to every line-graph widget (CPU/RAM/Net/Disk).

## Implementation
- `devices/cpu_breakdown.go`, `devices/mem_breakdown.go` — new collectors (snapshot-delta for CPU; component %s for RAM). No new dependency — gopsutil was already vendored.
- `widgets/cpu.go`, `widgets/mem.go` — `ShowBreakdown` mode; original modes untouched.
- `termui/linegraph.go` — `Tint()` shade helper + the label-column alignment.
- `config.go`, `layout/layout.go`, `cmd/skywire-cli/commands/gotop/root.go` — the `multiload` flag/config plumbing.

## Scope / follow-ups
Confined to the vendored gotop + its CLI command. Disk-I/O throughput and Network are the natural next graphs to give the same `-M` treatment (will stack onto this branch).

## Test
- `go build ./...`, `go vet`, `golangci-lint` clean on the touched packages.
- Verified live: default, `-a`, and `-M` all render without panic; `-M` shows the CPU per-state breakdown (iowait visible) and the RAM used/buff/cache breakdown.

### DIFF
--- a/cmd/skywire-cli/commands/gotop/root.go
+++ b/cmd/skywire-cli/commands/gotop/root.go
@@ -51,6 +51,7 @@ var (
 	fahrenheit   bool
 	percpu       bool
 	averagecpu   bool
+	multiload    bool
 	statusbar    bool
 	exportPort   string
 	mbps         bool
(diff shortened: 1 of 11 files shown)


0pcom:
image

```
skywire cli gotop --local
```

image


```
skywire cli gotop --local -M
```

PR #2978, opened 2026-06-03
fix(dmsg): detect zombie yamux sessions via stream-level liveness ping

0pcom:
## Problem

`SessionCommon.Ping()` — called by `pingSessionsLoop`'s dead-session detector — used `yamux.Session.Ping()` for yamux sessions. That's a **control-frame keepalive**: it round-trips at the mux-control layer over the existing TCP connection and proves only that the client↔server link is alive. It does **not** exercise the server's stream-forwarding path.

So a **zombie session** — control link healthy, but the dmsg server can no longer relay streams to/from us — passes the ping forever, never reaches `pingDeadThreshold`, and is never closed/redialed. `pingSessionsLoop`'s own doc comment already said yamux keepalive "does NOT detect the case where our side of the session is healthy but the dmsg server has torn down its view" and claimed an app-level ping covers it — but the yamux path never opened a stream, so the implementation contradicted its comment.

## Production confirmation

Found via live debugging of a route **setup-node** (a dial *sink* — every route setup dials *into* it). Decisive per-server experiment:

- **0/8** stream-dials to the node succeeded ("i/o deadline reached" on every delegated server)
- **8/8** `yamux.Ping()` reported the sessions alive; **0** "Closing dead session" events in 6h
- pprof: 8 active yamux sessions, healthy goroutines, clean ESTAB sockets

All 8 sessions were simultaneously zombie, so zero routable bytes could land — the node stayed unreachable until a manual restart. Dial *sources* (visors) barely notice the same zombies because they route around their own dead sessions when dialing out; a dial sink can't, which is why this bit the setup-node specifically and recurringly.

## Fix

The **smux** path (`smuxPing`) already did the right thing: open a stream, write the ping marker, read the server's echo. This brings **yamux** to parity with `yamuxPing()`, which round-trips through the server's `serveStream` ping-marker echo (which handles yamux and smux streams identically — server_session.go). Now the existing dead-session logic can see and recycle zombie yamux sessions, and the close+reconnect refreshes the discovery entry so the node stops advertising dead servers.

Scope: single site (`SessionCommon.Ping()`), inherited by **every** dmsg client (visors, setup-nodes, transport-setup, embedded RSN, all services) since they share `dmsg.Client`.

## Test

`TestSessionPing_YamuxStreamRoundTrip` spins up a real server + yamux client and asserts `Ping()` returns a positive-RTT stream round-trip — proving the new path works against the server echo and that healthy yamux sessions are not false-killed. Full `pkg/dmsg/dmsg` suite passes under `-race`.

### DIFF
--- a/pkg/dmsg/dmsg/session_common.go
+++ b/pkg/dmsg/dmsg/session_common.go
@@ -184,12 +184,52 @@ func (sc *SessionCommon) Ping() (time.Duration, error) {
 	sc.sm.mutx.RLock()
 	defer sc.sm.mutx.RUnlock()
 	if sc.sm.yamux != nil {
-		return sc.sm.yamux.Ping()
+		return sc.yamuxPing()
 	}
 	if sc.sm.smux != nil {
 		return sc.smuxPing()
 	}
 	return 0, fmt.Errorf("no mux session available for ping")
+}
+
+// yamuxPing implements ping over yamux the same way smuxPing does for smux:
+// open a temporary stream, write the ping marker, and wait for the echo.
+//
+// Why NOT yamux.Session.Ping(): that is a yamux *control-frame* keepalive —
+// it round-trips at the mux-control layer over the existing TCP connection
+// and proves only that the client<->server link is alive. It does NOT
+// exercise the server's stream-forwarding path, so it cannot detect a
+// "zombie" session where the control link is healthy but the dmsg server can
+// no longer relay streams to/from us (observed in production: a dial-target
+// service — e.g. a route setup-node behind a Docker bridge — kept advertising
+// 8 "delegated servers" on which 0/8 stream-dials succeeded, while
+// yamux.Ping() reported all 8 alive, so pingSessionsLoop never marked them
+// dead and the node stayed unreachable until a manual restart). Opening a real
+// stream and round-tripping the ping marker through the server's serveStream
+// echo (which handles yamux and smux streams identically) exposes exactly that
+// mismatch, so the existing pingDeadThreshold logic can close + redial the
+// dead session. The smux path already worked this way; this brings yamux to
+// parity.
+func (sc *SessionCommon) yamuxPing() (time.Duration, error) {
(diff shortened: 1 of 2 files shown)


0pcom:
Converting to draft. The production evidence that motivated this — "0/8 stream-dials succeed while yamux.Ping reports alive" — turned out to be a **measurement artifact**: the probe used `cli visor ping --dmsg` (targets `DmsgPingPort=8`), but the setup-node has no listener on port 8 (it listens on 80 + 136). A corrected probe to port 80 shows the node reachable (~211ms), and host-networking the container did not change the failure rate. So the zombie-session premise is not supported, and `yamux.Session.Ping()` does round-trip a control frame anyway.

The change brings yamux to parity with smux's stream-level ping and is arguably reasonable hardening, but it is **not** the fix for the observed route-setup failures and carries a small rollout/compat cost (yamux clients begin depending on the server-side pingMarker echo). Holding pending a re-diagnosis — the real failure is the setup-node not reaching remote *intermediate* visors (dmsg-100/202), i.e. the route-finder serving routes through unreachable/orphan hops.

0pcom:
Closing. The motivating diagnosis — a 'zombie' dmsg session where yamux.Ping() reports a server alive while 0/8 stream-dials succeed — was traced to a port-8 measurement artifact, not a real stream-forwarding zombie. The RSN/setup-node was in fact reachable; the orphan-reachability symptoms were the ghost-TPD-edge issue fixed in #2979/#2980.

The code observation still stands on its own: yamux.Ping() is a control-frame keepalive that only proves the TCP link is alive, whereas smuxPing() round-trips a real stream — so the two mux paths have asymmetric liveness coverage. But absent a confirmed stream-forwarding zombie, opening+closing a real yamux stream on every pingSessionsLoop tick across all sessions (stream churn + server forwarding load + false-positive-dead risk) isn't justified.

Approach preserved in history; happy to revive + de-risk (e.g. stream-ping only after repeated control-ping anomalies) if a genuine stream-forwarding zombie is ever confirmed.

PR #3073, opened 2026-06-11
feat(router,proxy): mux-auto GROW — adaptive multiplexed-route failover

0pcom:
## What

`mux-auto` previously only **PRUNED** excess/slow legs (#3055). This adds the **GROW** direction: when live legs drop below the preset's target count (`fastest`=2, `balanced`=4, `resilient`=8), it adds fresh **disjoint** legs so a multiplexed proxy session self-heals lost redundancy. Adaptive failover now works both ways.

## Design

New router method `GrowMuxRoute(desc, target, minHops)` — the runtime, additive counterpart to `establishMuxRoutes` (which plans the initial leg set at dial time). It:
1. computes the live legs' intermediate exclude-set (`intermediatesOfRouteGroup`),
2. plans `target - current` disjoint legs via `fetchBestRoutes` (local-calc fallback) with `ExcludeIntermediatePKs`, accumulating each leg's intermediates so successive legs diverge,
3. adds each via the existing `AddMuxRouteByHops` actuator.

- **Legacy path only** — uses `RouteGroupDialer.Dial` (setup nodes), **not the cascade**.
- **Visor-side planning** — `mux-info` only exposes each leg's *first-hop* PK, not the full hop chains, so the CLI can't build a correct disjoint exclude-set; the router can.
- `--min-hops` (default 2) floors the added legs' hop count so a multihop mux grows with multihop legs.
- `GrowMuxRoute` is **only reachable via the new mux-auto RPC** → zero effect on the normal routing path.

Wired: router method → `Router` interface + MockRouter → visor `API` + impl → RPC server/client/mock → proxy stub → `mux-auto` tick() (prune → grow) + summary line.

## Validation status

✅ `go build ./...` clean · ✅ gofmt clean · ✅ golangci-lint clean (only pre-existing usermanager gosec warnings remain) · ✅ local visor clean boot.

⚠️ **Not yet exercised live** — the leg-adding behavior needs a running mux proxy session (`skysocks-client --routes N`, prune a leg, `proxy mux-auto balanced --watch 5s`, observe `+ grew N leg(s)`). Recommend a live validation run before merge.

Follow-up (deferred): visor-integrate so the initial dial skips `fetchBestRoutes` when explicit hops are supplied (router_dial.go:1762).

### DIFF
--- a/cmd/skywire-cli/commands/proxy/mux_auto.go
+++ b/cmd/skywire-cli/commands/proxy/mux_auto.go
@@ -95,6 +95,18 @@ Example:
 			internal.PrintFatalError(cmd.Flags(), fmt.Errorf("unknown preset %q (want: fastest|balanced|resilient)", args[0]))
 		}
 
+		// GROW dials fresh disjoint routes server-side, which can exceed
+		// the 30s default RPC deadline when a candidate route is slow to
+		// set up. Without headroom the client gives up while the visor is
+		// still establishing the leg — the leg lands but the tick
+		// misreports "context deadline exceeded" / "grew 0". Give the
+		// loop's client room to cover GrowMuxRoute's server-side budget;
+		// the fast prune/info calls return immediately regardless. Only
+		// override the untouched default so an explicit --timeout wins.
+		if clirpc.Timeout == 30 {
+			clirpc.Timeout = 120
+		}
+
 		rpcClient, err := clirpc.Client(cmd.Flags())
 		if err != nil {
 			internal.PrintFatalError(cmd.Flags(), fmt.Errorf("unable to create RPC client: %w", err))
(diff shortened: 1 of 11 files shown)


0pcom:
**Live-validated on the local visor (skysocks-client mux session → 030baa87, --min-hops 2):**
- **GROW**: 3→4 legs (`+ grew 1`, new leg via a disjoint intermediate 0261d954); 6→8 legs (`+ grew 2`).
- **PRUNE** still correct: 5→4 (`pruned 1`).
- **Slow-GROW fix**: a 6→8 grow took **39.5s** — it would have false-reported `context deadline exceeded` / `grew 0` under the old 30s RPC client deadline (the leg lands server-side, the CLI gave up early). Follow-up commit bumps mux-auto's client timeout to 120s (covers GrowMuxRoute's 90s budget; fast prune/info calls unaffected; explicit --timeout wins). After the fix it reported `grew 2` correctly.
- ✅ `go build ./...` · ✅ gofmt · ✅ golangci-lint (changed pkgs) · ✅ `go test ./pkg/router ./pkg/visor` green · ✅ clean boot.

Scope note: a fully-collapsed session (all legs incl. primary die) can't be grown — that's a proxy restart by design; GROW maintains an *existing* rg's redundancy as legs flake.

PR #3182, opened 2026-06-19
feat(hypervisor): hv ui enable/disable — separate web UI from RPC + hv ls (also fixes tpviz restart panic)

0pcom:
Per the observation that disabling the hypervisor shouldn't also kill CLI access to connected-visor data: `hv disable` tore down the DMSG-RPC listener, managed-visor tracking, AND the web UI together — so it also broke `hv ls` (which gates on `IsEnabled`). You couldn't shrink the public web attack surface while keeping secure CLI access to connected visors.

## Change
Split web-UI serving from the hypervisor's RPC/tracking surface:
- `hv enable`/`hv disable` (unchanged) toggle the whole hypervisor — DMSG-RPC listener + managed-visor tracking + (unless `ui_disable`) the web UI.
- **New `hv ui enable` / `hv ui disable`** (with `-w` to persist) toggle ONLY the web UI (`:http_addr`). The DMSG-RPC listener, managed-visor tracking, and `hv ls` keep working.
- New config field `hypervisor.ui_disable` (default false = served; backward-compatible) persists the UI state.

Implementation: `uiServing` flag separate from `enabled`; extract `startUI`/`stopUI`; `Enable` starts the UI only when `!ui_disable`; `EnableUI`/`DisableUI`/`IsUIServing` on the API + RPC server/client + proxy/mock + CLI.

## Also fixes a pre-existing tpviz restart crash
`tpviz.Server.Stop()` does `close(s.stopChan)` with no double-close guard, and `stopChan` is created once in `NewServer` (never recreated) — so any enable→disable→enable cycle (and the new UI toggle) double-closed it → **panic**. tpviz is now started at most once (`sync.Once`) for the hypervisor lifetime and not stopped on UI toggle (it's a cheap background cache refresher; its routes are unreachable once the HTTP server is down anyway).

## Validation
Built/vet/lint green. Live on a hypervisor: `hv ui disable` drops `:8000` while `hv ls` still returns 18 connected visors and the process stays up; repeated enable/disable toggles are stable (no double-close panic).

## Follow-up (noted, not in this PR)
Surface the PKs that hit the hypervisor dmsg port and get `dmsg 306 (no listener)` as a "pending/attempting" list, so you can see would-be managed visors even when the listener is down.

### DIFF
--- a/cmd/skywire-cli/commands/config/gen.go
+++ b/cmd/skywire-cli/commands/config/gen.go
@@ -1144,9 +1144,9 @@ func configureLauncher(log *logging.Logger) {
 		BinPath:       skyenv.AppBinPath,
 		DisplayNodeIP: isDisplayNodeIP,
 	}
-	conf.UptimeTracker = &visorconfig.UptimeTracker{
-		Addr: services.UptimeTracker, //utilenv.UptimeTrackerAddr,
-	}
+	// The standalone uptime-tracker is deprecated (uptime is now tracked by the
+	// discovery services). Generated configs no longer carry an `uptime_tracker`
+	// block — conf.UptimeTracker stays nil and initUptimeTracker skips it.
 	if cliAddr != "" {
 		conf.CLIAddr = offsetAddr(cliAddr)
 	} else {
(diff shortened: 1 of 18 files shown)


0pcom:
Added the port-hit tracker per discussion: `cli dmsg port-hits` surfaces inbound dmsg requests to ports with no listener (dmsg 306) keyed by (src PK, dst port) with count + first/last-seen. Bounded (cap 256, evict-oldest) tracker on `EntityCommon`, recorded in the session serve loop; exposed via `Visor.DmsgPortHits()` RPC. Validated live — it cleanly lists every visor (incl. the managed SBCs) that hit the hypervisor :46 during the pre-enable window. Unit test covers coalesce/evict/order/nil-safety.

PR #3191, opened 2026-06-20
fix(visor): dmsg-only deployment services — rip out plain-HTTP fallback + autoconnect wedge

0pcom:
## Symptom
A user's **v1.3.77 Windows visor stopped creating transports** via public autoconnect after running a while (manual transport-setup-node creation still worked).

## Diagnosis (live, over the resolving SOCKS5 proxy)
pprof goroutine dump of the visor showed its autoconnect goroutine **wedged at `autoconnect.go:446`** — inside `fetchPubAddresses`'s retrier, in the clearnet HTTP service-discovery fetch (`a.client.Services`).

## Root cause
`NewDefaultRetrier` uses `DefaultTries = 0` → **retry forever**. On a dmsg-only deployment the public-visor source is the CXO snapshot of the SD services tree; when that snapshot is momentarily empty, `fetchPubAddresses` falls through to the HTTP SD client. If that SD is unreachable (a clearnet `sd.skycoin.com` is dead on the dmsg-only deployment), `retrier.Do` **never returns** — the whole autoconnect `Run` loop wedges on a single tick. No public visors are fetched, no transports created, and since the CXO snapshot is (re)triggered at the *top* of `fetchPubAddresses` each tick, CXO never gets another chance to populate either.

## Fix
A **bounded** retrier (3 tries) + a **20s per-fetch timeout**. The `Run` loop's ticker (`PublicServiceDelay`) already provides the retry cadence, so the per-tick fetch fails fast: a dead SD degrades to "no public visors this tick", the loop moves on, and the next tick re-checks (and re-triggers) the CXO snapshot — **self-healing**.

## Validation
Builds + lints clean. The fix turns an unbounded retry into a bounded one, which definitionally prevents the wedge. Needs deployment to the affected visor / fleet to confirm transports resume; the diagnosis (wedged goroutine in the infinite retrier) is conclusive on the mechanism.

Follow-up worth noting: *why* the CXO SD snapshot went empty on that visor (the upstream trigger) is a separate CXO-subscription-lifecycle question — but this fix makes autoconnect resilient to it regardless.

### DIFF
--- a/pkg/visor/api_services.go
+++ b/pkg/visor/api_services.go
@@ -74,44 +74,36 @@ func (v *Visor) ServiceHealth() ([]ServiceHealthEntry, error) {
 		}{"Uptime Tracker", svcURLs{v.conf.UptimeTracker.Addr, v.conf.UptimeTracker.AddrDmsg}})
 	}
 
-	// Use the visor's direct DMSG HTTP client first (v.dmsgHTTP). It
-	// connects through a direct client with pre-loaded entries for all
-	// deployment services — no discovery lookup needed. Falls back to
-	// HTTP if the direct client isn't ready or the probe fails.
-	httpClient := &http.Client{Timeout: 10 * time.Second}
+	// Probe every service over dmsg only — plain HTTP to deployment services is
+	// no longer supported. v.dmsgHTTP connects through a direct client with
+	// pre-loaded entries for all deployment services (no discovery lookup). The
+	// dmsg URL may live in either field: the dmsg-only default config stores it
+	// in the "http" field (e.g. service_discovery = "dmsg://..."), legacy dual
+	// configs use the *_dmsg field — pick whichever holds the dmsg:// URL.
 	var dmsgClient *http.Client
 	select {
 	case <-v.dmsgHTTPReady:
 		dmsgClient = v.dmsgHTTP
 	default:
-		// Not ready yet; will use HTTP only.
+		// Not ready yet; services report N/A this poll.
 	}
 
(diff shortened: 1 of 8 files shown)


0pcom:
Expanded into the full **rip-out of plain-HTTP to deployment services** (per operator directive — plain HTTP to deployment services is no longer supported):

- **getHTTPClient** (central transport builder for AR/TPD/RF/SD/UT/TPS/disc): rejects any non-dmsg:// URL with a loud error instead of building a clearnet client.
- **initAddressResolver**: dropped the "DMSG-HTTP failed, using plain HTTP" fallback + the v6-forced clearnet client.
- **api_services**: `serviceFetchOrder`/`FetchServiceData` dmsg-only (also fixed a latent bug — the dmsg-only default stores the dmsg URL in the "http" field, which the old code GET'd over clearnet); `servicesFromHTTP`→`servicesFromDmsg`; **ServiceHealth** probes dmsg-only.
- **hypervisor UI proxies** (TPD metrics, reward system): dropped the plain-HTTP fallback steps.
- **autoconnect**: bounded SD retrier (the original wedge fix) + prefer-dmsg SD URL.

**Validated live**: visor boots clean; `svc health` shows all services OK over **dmsg** (Config/TPD/DMSG-D/AR/RF/SD); address resolver binds STCPR/SUDPH over dmsg. No clearnet anywhere.

Remaining (separate, dmsg-package layer): the dmsg client's own discovery upgrade (`dmsgfirst`) still carries an HTTP fallback — inert on dmsg-only (disc URL is dmsg://), but it's the dmsg-bootstrap path so it warrants its own careful PR.

PR #3285, opened 2026-06-25
feat(hv-ui): SkywireHttpBackend — one UI build for native REST or in-tab wasm (step 1)

0pcom:
Step 1 of the wasm-HV-SPA backend abstraction (`docs/wasm-hv-spa-backend.md`, #3284). Adds a custom Angular `HttpBackend` at the bottom of the `HttpClient` pipeline so the **same** hypervisor UI build can talk to either backend — **without editing `ApiService` or any feature service**:

- **native** (visor-served; default when `window.__SKYWIRE_HV__` is unset): a transparent pass-through to the real `HttpXhrBackend` — byte-for-byte prior behavior.
- **in-tab wasm / standalone / remote-viewer**: route `/api/*` to the gateway (`skywireVisor.hvApi` / `skywireDmsg.hvApi` / `skywireDmsg.fetch`), translating the request and the gateway's `{status,body,headers}` into an Angular `HttpResponse`/`HttpErrorResponse` (so `ApiService.errorHandler` 401→login still works). Non-`/api` requests (assets) delegate to XHR.

This is the Angular-native, DI-injected, unit-testable replacement for `override.js`'s fetch/XHR monkey-patch **routing**. The wasm-mode wiring (set `__SKYWIRE_HV__`, stop injecting `override.js`) + the gateway-boot path are **step 2**.

## Validated
- Angular production build green; the new backend type-checks and is wired via `{ provide: HttpBackend, useClass: SkywireHttpBackend }`.
- Rolled the running visor: it serves the **new bundle**, `/api/visors-summary` works unchanged, and the bundle contains `SkywireHttpBackend` (dormant in native mode).
- **Native mode is a 2-line delegate to `HttpXhrBackend`** — but a browser spot-check of the native HV UI (`http://<hv>:8000`) is worth doing before merge since full client-side validation needs a browser.

Includes the rebuilt `pkg/visor/static` + `dist` bundle (repo convention).


0pcom:
Stacked two more commits on this branch (per the no-divergent-branches workflow — these unblock the same CI):
- `fix(lint)`: clears pre-existing develop lint debt (gofmt/misspell/gosec/errcheck) that was red-flagging every PR's merge-tree check. None of it from this branch's own files.
- `feat(cli): --transport-port`: unified master transport port flag for `config gen` (one port carries all transport types; per-type ports become break-out overrides).

Scope grew beyond 'SkywireHttpBackend step 1'; merging as a bundle to get everything deploying.

PR #3336, opened 2026-06-28
feat(wasmhv): clearnet browsing in the iframe via skysocks (IP-anonymous)

0pcom:
## What

The in-tab skysocks client (`skywireVisor.fetchClearnet` — route → yamux → SOCKS5 → TLS-in-tab) is **validated working** (fetched `https://api.ipify.org` through a GB exit → `status=200 ip=136.244.78.62`, the exit's IP). It just wasn't reachable from the iframe browser, which only fetched dmsg/skynet sites and left clearnet URLs untouched.

This wires it in: the browse window's address bar now accepts an `https://` URL and routes it through a skysocks exit.

- `browseToClearnet(url)` → `fetchClearnet(exit, …)` → `renderClearnet`.
- `renderClearnet` mirrors the dmsg `renderSite`: same-origin stylesheets + images are re-fetched **through the same exit** and inlined as `data:` URIs, so the sandboxed iframe never touches clearnet directly (**no IP leak**). Scripts are stripped — a static, read-mostly render.
- Exit server PK is localStorage-persisted (prompted once; key `skywire-clearnet-exit`).
- A bare PK / `pk:port` still browses dmsg as before.

## Scope / limits

Static render: cross-origin sub-resources and JS-driven sites won't fully render (the sandbox can't reach clearnet, by design). Good for articles/text/simple sites; richer rendering (full sub-resource proxying) is a follow-up.

## Test

- `node --check`, `go build ./pkg/wasmhv/`, `go test ./pkg/wasmhv/` — clean
- `fetchClearnet` validated live IP-anonymous; iframe wiring served via `hv serve` (browse.js is embedded, not in the wasm blob)

### DIFF
--- a/pkg/wasmhv/browse.js
+++ b/pkg/wasmhv/browse.js
@@ -92,6 +92,9 @@
   function createBrowser(opts) {
     var frame = opts.frame;
     var fetchDmsg = opts.fetchDmsg || function () { return globalThis.skywireVisor.fetchDmsg.apply(null, arguments); };
+    // fetchClearnet(exitPK, method, url, body) → {status, body, headers}: a CLEARNET
+    // fetch tunneled through a skysocks exit over a skywire route (IP-anonymous).
+    var fetchClearnet = opts.fetchClearnet || function () { return globalThis.skywireVisor.fetchClearnet.apply(null, arguments); };
     var log = opts.log || function () {};
     var currentSitePK = "";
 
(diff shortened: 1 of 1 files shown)


0pcom:
Superseded by #3338 (merged). The clearnet-via-skysocks browsing this PR introduced is already on develop, in an improved form: #3337 added back/forward/reload/cancel, and #3338 replaced the prompt-for-exit model with the block-by-default upstream-proxy gating (self-PK = direct, other-PK = skysocks), a CSP catch-all against IP leaks, and browser-style error pages. This PR now only conflicts (it would revert the gating), and its CI failure is an unrelated Nix flake-download 429. Closing as done-via-#3338; the native HV-UI port is #3340.

PR #3415, opened 2026-07-07
feat(wasm-hv): file sharing for the browser-tab skychat (parity with native)

0pcom:
Brings the file sharing from #3414 to the **wasm visor's in-process skychat**, so the browser tab reaches native parity. It reuses the **same `pkg/skychat/xfer` primitive** over the **same `app.Client`** the browser chat already uses, on **`skyenv.SkychatFilePort`** — so a browser tab and a native visor exchange files **wire-compatibly** (offer → accept → chunked body → sha256 trailer → receipt, all over the encrypted transport, never the clearnet).

### `filexfer_js.go`
`xfer.Manager` on the browser skychat's `app.Client`, listening on the file port over dmsg + skynet; dmsg-first `DialFunc`.

**Browser-specific bit** — a tab has no filesystem:
- **Received** file buffers in memory (keyed by transfer id) and is handed to the page as base64 via `skychatFile(id)`.
- **Sent** file's bytes come from the page (a `File` → base64) into `skychatSendFile()`.
- Transfers surface as `chatMsg` entries (`is_file=true`) in the same buffer as text, so the UI renders them inline (Telegram-style).
- **Auto-accept**: the tab is the user's own trust boundary; a stranger can't route to a tab without the user's PK.

### JS bridge (`main.go`)
`skywireVisor.skychatSendFile(peerPk, name, base64[, net])` and `skywireVisor.skychatFile(id)`.

### UI (`browse.js` chat window)
📎 **attach** button (reads the chosen `File` → base64 → send); file messages render with a **download** button that fetches the received bytes and saves them as a Blob.

---
`GOOS=js GOARCH=wasm` build + vet clean · `node --check browse.js` clean · native build unaffected (files are `js && wasm`-tagged).

**Not yet exercised live:** an actual browser-tab ↔ native transfer (needs the built wasm page + two peers) — the same two-machine validation as #3414.

> **Note:** the served standalone wasm-visor embeds a pre-built wasm blob — run `make embed-wasm-visor` + rebuild to ship this to the served page.

Follow-up: group file sharing in the wasm tab (the native side fans out to CXO-group members; the browser group path can do the same once it tracks member PKs).


0pcom:
✅ **Live-validated end-to-end, both directions, over real dmsg** (native visor `0323272a…` with #3414 ↔ wasm browser tab `02325abde…` with this PR):

- **native → wasm:** `cli skychat send-file --to <wasmPK> skytest.txt` (90 B) → wasm tab surfaces `📎 skytest.txt (90 bytes)`, `file_ok:true`; `skychatFile(id)` returns the bytes — **sha256 matched the original**.
- **wasm → native:** `skychatSend()` to establish the chat, then `skychatSendFile()` → file lands in the native visor's `skychat-downloads/` — **byte-identical**.

Both transfers rode the encrypted transport on `skyenv.SkychatFilePort`; the sha256 trailer verified on each. Confirms wire-compat between the wasm `xfer` path and the native one.

PR #3438, opened 2026-07-10
feat(visor): fibercoin node discovery over dmsg (ServiceTypeCoin phase 1b)

0pcom:
Visor-side wiring for coin-node service discovery (follows #3436 which added the generic `coin` service type + `CoinInfo`).

For each configured `CoinNode`, the visor:
1. reverse-proxies the node's local HTTP API over dmsg on `DmsgPort` (reusing the forwarded-ports infra), so a remote thin-client wallet — including the browser wasm visor over its HTTP-over-dmsg shim — can reach the node's API through the mesh; and
2. on the SD heartbeat interval, probes `<LocalAddr>/api/v1/health`, builds a **detected** `CoinInfo` (`blockchain_pubkey`/`fiber`/`version`/`commit`/`head_seq` — never operator-claimed identity), and registers a `type=coin` SD entry **only while the node answers** — deregistering when it goes away, so the advertisement is always truthful (no dead entries).

The coin node runs **independently** of the visor (systemd/terminal); the visor never starts or stops it. Desktop hosts are unaffected unless `coin_nodes` is configured — this exists to make `skycoin web` (and any fibercoin thin client) work over dmsg, chiefly from the wasm visor as a substitute for the discontinued mobile wallets.

**Changes**
- `visorconfig`: `CoinNodeConfig{LocalAddr,DmsgPort,Whitelist}` + `V1.CoinNodes`, mirrored in `config_compat.go` (v1JSON + copyback).
- `servicedisc`: `HTTPClient.SetCoinInfo` to attach the detected `CoinInfo` to the registered entry.
- `pkg/visor/init_coinnode.go`: the forward + health-gated registrar loop.
- Health-JSON → `CoinInfo` parser unit-tested (skycoin mainnet sample, fiber fallback, malformed).

Depends on the `blockchain_pubkey` field in skycoin `/health` (skycoin #2912, merged); the registrar reads it from the node's live JSON at runtime so it works regardless of the visor's vendored skycoin version.

Phase 2 (wasm serves skycoin-web + routes its API over dmsg via the shim) follows.

### DIFF
--- a/cmd/skywire-cli/commands/config/envfiles.go
+++ b/cmd/skywire-cli/commands/config/envfiles.go
@@ -229,6 +229,20 @@ const envfileLinux = `#
 #	the default skycoin-web entry already has).
 #SKYCOINWEBUSER='youruser'
 
+#--	Advertise already-running fibercoin nodes over dmsg + service
+#	discovery (type=coin), so a thin-client wallet — including the
+#	browser wasm visor — can discover and reach a node for the right
+#	coin over the mesh, with no local full node and no launch flags.
+#	Bash array. Each entry is "<local_addr>[@<dmsg_port>]" where
+#	local_addr is the node's HTTP API on this host; dmsg_port defaults
+#	to local_addr's port. The node runs INDEPENDENTLY of the visor
+#	(it need not be a skywire SKYCOIND app) and is health-gated: the
+#	visor advertises it only while its /api/v1/health answers.
+#		COIN_NODES=('127.0.0.1:6420')
+#	Multi-coin (skycoin + a fibercoin on 6430):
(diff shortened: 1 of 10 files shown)


0pcom:
Added `config gen` support so this survives autoconfig/reinstall: a hidden `--coin-nodes` flag + a `COIN_NODES[@]` array in the `/etc/skywire.conf` template (envfiles.go), mirroring the `SKYCOIND*` pattern. Each entry is `<local_addr>[@<dmsg_port>]`; dmsg port defaults to the local addr's port. e.g.:
```
COIN_NODES=('127.0.0.1:6420')
```
Verified end-to-end: flag path, `${COIN_NODES[@]}` env path, and a real `SKYENV` conf file all produce the right `coin_nodes` in the generated config.

PR #3453, opened 2026-07-12
feat(hv-ui): HV-served skycoin wallet on the native visor (/wallet/ over dmsg)

0pcom:
Makes the **native** hypervisor serve the wallet exactly like the **wasm** visor: the embedded `/wallet/` static bundle + the node API proxied over the visor's dmsg client — **no skycoin-web process, no listening port.** The native equivalent of the wasm `/wallet/`, so `☰ → wallet` works on both. This is the "HV-served (serverless)" default mode from `docs/design/gui-app-serving-modes.md` (#3452).

- `pkg/visor/hypervisor_handlers_wallet.go` — serve the embedded wallet static (index `<base href>` → `/wallet/`) + `walletNodeProxy` forwarding `/wallet/api/v1|v2` to the skycoin node over `hv.visor.DmsgHTTP` (the server-side twin of the wasm `fetchDmsg` shim).
- `hypervisor.go` — mount `/wallet/*` before the UI catch-all.
- `browse.js` — ungate the `☰` wallet item (both visors serve `/wallet/` now).

### Status
Builds + `golangci-lint ./pkg/visor/` clean; browse.js syntax OK. **Draft because it needs live verification on a native visor** with a dmsg-reachable skycoin node — the dmsg-proxy path can't be exercised headless. Left as draft until confirmed on a real `:8000` HV.

### Follow-ups (tracked, not in this PR)
- config-sourced node address (kill the hard-coded default), and **multi-coin** `/coin/N` → Nth node routing;
- the wasm shim gaining the same `/coin/N` routing so multi-coin works there too.

Stacks on #3451 (About menu) since both touch `browse.js`; diff here is just the wallet ungate.


0pcom:
**Live-verified on a native visor** (`skywire visor -q http` on :8000):
- `GET /wallet/` → 200, `<base href="/wallet/">` rewritten
- `GET /wallet/api/v1/health` → **200 JSON** — node API proxied over dmsg to the deployment node
- Browser: the full skycoin Create-Wallet UI renders at `/wallet/`, client-side seed generation works (skycoin-lite.wasm), identical to the wasm visor.

Undrafting. The single-node / config-node-addr / multi-coin `/coin/N` follow-ups still stand.

PR #3480, opened 2026-07-14
feat(hv-ui): show incoming vs outgoing transports in the node list

0pcom:
Follow-up to #3479 (backend `TransportSummary.initiator` field + CLI in/out view) — surfaces the same direction in the hypervisor UI's **Transports** column.

## Changes
- `Transport` datatype gains `initiator?: boolean`; `node.service.ts` maps it from `overview.transports` in both the node-list (`getNodes`) and single-node (`getNode`) paths.
- `getTransportCounts` now returns per-type `incoming`/`outgoing` (outgoing = we dialed, `initiator=true`; incoming = we accepted the dial), plus a `getTransportTotals` helper for the Total line.
- The Transports column (desktop + mobile) renders per-type and total counts with a green ↓ (incoming) and blue ↑ (outgoing) indicator. Legacy `?`-placeholder-only nodes (sub-HVs predating TransportSummaries, #2789) show just the total — no misleading 0/0 split. Colors are green-800 / blue-800 for legibility on the light row background.

## Bundle
Re-embedded the Angular bundle (`dist/` + `pkg/visor/static/`) with `make build-ui` so the change is actually served. `embed-wallet`'s separately-embedded `wallet/` subtree was preserved (build-ui's `rm -rf` wipes it — restored). Built bundle verified to contain the new symbols (`getTransportTotals`, `tp-in`, `tp-out`).


0pcom:
Fixed the indicator colors: the manager UI is **dark-themed** (`#1f2533` rows, white text), so the initial green-800/blue-800 would have been near-invisible. Switched to `#4caf50` (green — the same status-green already used elsewhere in this stylesheet on the dark background) for incoming and `#42a5f5` (light blue) for outgoing, both ~6:1 contrast on the dark rows. Bundle rebuilt + re-embedded.

PR #3553, opened 2026-07-22
feat(vpn): mesh gateway for .dmsg/.skynet — vpn-router + vpn-client

0pcom:
## What

Turns the **vpn-router into a full mesh gateway**: any LAN device behind it can reach dmsg/skynet services **by name** — `curl http://<pk>.dmsg`, open `http://<pk>.skynet` in a browser — with **no per-device configuration and no SOCKS proxy**.

## Why it needs more than NAT

A `.dmsg` / `.skynet` name isn't a normal hostname. It encodes a **public key**, and the destination is a **routing port reachable only over the mesh** — often there's no externally-listening TCP port and no IP to route to at all. So the gateway can't just forward/NAT it. It works in two halves:

1. **DNS** — the router's embedded resolver hands `*.dmsg` / `*.skynet` queries to `meshgw`, which parses the name to a PK (reusing dmsgweb's `skynetweb.ParseResolverHost` grammar: raw hex, DNSLabel, or alias), leases a **synthetic IP** from a private pool (default `100.64.0.0/16`), records `synthetic-IP → (scheme, pk)`, and answers with that IP.
2. **Transparent proxy** — a `nat/PREROUTING` `REDIRECT` sends TCP aimed at the synthetic pool to a local proxy. It reads `SO_ORIGINAL_DST`: the synthetic IP maps back to the PK, and the **original destination port is the mesh routing port**. It dials `<pk>:<port>` over the mesh (injected `MeshDial`) and splices the two streams.

## Bypasses the tunnel by design

The mesh path deliberately does **not** go through the VPN tunnel: the `REDIRECT` fires in `PREROUTING` before any routing decision, so pool traffic never hits the LAN→tun policy route; and the proxy's own mesh dials originate locally (no `iif`) on the main table. Mesh services are reached over the visor's transports directly, not via the exit server.

## Decoupled so the vpn-client can reuse it

`pkg/vpnrouter/meshgw` is decoupled from the router via an injected `MeshDial`. The **same component can later back the vpn-client** (localhost DNS + `OUTPUT`-chain REDIRECT instead of LAN DNS + `PREROUTING`) — it's really a visor-level capability with the vpn-router and vpn-client as two exposure points.

## Changes

- **`pkg/vpnrouter/meshgw`** (new): `Gateway` (resolve / lookup / `ServeTransparent`), synthetic-IP pool, `.dmsg`/`.skynet` DNS zone handlers, `SO_ORIGINAL_DST` reader (+ non-linux stub), unit tests.
- **`pkg/vpn`**: `RouterConfig.MeshDial` / `MeshGatewayCIDR`; `router_linux` builds the gateway, starts the proxy + REDIRECT, installs the DNS zones via `StartLAN`, and tears it all down on exit.
- **`cmd/apps/vpn-router`**: `--mesh-gateway` / `--mesh-gateway-cidr`; `MeshDial` implemented over the app's `app.Client` (visor-launched only — standalone mode errors clearly, since it has no app-server to dial the mesh).
- **config-gen + autoconfig**: `VPNROUTERMESHGW` / `VPNROUTERMESHGWCIDR` knobs, documented in the `skywire.conf` template.

## Status

Builds, unit-tested (`go test ./pkg/vpnrouter/meshgw/...`), `golangci-lint` clean. **End-to-end on-hardware validation (LAN client → `.dmsg` service) is still pending** — the SBC test boards are in flux and this is the plaintext/by-PK foundation.

## Follow-ups (not in this PR)

- TLS-MITM for HTTPS to `.dmsg` hosts (à la dmsgweb + skynetca `LeafMinter`).
- alias → PK resolution via service discovery.
- wire the same `meshgw` into the **vpn-client** for single-machine mesh-name resolution.

### DIFF
--- a/cmd/apps/vpn-client/commands/vpn-client.go
+++ b/cmd/apps/vpn-client/commands/vpn-client.go
@@ -9,6 +9,7 @@ import (
 	"net"
 	"os"
 	"os/signal"
+	"path/filepath"
 	"runtime"
 	"strconv"
 	"syscall"
@@ -39,7 +40,17 @@ var (
 	appPort     uint16
 	muxRoutes   int
 	minHops     int
+	meshGateway bool
+	meshGWCIDR  string
+	meshTLS     bool
+	meshCACert  string
+	meshCAKey   string
+	meshAliases []string
 )
+
+// defaultMeshCADir is where the client mesh gateway persists its self-generated
+// CA so the host only needs to trust it once (survives restarts).
+var defaultMeshCADir = filepath.Join(skyenv.PackageConfig().LocalPath, "mesh-gateway-ca")
 
 func init() {
 	launcher.RegisterApp("vpn-client", RunVPNClient)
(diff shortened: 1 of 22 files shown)


0pcom:
Pushed two of the listed follow-ups:

- **HTTPS (TLS-MITM)** — `meshgw.EnableTLSMITM` terminates the client's TLS with a per-SNI minted leaf and bridges plaintext to the plain-HTTP mesh upstream (dmsg/skynet counterpart of dmsgweb's browser MITM). The vpn-router self-generates + persists a CA under `<local>/mesh-gateway-ca/`, logging its fingerprint for operators to trust on LAN clients. Flags `--mesh-gateway-tls`/`--mesh-gateway-ca[-key]`, knob `VPNROUTERMESHTLS`. Covered by a new end-to-end `spliceTLS` test.
- **Friendly aliases** — `RouterConfig.MeshAliases` threaded into `meshgw.New`; app takes repeatable `--mesh-alias name=<pk>` so clients can reach `http://<alias>.dmsg`.

Still to come: wiring the same `meshgw` into the **vpn-client** (localhost DNS + OUTPUT-chain REDIRECT) for single-machine mesh-name resolution — opt-in, since on a user's own machine hijacking DNS + iptables is more intrusive than on a dedicated router.

0pcom:
Third follow-up landed: **vpn-client integration**.

The same `meshgw` now drives a single-machine mesh gateway on the vpn-client — the reuse the injected `MeshDial` was designed for. Where the router uses LAN DNS + `PREROUTING` REDIRECT, the client uses a **loopback resolver** (router7's upstream-forwarding DNS + meshgw's zones) + **`OUTPUT`-chain REDIRECT**:
- DNS → loopback resolver (its own upstreams `RETURN`-exempted first to avoid a redirect loop);
- synthetic-pool TCP → the transparent proxy;
- the mesh dial (`app.Client`) bypasses the tunnel — it rides the visor's transports, already exempted from the tun via the client's direct routes — so mesh names resolve while everything else egresses the VPN.

Strictly **opt-in** (`--mesh-gateway`, Linux-only, non-fatal if setup fails) — on a user's own machine, hijacking DNS + `OUTPUT` nat is more intrusive than on a dedicated router. HTTPS + aliases work here too. The `MeshDialer`/`ParseMeshAliases`/`LoadOrCreateMeshCA` wiring is factored into `pkg/vpn` and shared by both apps.

All three follow-ups from the original description are now in this PR. Everything builds, `meshgw` tests pass (incl. the TLS-MITM e2e), `golangci-lint` 0 issues. **On-hardware validation is still pending the dmsg-server migration** (see below).

PR #3556, opened 2026-07-23
feat(vpn): Go-native control plane — netlink (default) + nftables firewall (tag-gated)

0pcom:
First step of the **gokrazy appliance** effort (see the tracking issue): make the VPN router/client's control plane Go-native so it needs no external `ip`/`sysctl` binaries.

## What
Replace the `ip` and `sysctl` shell-outs (via `osutil.RunElevated`) with a netlink + `/proc` backend, `pkg/vpn/netctl`. Talking netlink directly is exactly what `ip` does under the hood — **identical kernel behavior** — but with:
- no iproute2 dependency, no `PATH` assumptions;
- no per-command `pkexec`/`sudo` (kills the "pkexec flood" on non-root hosts);
- a router that can run on a userland-less appliance (gokrazy ships no `ip`/`sysctl`). The data plane is already Go-native (tun via `songgao/water`, DHCP+DNS via the vendored gokrazy router7); this converts the control plane.

## Coverage
`netctl` maps 1:1 to every `ip`/`sysctl` site: address add/flush/del, link up + MTU, route add (`EEXIST`-aware) / replace / del via gateway, default-dev route + table flush, `iif` policy rules, IPv4/IPv6 forwarding via `/proc`. The process still needs `CAP_NET_ADMIN` — which it already has wherever it does network setup (the existing `setSysPrivileges`/`Setuid(0)` window is preserved). The old `ip r add` stderr handling (`File exists` → ok, `Operation not permitted` → denied) is reproduced via `EEXIST`/`EPERM` errno checks.

## Not in this PR
`iptables` (masquerade / FORWARD / REDIRECT / TCPMSS clamp) still shells out — the nftables conversion is a separate, carefully-validated follow-up (Phase 0b), because the MSS-clamp/REDIRECT expressions are delicate and warrant their own review.

## Validation
Against **real kernel netlink** in a rootless user+net namespace (no root, no host impact):
```
unshare -rn go test -tags netctlint -run TestNetctlNetns ./pkg/vpn/netctl/
```
Exercises every op and asserts via independent netlink queries (incl. `RouteAdd` correctly failing `EEXIST`). Builds clean, `golangci-lint` 0 issues.

⚠️ **Fleet note:** this touches production VPN router/client networking and I could only validate in netns, not on fleet hardware — worth a board smoke-test before merge.


0pcom:
Added the **Phase 0b nftables firewall backend** (`pkg/vpn/netctl/firewall_nft_linux.go`), tag-gated behind `nftfw` so the default build is unchanged (fleet keeps iptables).

Covers masquerade, FORWARD accept (+ stateful return path), the mesh-gateway REDIRECT (prerouting + output hooks + DNS return-exempt), and the TCPMSS clamp — the last via nftables' native `rt tcpmss` ("clamp to pmtu"). All rules live in a dedicated `skywire` table; `Teardown` drops the whole table.

**Validated** against the real kernel in a rootless netns (`unshare -rn go test -tags 'nftfw netctlint' -run TestNFTFirewall ./pkg/vpn/netctl/`): every rule installs (kernel rejects malformed bytecode → a clean install proves the expressions are valid nftables), per-chain rule counts asserted, and the OUTPUT REDIRECT proven **end-to-end** (a local connection to the synthetic pool lands on the proxy).

It's not yet called from the router/client firewall paths — that wiring (behind the `nftfw` tag, via a small abstraction so the default stays iptables) lands with the gokrazy appliance image in Phase 1, where the whole `nftfw` build is exercised together. Tracking: #3557.

0pcom:
**Validated on real hardware** ✅ — an Orange Pi Prime (aarch64, kernel `5.15.48-sunxi64`), the exact SBC class the fleet runs. Cross-compiled the netctl test binary for arm64, pushed it to the board over the mesh (`dmsg scp`), and ran it in an isolated network namespace (leaving the board's real networking untouched — confirmed no residue, skywire stayed up):

```
=== running netctl + nftables tests on aarch64 kernel 5.15.48-sunxi64 in isolated netns ===
--- PASS: TestNFTFirewall     (nftables masq/forward/mss-clamp/redirect + e2e REDIRECT)
--- PASS: TestNetctlRealTUN    (netctl on a real POINTOPOINT tun)
--- PASS: TestNetctlNetns      (every netctl op)
PASS
```

So both the netlink control plane (0a) and the nftables firewall backend (0b) work on the real target kernel, not just my dev box. That closes the on-hardware gap that was the merge caveat.

Issue #3559, opened 2026-07-23
Improvement on Skychat


mrpalide:
Some simple and basic features we need on Skychat:   
  - [x] render media in chat
  - [x] reply, edit, delete (on dm and group)
  - [x] message status (sending/sent/read)
  - [x] notifications
    - [x] notification hub
  - [ ] integrate/sync skychat app with embbeded skychat on wasm visor ui
  - [ ] basic equivalency to telegram

0pcom:
All concrete checklist items are done: media rendering (#3592), reply/edit/delete, message status, notifications + hub (#3666), and skychat<->wasm-UI integration (#3641/#3720 — the wasm desktop hosts the same Angular skychat component). The remaining "basic equivalency to telegram" is open-ended; re-filed as a scoped feature enumeration in #3782. Closing.

PR #3631, opened 2026-07-30
[WIP] Improve group feature of Skychat, both on features and infrastructures 

mrpalide:

Fixes: Part of #3559 

 Changes:	
- [x] add group admission control — join requests, approval queue, kick/ban/mute/read-only
- [x] reject stale roster/admin/mod gossip — replaying an old signed leaf silently undid evictions
- [x] let an invite name several admins — join requests pointed at OwnerPK alone, so an offline founder meant nobody could join
- [x] promote/demote from the members panel — the RPC existed, nothing called it, so a group could never get a second admin
- [x] rotate the group key on eviction — a ban took the seat but left the key, so a copy of the feed kept opening
- [x] seal group keys at rest — groups.db handed over every private group's AES key to anyone who could read the file
- [x] serve group history from any online member, not just admins — admin-toggleable per group, choosable at create time
- [x] price and rate-limit joining — free public keys meant a flood could outrun any per-PK gate, including ban
- [x] bound compromise in time — DH-ratcheted epoch keys for DMs, ephemeral wraps and age/volume rotation for groups
- [x] fix CI — realistic per-package test timeout, and move DmsgVisorRPCPort off a port the visor already binds
- [x] encrypt governance gossip — roster/admin/mod leaves are signed but plaintext, so who you banned and muted is readable even in an encrypted group
- [x] encrypt group file bytes — messages are sealed, attachments aren't; files ride the backfill path with no group-key encryption
- [x] "ask again" after a declined request — denial is terminal by design, so the UI needs an explicit retry button
- ~[ ] retention / disappearing messages — no TTL, no expiry; data accumulates forever and delete-for-everyone is cooperative~
- ~[ ] metadata reduction — SenderPK and exact timestamps sit in the clear beside the ciphertext, and the federated topology exposes the subscription graph~

How to test this PR:

Browser, against three real visors:

    make e2e-build && make e2e-run && make e2e-skychat

then, across the three UIs it prints:

1. Create a private group on visor A. Invite link alone should NOT let B in —
   B lands in A's approval queue; approve it and B gets the key with the
   approval, not from the link.
2. Create a public group. The link alone should be enough, and the first join
   should visibly cost a moment (proof-of-work) rather than being instant.
3. Ban B from the private group. B loses the seat AND stops being able to read
   new messages; A's admin panel shows the key epoch advance. Older messages
   stay readable for everyone still in.
4. Promote C to admin, take A offline, then have a fourth visor join using the
   same invite link — C should be able to admit it.
5. Open a DM between two paired visors and send one message each way. The chat
   header should move from "encrypted · keys not exchanged yet" to
   "forward secret · key <id>" once both sides have spoken.


0pcom:
Landed on `develop` — thank you for all of this work. 🙏

Your PR conflicted only because of the `cmd/apps/skychat/group` → `pkg/skychat/group` relocation, so I replicated your branch over the move and merged it via #3644, **with all your commits preserved under your authorship**. Everything in the checked list shipped: admission control, stale-gossip rejection, multi-admin invites, promote/demote, key rotation on eviction, sealed keys at rest, member-served history, PoW + join rate-limiting, epoch-key ratcheting, encrypted governance gossip + attachments, and the CI test-timeout fix (I also extended that to the darwin lane, which was still on the inline 5m).

I finished the one remaining checked-but-open item — **"ask again" after a decline** — end to end (core + RPC + native app HTTP + wasm hook + CLI + HV UI badge/button), so a declined requester can deliberately re-queue while a passive re-ask stays terminal (no poll-flood).

The two still-unchecked items — **retention/TTL (disappearing messages)** and **metadata reduction** — I deferred rather than block the merge; they fold into a message-persistence + anti-abuse design pass we're planning (bounded store + disk-fill protection). Happy to collaborate on those. This PR can be closed as landed.

PR #3659, opened 2026-08-01
feat(visor): Suspend/Resume RPC — quiesce the visor without stopping the process

0pcom:
Adds a **Suspend/Resume** lifecycle so an operator or the systray can pause/unpause a node **without privilege to control the service-manager unit** — the privesc-free alternative to `systemctl stop/start`.

### Behavior
- **Suspend** tears down every network subsystem (transports, routing, apps, dmsg, autoconnect, AR/TPD registration, reward heartbeat, and the dmsg/transport-bound RPC surfaces) while keeping the **local CLI RPC listener** alive and the process running. The node earns nothing, routes nothing, and drops off dmsg discovery, but stays controllable on `CLIAddr`.
- **Resume** re-runs the module graph to bring it fully back online.
- **Local-only by design:** a suspended visor is unreachable over dmsg, so the hypervisor-proxied API returns an explanatory error; Suspend/Resume must be called on the target's own host RPC.

### Mechanism
- `NewVisor` wraps the module ctx in a cancelable child (`networkCtx`/`networkCancel`) stored on the visor; Suspend cancels it so ctx-honoring goroutines unwind before their listeners/clients close.
- `initCLI` is split: the **local** RPC surface (cli.listener + cmux + local gRPC + net/rpc accept) is created once and guarded by `cliLocalUp` so a Resume re-run doesn't re-bind `CLIAddr`; the dmsg/transport-bound surfaces re-create every init.
- Suspend walks the closeStack in reverse (same LIFO order as `Close`), runs every closer except the kept local-RPC entries, retains only those, drops references to the torn-down subsystems, and resets the one-shot readiness channels (stun/dmsgTracker also reset their `sync.Once`, matching the `api_network` reconnect pattern).

### Surfaces
`API.Suspend/Resume/IsSuspended`, RPC wrappers + client + mock, and `skywire cli visor suspend|resume|suspended`.

### Testing
- **Unit-tested (`api_suspend_test.go`):** Suspend runs non-kept closers in LIFO order, retains only the kept local-RPC entries, resets the one-shot channels (verified open afterward), and is idempotent.
- `pkg/visor` suite + `golangci-lint` clean.

> ⚠️ **Not for auto-merge.** Resume's full module-graph re-init needs live validation on a running node (suspend → confirm quiescent + local RPC still answers → resume → confirm transports/routing/apps/dmsg all return). I could not run this locally without standing up a second native visor (against our operational rule), so please validate on a spare/non-critical node before merging. Design + teardown logic are covered by the unit test; the re-init path is the part to exercise.

### DIFF
--- a/cmd/skywire-cli/commands/visor/process.go
+++ b/cmd/skywire-cli/commands/visor/process.go
@@ -122,16 +122,40 @@ the service manager unit. Resume with 'skywire cli visor resume'.`,
 var resumeCmd = &cobra.Command{
 	Use:   "resume",
 	Short: "Resume a suspended visor",
-	Long:  "Bring a suspended visor fully back online by re-running its module graph.",
+	Long: `Bring a suspended visor fully back online by re-running its module graph.
+
+Resume re-inits every network subsystem, which takes several seconds and
+severs the in-flight RPC connection mid-re-init (the same way 'reload'
+does — the local RPC LISTENER stays up, only this one call's connection
+drops). So it's fired in the background and the result is confirmed by
+polling the suspend state on fresh connections rather than waiting on the
+severed call.`,
 	Run: func(cmd *cobra.Command, _ []string) {
 		rpcClient, err := clirpc.Client(cmd.Flags())
 		if err != nil {
 			os.Exit(1)
 		}
-		if err := rpcClient.Resume(); err != nil {
-			internal.PrintFatalError(cmd.Flags(), fmt.Errorf("error resuming visor: %v", err))
+		// Fire Resume; its own connection is expected to EOF during the
+		// re-init, so don't treat that as fatal — the state is polled below.
+		go func() { _ = rpcClient.Resume() }() //nolint:errcheck
+
+		// Poll IsSuspended on fresh connections until it clears (resume
+		// finished flipping state) or we give up waiting.
+		deadline := time.Now().Add(45 * time.Second)
+		for time.Now().Before(deadline) {
+			<-time.After(2 * time.Second)
+			rc, e := clirpc.Client(cmd.Flags())
+			if e != nil {
+				continue
+			}
+			suspended, e := rc.IsSuspended()
+			if e == nil && !suspended {
+				internal.PrintOutput(cmd.Flags(), "Visor resumed", fmt.Sprintln("Visor resumed"))
+				return
+			}
 		}
-		internal.PrintOutput(cmd.Flags(), "Visor resumed", fmt.Sprintln("Visor resumed"))
+		internal.PrintOutput(cmd.Flags(), "Visor resume initiated",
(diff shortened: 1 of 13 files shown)


0pcom:
**Live validation on a local visor (rebuild-restart loop) — found a blocking defect; do not merge yet.**

Two rounds against a real visor:

1. **Happy path worked:** `suspend` tore down all networking (transports 2→0) while keeping the local RPC listener alive (`suspended`=true, `visor pk` still answered); `resume` re-ran the module graph and the visor fully recovered (18 transports reformed, all apps running, `suspended`=false).

2. **But `resume`'s RPC call returned "unexpected EOF"** — the multi-second re-init severs its own in-flight connection (same self-severing property `reload` has; the listener survives). Fixed CLI-side by firing resume in the background and confirming via polling `IsSuspended` (commit pushed — mirrors the `reload` command).

3. **Blocking defect:** on a second run, `suspend` **crashed the visor** (process exited at 18:10:21, the rebuild-loop restarted it fresh — not a resume). Root cause is almost certainly the **nil'd-field race**: `Suspend()` sets `v.tpM`/`v.router`/`v.dmsgC`/… to nil, but a surviving goroutine that isn't bound to `networkCtx` dereferences one → nil-pointer panic → process exit. The first run got lucky on timing.

**Fix needed before merge:** either don't nil the torn-down subsystem fields (leave them pointing at the closed-but-non-nil objects), or guarantee every goroutine referencing them is stopped before Suspend nils them (the design agent flagged this exact dangling-reference risk for the full-teardown model). Keeping open until that's resolved + re-validated.

0pcom:
**Update after deeper live testing — suspend now crash-safe; resume needs a design fix.**

- **Suspend: FIXED.** Two fixes landed (commits pushed): `runCloser` now recovers a panicking closer (Suspend runs closers on a LIVE process, unlike Close at exit), and Suspend no longer nils the torn-down fields (leaving them closed-but-non-nil avoids a nil-deref race with in-flight goroutines). Verified: the visor's PID stays stable through `suspend` (no crash), local RPC stays up, `suspended`=true.

- **Resume: still crashes — design-level cause identified.** `Resume()` re-runs the **entire** module graph via `InitConcurrent`, but Suspend only tore down *part* of it (the network subsystems). So resume **double-initializes** the modules that survived suspend → a double-spawn / double-close panic in a **goroutine spawned by an init** (not the init hook itself, so the new per-hook recover in `withInitCtx` can't catch it) → process exit (the rebuild-loop then restarts it fresh, masking the crash).

**Fix path (not a quick patch):** Resume must re-run **only the modules Suspend tore down** (rebuild a sub-graph of the torn-down set + their torn-down deps), or every module init must be made fully idempotent. Capturing the panic stack (visor stderr, currently only in the terminal) will confirm the first offending module. Keeping this PR open; the CLI resume-EOF handling + the crash-hardening are already in.

PR #3707, opened 2026-08-04
feat(hv-ui): mount the shared pkg/tpviz visualizer in the tab (retire the reimplementation)

0pcom:
Pilot for the GUI embedding standardization RFC (#3706,
`docs/design/gui-embedding-standardization.md`).

The "Network Visualizer" tab was a separate 221-line Angular `vis-network`
reimplementation of what `pkg/tpviz` already does in full. This makes the tab host
the **one** tpviz implementation by **mounting its bundle into the tab's `<div>`**
— no iframe, no `/tp-viz/` break-out, staying in the SPA's own JS/DOM context.

### tpviz side (`pkg/tpviz/ui`)
- `src/mount.ts`: `SkywireTpviz.mount(root)` injects the app's own DOM + a **scoped**
  copy of its CSS into any container. A small CSS scoper confines the tpviz
  reset/base rules under `.tpviz-root` so they don't leak onto the host SPA, plus a
  width/height override so `#container` fills the mount container instead of
  `100vw/100vh`. Returns an `unmount()` handle.
- `index.ts` exposes `SkywireTpviz` and only auto-wires when the standalone DOM is
  present (`#container`) — so `/tp-viz/` still works unchanged; embedding is
  host-driven.
- `build.mjs` imports `index.html` as text so the bundle carries its own template.

### Angular side
- `network-visualizer.component` loads `/tp-viz/bundle.js` once and mounts it into
  its `<div>`; unmounts on destroy. The vis-network reimplementation is removed.

### Verified live (native HV)
The full tpviz UI — Transport Discovery / DMSG Discovery / Filters & Legend
sidebar, Local Visor / Apps / Network Ping, Globe/Flat/WebGL toggle, country-grouped
graph — renders inside the tab, SPA nav intact, CSS scoped (screenshot in the
thread).

### Follow-ups
- Serve `/tp-viz/bundle.js` in `hv serve` so the **wasm** surface can mount it too
  (today it works on the native HV only).
- Give tpviz a teardown so `unmount()` also clears its auto-refresh timers.
- Apply the same `mount()` contract to skychat / wallet-config next.


0pcom:
Superseded by #3709, which includes this mount pilot commit plus the follow-up polish (WebGL↔Flat parity, SD-over-dmsg country data, tighter grouping, console-spam fix). Closing in favor of the consolidated PR.

PR #3793, opened 2026-08-07
feat(wasm-visor): dynamic clearnet render mode (draft)

0pcom:
**Draft** — stacked on #3788 (shares its commits until it merges).

Opt-in per-window `⚡ dynamic` mode for the iframe proxy browser: runs the page's own JS and bridges its runtime network (fetch / XHR / dynamic `<script>` chunks / images) through the skysocks exit via the existing `dmsgreq`→`fetchClearnet` postMessage bridge. Static mode stays the default and is unchanged. The dynamic CSP keeps `connect-src` off http(s) so anything unpatched fails closed — the render stays IP-anonymous.

Also adds a `navigator.serviceWorker` no-op stub to `storageShimSrc` (reading it throws `SecurityError` in an opaque srcdoc, killing CRA-style apps at startup) — same class as the existing storage/history shims.

### Status
Component-level validated: the fetch/XHR bridge relays real requests through the exit; the serviceWorker `SecurityError` is gone. **Full end-to-end render of complex SPAs is gated on skysocks-lite connection robustness** (each browse window currently cold-establishes its own per-window session instead of reusing one shared default) — tracked separately. Kept draft until that lands and a clean end-to-end demo is captured.


0pcom:
Superseded by #3794 (which contains the dynamic-clearnet-mode commit).

PR #3994, opened 2026-08-18
feat(routing): per-app override of the visor-global min_hops/mux_routes

0pcom:
A visor configured with `min_hops>1` / `mux_routes>1` forced **every** skynet app dial into multi-hop mux, with no way for an individual app to opt out. Two problems:

1. It breaks a **1:1 forward** — `skynet-client` dials the built-in forwarding service on `SkyForwardingServerPort` (a single-route port); a forced 4-way mux route group to it fails (0 bytes). Root-caused live: `mux-bw` to the same peer with the identical `routes=4 min-hops=2` works (15.6 Mbps), and the in-process resolver works — only the app-path forward, which blindly inherits the global mux, fails.
2. It's just **annoying**: an app couldn't ask for a single/direct route when the global default is aggressive.

The networker already honors an explicit opts value (it applies the global only when `opts.MuxRoutes==0`), but an explicit **1** never reached it:
- CLI `skynet start` only forwarded `--routes`/`--min-hops` when `> 1` → `--routes 1` was silently dropped.
- `skynet-client` only took the `DialWithOptions` path when a field was `> 1`.
- app-server ingress `dialWithMuxRoutes` treated `<= 1` as unset → global-inheriting default dial.

**Fix:** any *explicit* per-call value (including `1`) is a per-app override that reaches the router; only a zero value inherits the visor-global. So `skynet start --routes 1 --min-hops 1` forces a single/direct route out from under a global `mux_routes=4/min_hops=2`.

Verified live: `--routes 1` now lands in the app's args (previously swallowed as ≤1). `go build .` + `golangci-lint` clean. (Full byte-level forward proof pending a stable far-end sink.)

### DIFF
--- a/cmd/apps/skynet-client/commands/skynet-client.go
+++ b/cmd/apps/skynet-client/commands/skynet-client.go
@@ -225,7 +225,12 @@ func RunSkynetClient(ctx context.Context, args []string) error {
 	appCl.Log().Infof("Per-accept dial shape: %s", dialShape)
 
 	rawDial := func() (net.Conn, error) {
-		if direct || routes > 1 || minHops > 1 || fwdMinHops > 1 || revMinHops > 1 || fwdMux > 1 || revMux > 1 {
+		// Any EXPLICIT per-call value (>=1), including 1, is a per-app override
+		// of the visor-global min_hops/mux_routes — so `--routes 1` / `--min-hops 1`
+		// force a single / direct route out from under a global mux_routes/min_hops>1
+		// (which otherwise forces this 1:1 forward into multi-hop mux and breaks it).
+		// Unset (0) still inherits the global default via the plain Dial path.
+		if direct || routes >= 1 || minHops >= 1 || fwdMinHops >= 1 || revMinHops >= 1 || fwdMux >= 1 || revMux >= 1 {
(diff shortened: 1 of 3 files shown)


0pcom:
Holding — do not merge yet. The ingress change (`<=1` → `==0`) is correct for an EXPLICIT override, but `proxy start`/`vpn` pass their `--mux` **default of 1 unconditionally**, so this flips skysocks/vpn from *inheriting the visor-global mux* to their *flag-default single route* — a behavior change (observed: default skysocks came up with 1 leg instead of the global 4). Needs the unset-vs-explicit distinction threaded through the proxy/vpn CLIs (gate `--mux`/`--min-hops` on `cmd.Flags().Changed(...)`, or a sentinel) before this is safe. skynet-client half is already Changed-gated and correct.

0pcom:
Un-holding — the regression fear was DISPROVEN. Tested on clean develop: default `proxy start` (no --mux) comes up with **1 leg there too** (proxy's --mux default is 1; skysocks-client's default dial is all-zero `appCl.Dial`, which the `==0` gate does not touch). So this is non-regressing — only *explicit* non-zero per-call values now override the visor-global; the all-zero default path is unchanged. Validated: `--routes 1` now reaches the app's args (was swallowed). The skynet forward itself works (`--direct` pulled 1.2 MB / HTTP 200); a fuller mux-forward path is a separate follow-up (warm-pool ladder).

PR #4047, opened 2026-08-21
feat(routing-policy): adaptive default proactively holds warm standby + asymmetric F/R (app-agnostic)

0pcom:
## What

Makes the **`adaptive`** composite preset — the wired default (`preset:adaptive`) — actually manifest its mux / warm-standby / asymmetric behavior on real apps, not just the ad-hoc `mux-bw` rig.

Previously `decideAdaptive` seeded `Mux=1` for three hardcoded client-app names and the empty spec for everything else, so the default route group came up **single-leg** and only grew reactively under sustained load. The router-layer machinery to reserve N disjoint routes, hold warm standby (on_tick demote/promote), and shape asymmetric forward/reverse mux **already exists and is app-agnostic** (`establishMuxRoutes`, `EffectiveMuxRoutes`, `appendRouteAsymmetric`, `setLegStandby`). The only gap was the default policy. This wires it.

## Changes (policy layer only — no router core changes)

- **App-agnostic default.** `decideAdaptive` now applies to *every* overlay app that dials a route group (proxy, vpn, skynet-client, the resolving-proxy / browser chain, custom-named sessions). Only latency-sensitive chat stays a single lean route (same idiom as `rotating-bw`). No hardcoded allowlist of binary names — the old three-name switch made the policy a silent no-op for renamed sessions.
- **Asymmetric F/R by default.** `ForwardMux=1` (one lean upstream/request leg, no mux head-of-line cost) + `ReverseMux=adaptRevActive+adaptStandbyMax` (a wider bulk-download mux sized active + a warm-standby reserve).
- **Proactive warm standby.** `tickAdaptive` parks the surplus reverse legs as warm standby (down to the steady active target) once no leg is a latency outlier and the group is not saturated — holding instant-promote spares **without waiting for an idle signal**. It parks the newest surplus leg, never leg 0 (the primary/forward leg the router refuses to demote). Under load the saturated branch promotes a spare (dip-free) before adding a cold leg.

## Parity & safety

- **Native == wazero parity preserved:** `bundle.wasm` rebuilt from the shared `pkg/router/policy/preset` package via TinyGo; `parity_test` (decide + multi-tick on_tick) green.
- **Graceful degradation preserved:** no candidates / unknown transport kinds → `Chosen` nil, router picks the path exactly as before (the wasm / no-provider path).
- **`mux=1` apps and single-route dials unchanged.**

## Not included (design blocker)

Runtime *load-based F/R flip* (swap which direction is wide from `on_tick` per-leg {bps}) is **not** implemented: the `RotationAction` ABI has no forward/reverse-width field, and per-direction rules are provisioned at setup, so re-shaping the fwd/rev leg allocation at runtime needs a setup-node re-provision (a dip) or an ABI + wire extension. Reported rather than shipped as a fragile re-dial. The **static** asymmetric shape (lean forward / wide reverse) is shipped.

## Tests

`go build .`, `go vet`, `go test ./pkg/router/policy/... ./pkg/router/` all green, including the native↔wazero parity gate and a new `TestEngine_OnTick_AdaptiveHoldsWarmStandby`.


0pcom:
## Update: stability hardening (folds in live findings) + always-on standby floor

Two follow-ups from live use, both now in the branch.

### 1. Standby FLOOR (never drained by load)
Previously saturation promoted **all** standby → active, so peak load left `standby=0` and a subsequent leg drop ate a re-establish dip. Now:
- `adaptStandbyMin` (=1) is a hard floor load may never cross — under saturation the tick grows the active width with a **fresh** leg rather than promoting the last warm spare.
- An active-leg **drop** is covered by an instant healthy-spare promote **plus** a re-establish that restores the floor — zero re-dial dip.

### 2. Health gate + anti-churn (the "cuts in and out" fix)
Live, adaptive-as-default made real long-lived connections (interactive browsing, chat-over-proxy) stutter/drop, because the mux placed **bad legs** (a 2314 ms leg, dead 0-byte legs) in the active set — the no-skip reorder buffer then head-of-line-stalled — and the active set **churned** every few ticks. Fixed in the `tickAdaptive` rewrite:
- **Interactive rides ONE leg.** Steady active width is now `1` (`adaptRevActive=1`): leg 0 alone carries an idle/interactive flow (no muxing/reordering). The reverse mux widens **only under sustained bulk saturation** and shrinks back after.
- **Health gate.** A gross-latency-outlier (`> adaptLatCeilingMs` absolute, or `> adaptOutlierMult × active median`) or alive-but-no-traffic (dead 0-byte) leg is kept OUT of the active mux — evicted (hot-swap to a healthy spare, else park/drop) after a sustained window; standby promotion skips known-bad legs.
- **Anti-churn.** A signal must persist `adaptHysteresis` (3) ticks before any reshape, and `adaptReshapeCooldown` (3) ticks of steadiness follow one. Leg 0 is never disrupted. The periodic explore/probe (a churn source) was removed.

**Constants chosen:** `adaptRevActive=1`, `adaptStandbyMax=2`, `adaptStandbyMin=1`, `adaptLatCeilingMs=1500`, `adaptOutlierMult=3`, `adaptHysteresis=3`, `adaptReshapeCooldown=3`.

**How the saturated branch now adds-instead-of-drains:** growth promotes a healthy spare **only** while `standbyCount > adaptStandbyMin`; at/below the floor it emits `AddLeg` (a fresh leg) so the reserve is preserved.

New tests: gross-outlier active leg evicted (single spike below hysteresis is not — no churn); active set converges + holds under steady load (bounded reshapes + long no-op tail); forward path stays single-leg; standby floor never drained under sustained saturation and restored after a drop; leg 0 never parked/evicted. `go build .`, `go vet`, `go test ./pkg/router/policy/... ./pkg/router/`, native↔wazero parity, and the TinyGo bundle rebuild are all green. No AI attribution.

PR #4057, opened 2026-08-21
fix(router): retransmit route-group setup handshake to prevent intermittent skynet route drops

0pcom:
## Problem

Skynet route setups still intermittently drop with:

```
WARN [router]: Remote handshake not received within timeout — failing dial ... timeout=10s
```

`#4051` made these drops *self-heal* (the dial now returns an error and the app retries) but did not stop them from happening. This PR addresses the residual root cause.

## Root cause

The **route-group setup handshake** (`saveRouteGroupRules`, `pkg/router/router_serve.go`) is a single unreliable packet:

- The initiator calls `rg.sendHandshake(true)` **exactly once**, then blocks on `rg.handshakeProcessed` for the full `handshakeAwaitTimeout` (10s).
- That handshake packet is **not** carried by the mux reorder/SACK machinery — that layer is only built *after* the handshake completes (`handlePacket` → `HandshakePacket` sets up `rg.mux`). So the setup handshake has **no ARQ**: a single lost handshake packet on any hop has nothing to recover it.
- The initiator then waits out the entire 10s and hard-fails an otherwise-live dial.

Raising the deadline cannot help — the history of bumping `6s → 10s` only made each doomed attempt slower to fail. A dropped packet is gone regardless of how long you wait.

Evidence: on a public visor, repeated back-to-back 10s handshake timeouts to a **healthy** exit whose measured ICMP RTT is ~134ms. A 1-RTT setup handshake genuinely exceeding 10s over a 134ms path is packet loss, not latency. The handshake packet also bypasses the reorder buffer (`dispatchToRouteGroup` → `rg.handlePacket` directly), so a *delivered* handshake is always processed — the failure is non-delivery.

## Fix (minimal, configurable-by-default)

- The **initiator** re-sends its setup handshake every `handshakeRetransmitInterval` (default **1s**) while waiting for the reciprocal handshake, so a lost handshake recovers in ~1s instead of failing at 10s. A slow-but-alive path is unaffected — the reciprocal handshake closes `handshakeProcessed` and stops the retransmits immediately.
- The **responder** re-emits its handshake whenever it receives a *duplicate* forward handshake (which only happens because the initiator is still retransmitting), covering the case where the reverse ack itself was the lost packet.
- Setting `handshakeRetransmitInterval = 0` restores the legacy single-shot behavior (no migration shim; the new behavior is just the default).

The wait/retransmit is factored into a small pure helper (`awaitHandshakeAck`) that is unit-tested directly.

## Tests

- `awaitHandshakeAck`: stops on reciprocal handshake, responder path does no retransmit, and it retransmits several times before returning the context error on a dead peer.
- Route-group level: an initiator with no reciprocal ack repeatedly re-sends the real handshake packet over its transport; a responder re-acks on a duplicate forward handshake.
- Full `pkg/router` suite passes under `-race`.

## needs local-visor validation before merge (route-setup core change)

This is core routing. It must be live-validated on a running visor (a restart the maintainer approves) before merge — do **not** auto-merge.

### DIFF
--- a/pkg/router/handshake_retransmit_test.go
+++ b/pkg/router/handshake_retransmit_test.go
@@ -1,0 +1,168 @@
+// Package router pkg/router/handshake_retransmit_test.go
+package router
+
+import (
+	"context"
+	"net"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/skycoin/skywire/pkg/cipher"
+	"github.com/skycoin/skywire/pkg/routing"
+	"github.com/skycoin/skywire/pkg/transport"
+	types "github.com/skycoin/skywire/pkg/transport/types"
+)
+
+// countingTransport is a working transport that records how many times Write
+// was called, so a test can observe (re)transmitted handshake packets.
+type countingTransport struct {
+	closed   chan struct{}
+	writes   int32
+	pk1, pk2 cipher.PubKey
+}
+
+func newCountingTransport() *countingTransport {
+	pk1, _ := cipher.GenerateKeyPair()
+	pk2, _ := cipher.GenerateKeyPair()
+	return &countingTransport{closed: make(chan struct{}), pk1: pk1, pk2: pk2}
+}
+
+func (t *countingTransport) Write(b []byte) (int, error) {
+	select {
+	case <-t.closed:
+		return 0, net.ErrClosed
+	default:
+		atomic.AddInt32(&t.writes, 1)
+		return len(b), nil
+	}
+}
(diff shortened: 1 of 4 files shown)


0pcom:
Local validation complete on the running visor (build `173da3c5c`): 5-min soak, **244 route-group establishments to the Frankfurt exit, 0 `Remote handshake not received` drops**, 0 pump failures, mux held 5 transport-diverse legs @ 4.6 Mbps. `pkg/router` passes under `-race`; lint-clean on touched files. The old binary logged handshake-timeout failures in bursts; this binary had zero across 244 setups. Retransmit recovery path is covered by the deterministic unit tests. CI `linux/darwin/windows` red is pre-existing baseline debt (same on #4056/#4058, which merged), not a regression from this PR.

PR #4063, opened 2026-08-21
fix(router): self-heal failed-path exclusion for dead-edge route setup (#22)

0pcom:
Bounded-grace + self-heal failed-path exclusion (task #22). When a route-group setup handshake fails the full handshakeAwaitTimeout window (even with #4057 retransmit — e.g. all candidates funnel through a dead intermediary, observed live: skysocks→Frankfurt wedged 60s+, every candidate shared dead hop 038e009f, fresh route-find recovered in ~15s), the failed route's intermediate PKs are held on a per-destination TTL suspect list; the next dial excludes them via DialRoutes ExcludeIntermediatePKs → picks a different route instead of re-hammering the dead path. Self-heals on TTL expiry (no permanent blacklist); configurable via routing FailedHopExclusionTTL (0=default, on). New pkg/router/suspect_hops.go (+test). go build/vet/test ./pkg/router/ pass under -race (targeted, to avoid concurrent-golangci contention). NEEDS local-visor validation before merge (core routing); NOT auto-merged.

### DIFF
--- a/pkg/router/router.go
+++ b/pkg/router/router.go
@@ -181,6 +181,22 @@ type Config struct {
 	// query deliverer is configured (SetTransportQueryDeliverer); see
 	// rsn_oracle_routes.go. TPD is still used for routes with >=2 intermediates.
 	EnableRSNOracleRoutes bool
+
+	// FailedHopExclusionTTL configures the self-heal failed-path exclusion
+	// (dead-edge route-setup fix, the route-level-failover follow-up to the
+	// #4057 handshake retransmit). When a route's setup handshake fails to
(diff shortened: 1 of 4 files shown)


0pcom:
Superseded by #4078 (parallel candidate route-group setup), which landed the suspect-hop self-heal cache (newSuspectHopCache/suspectHopCache in parallel_route_config.go) as part of the steady-connection fix. This PR's parallel suspect_hops.go + FailedHopExclusionTTL implementation is the older competing version and no longer applies cleanly. Closing.

PR #4090, opened 2026-08-22
fix(proxystatus): show both directions of per-leg mux bandwidth (add recv bar)

0pcom:
The status.skysocks per-leg mux table had a sent column, a sent-share bar, and a recv column — but **no recv bar**, and the single bar was labelled 'bandwidth (sent share)'. Proxy traffic is recv-heavy (downloads) with tiny sent, so the only bandwidth *visual* tracked the near-zero sent direction — which reads as 'only showing sent bandwidth'.

Adds a recv-share bar scaled to its own per-direction maximum (scaling recv against maxSent would flatten a download leg to nothing), in a distinct colour, so both directions are visible at a glance. State colours (standby/warn) still win per leg.

### DIFF
--- a/pkg/proxystatus/render.go
+++ b/pkg/proxystatus/render.go
@@ -83,28 +83,40 @@ func writeMuxSection(b *strings.Builder, snap Snapshot) {
 			`A leg appears here once a route to a destination is warm.</p></section>`)
 		return
 	}
-	var maxSent uint64 = 1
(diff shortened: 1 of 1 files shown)


0pcom:
Superseded by #4106, which adds the recv-bandwidth bar (per-direction scaling — each of sent/recv scaled to its own busiest leg) plus per-leg route latency and the direct/multihop badge. Closing to avoid the conflicting duplicate.

PR #4145, opened 2026-08-23
tui: browse the command tree and its help interactively

0pcom:
**Did you run `make format && make check`?**

Partly: `goimports -local`, `golangci-lint` and `go vet` with the repo's
`withoutsystray withoutgotop` tags (both clean), and the tests for
`cmd/skywire/tui` and `pkg/flags`. Not the full suite or `check-cg`, neither of
which this touches.

**Fixes:** n/a — follows #4134

**Changes:**

`skywire --tui`, or `--tui` alongside any `--help`, opens a browser over the
command tree: subcommands on the left, the selected command's help on the
right, the code rain animated behind both. `↑↓`/`jk` move, `→`/`enter` opens a
group, `←` goes back, `PgUp`/`PgDn` scroll the help, `?` shows the keys, `q`
quits.

- It shows what is already there rather than describing it again. The tree is
  walked from the live `*cobra.Command`, and the right-hand pane is that
  command's own help captured as text — the same bytes `--help` prints,
  coloredcobra's colors included. Nothing here can drift out of step with the
  real CLI.
- Hooking help rather than adding a command covers both spellings for free: the
  root prints its help when run bare, and `--help` is a help call that already
  knows which command you meant — so `skywire cli visor --help --tui` opens on
  that command with its ancestors expanded.
- Every command is hooked, not just the root. Cobra passes the help function
  down only to commands that have none of their own, and skywire-cli's root
  installs one for `--json`; with the root alone,
  `skywire cli ... --help --tui` printed help instead of opening. Verified both
  ways under a pty.
- Read-only. Nothing here runs anything.

Off for everything that is not someone at a terminal — a pipe or redirect,
`--json`/`--jq`/`--shape`, `help -r`, `help -d` — and a browser that fails to
start prints the help that was asked for rather than swallowing it.

**Dependencies:** none new. bubbletea, bubbles and lipgloss are already direct
dependencies here, and this matches the six `tea.WithAltScreen()` programs
already in the tree. `flags.WithPlainHelp` is exported so anything capturing
help text rather than showing it gets it without a backdrop painted in.

**Tests:** nine, driving the model through its own `Update`/`View`, which needs
no terminal — cursor movement, expand/collapse, that the pane holds the
command's real help and follows the selection, that a frame is exactly the
terminal's height and no row exceeds its width, that the rain advances between
ticks, the quit keys, and the deepest-first ordering the hooking depends on.

**How to test this PR:**

```
go run . --tui
go run . cli visor --help --tui
go run . cli --help --tui
go run . --tui | cat        # prints help, does not open
go run . --json --help      # still JSON
```

### DIFF
--- a/cmd/skywire/commands/root.go
+++ b/cmd/skywire/commands/root.go
@@ -26,6 +26,7 @@ import (
 	dmsgprobe "github.com/skycoin/skywire/cmd/dmsg/dmsgprobe/commands"
 	scli "github.com/skycoin/skywire/cmd/skywire-cli/commands"
 	"github.com/skycoin/skywire/cmd/skywire/commands/doc"
+	"github.com/skycoin/skywire/cmd/skywire/tui"
 	services "github.com/skycoin/skywire/cmd/svc/skywire-services/commands"
 	"github.com/skycoin/skywire/pkg/buildinfo"
 	"github.com/skycoin/skywire/pkg/calvin"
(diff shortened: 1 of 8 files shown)


0pcom:
develop's Test workflow was red independently of this PR — since #4146 for a `behaviour` misspell and an errcheck `_ =` under `check-blank`, then a `modelling` misspell from #4149. Fixed in #4151 and #4153, both merged, so a rebase on develop drops those three inherited failures.

What remains is this branch's own, all in `cmd/skywire/tui`:

    tui.go:80:67   `colours`  -> colors
    tui.go:115:47  `coloured` -> colored
    tui.go:223:35  `coloured` -> colored
    tui.go:254:72  `coloured` -> colored
    tui.go:260:5   `colour`   -> color
    color_test.go:13:40, 15:43  `colour` -> color
    tui.go:210:8   G204: subprocess launched with potential tainted input (gosec)

Left alone since the branch is checked out in a live worktree.

Issue #4287, opened 2026-08-28
feat(router): RACK-TLP loss detection for the mux retransmit path


0pcom:
## Problem
`pkg/router/hol_retx.go` says it plainly: the SACK/retransmit path is "REACTIVE and slow — the sender's retxMinAge (750ms) presumes a missing seq is merely in flight." A fixed 750ms constant, plus a proactive-nudge heuristic (`gapAge > fastestLegRTT`), is exactly what RACK replaces. Multi-leg mux delivery is reordering **by construction**, so a fixed timer is structurally wrong.

## Proposal — RACK-TLP (RFC 8985)
- Record a **send timestamp per segment** (extend the retx buffer entries).
- Recompute each in-flight segment's expiry from the **latest RTT sample**, not a constant.
- Derive the **reordering window from measured reordering** (skywire reorders across legs continuously — measure it, don't guess).
- **TLP** probe to avoid an RTO tail.
- Fire the proactive HoL retransmit on **evidence** (reorder window elapsed for the frontier seq) instead of the timer.

## Fit
- **No wire change**: the retx buffer already holds unACKed frames; per-leg RTT already exists (`legCounters` EWMA + `ManagedTransport.GetLatencyStats`). This reuses `CapHOLRetx`'s existing SACK channel.
- **Composes with FEC (#4270)**: FEC is the coded fast-path (removes the frontier wait for coded blocks); RACK is the fast **uncoded fallback** for gaps FEC can't cover (>R erasures, or before repair arrives). Together: coded fast-path + evidence-based retransmit.

## Files
`pkg/router/hol_retx.go`, `pkg/router/route_mux.go` (retx buffer + per-seq send times), `pkg/router/route_group.go` (proactive SACK nudge). Standardized, deployed in Linux; highest value-to-effort of the current multipath backlog.

0pcom:
One correction to the plan, from reading the retx buffer — the first bullet is already done, and there is a different gap underneath it.

**`retxEntry` already carries a send timestamp**: `sentAt time.Time` (`sack.go:127`), set in `Add` (`sack.go:180`) and read at `sack.go:214` as `time.Since(e.sentAt) >= retxMinAge`. So no new field is needed.

[...]

PR #4298, opened 2026-08-29
feat(router): RACK-TLP tail-loss probe + DSACK reorder-window adaptation

0pcom:
Completes RFC 8985 loss recovery on the mux, on top of the RTT-derived retransmit threshold merged in #4295.

**Tail-Loss Probe (TLP).** The SACK path only recovers a hole once a later sequence arrives to expose it. If the *tail* of a burst is lost there is no later sequence — the receiver's bitmap simply ends at the last seq it got, the sender treats the missing tail as still in flight, and the stream stalls until the retx entry ages out. TLP re-sends the in-flight tail after an idle probe timeout (PTO ≈ 2×slowest-leg RTT), which either fills the loss or draws a SACK that finally reports the gap. Probe budget is bounded per stall and reset on ack progress; it reuses the existing SACK retransmit path.

**DSACK reorder-window adaptation.** When the receiver reports a duplicate (a DSACK — on these reliable, ordered legs almost always our own spurious retransmit), the sender widens its reorder factor so the RACK threshold waits longer before the next retransmit; clean acks decay it back toward the static baseline. The DSACK rides an optional trailing field on the existing SACK packet (flag+seq after the bitmap); a peer that predates it reads only the bitmap, so the wire stays backward compatible with no new capability bit.

Both are inert unless SACK was negotiated. Tests cover PTO derivation/floor/cap, TLP due-conditions and budget reset, grow/decay bounds, the DSACK→threshold closed loop, receiver-side duplicate detection, and SACK-packet round-trip + backward-compat.

### DIFF
--- a/pkg/router/route_mux.go
+++ b/pkg/router/route_mux.go
@@ -940,7 +940,56 @@ func (m *routeMux) processSACK(lastContig uint32, words []uint64) []uint32 {
 	if !m.sackEnabled || m.retxBuf == nil {
 		return nil
 	}
-	return m.retxBuf.ProcessSACK(lastContig, words)
+	return m.retxBuf.ProcessSACK(lastContig, words, m.rackThreshold())
+}
+
+// RACK-TLP retransmit-threshold bounds (RFC 8985 in spirit). Replaces the fixed
+// 750ms retxMinAge with a value derived from the live per-leg RTTs, so loss on a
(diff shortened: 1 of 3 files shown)


0pcom:
Superseded — this picked up a stale reused branch tip (the already-merged #4287/#4295 RACK-core commit) instead of the TLP+DSACK work. Reopening from a clean branch.

PR #4318, opened 2026-08-29
proxystatus: visually distinguish the two mux layers in the route tree

0pcom:
The skysocks proxy-status route tree stacks two independent kinds of route multiplexing — STREAM level (`--tunnels`: disjoint route groups to the exit) over PACKET level (mux legs striping one stream's packets across routes) — but nothing distinguished them, and with multiple streams the nested tunnel→leg layout also dropped each leg's left summary (state, route-rtt, bandwidth, share bars), since `bitree` only draws a left annotation for top-level spine routes.

With more than one stream the tree now renders each stream as a stream-boundary **header node** on the spine (`▚ stream N · k legs · mux on/off`) followed by that stream's legs kept top-level so their rich summaries render again. Each leg summary carries a per-stream accent **band** (`▏sN`), and a small five-colour palette cycles stably by stream index across the ~1s live WebSocket pushes. The per-stream accent composes with — never overrides — the existing active/standby (●/○) and hop-depth/exit PK coloring, and a legend above the tree names the two layers. Colours are chosen clear of the state/exit hues and re-darkened for the light theme to hold the page's contrast floor. A single stream renders clean with no stream chrome.

`skywire cli proxy tree` shares the same `RouteTree` model, so the headers and bands appear there too (bold-magenta header, monochrome bands). Verified with `go build .`, `go vet`, gofmt and `go test ./pkg/proxystatus/...`; a representative three-stream page was rendered to a static HTML file for eyeballing.

### DIFF
--- a/cmd/skywire-cli/commands/proxy/tree.go
+++ b/cmd/skywire-cli/commands/proxy/tree.go
@@ -102,12 +102,13 @@ func wantColor(mode string) bool {
 }
 
 const (
-	ansiReset  = "\x1b[0m"
-	ansiBold   = "\x1b[1m"
-	ansiDim    = "\x1b[2m"
-	ansiCyan   = "\x1b[36m"
-	ansiGreen  = "\x1b[32m"
-	ansiYellow = "\x1b[33m"
+	ansiReset   = "\x1b[0m"
+	ansiBold    = "\x1b[1m"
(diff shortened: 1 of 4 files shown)


0pcom:
Superseded by #4323 (restores the per-leg summaries the two-level tree dropped, keeps the grouping).

PR #4324, opened 2026-08-29
fix(router): clamp the warm-standby mux target so the self-heal can't storm

0pcom:
The adaptive preset requests a ~513-leg "full disjoint pool" Mux. On a real mesh no single exit offers hundreds of disjoint paths, so the background self-heal never reaches the target and re-dials toward it every tick — a setup-node dial storm that resets the route group and collapses bulk transfers mid-stream (measured: a 10 MB download rode 44+ churning legs at ~8x byte overhead and then stalled; subsequent transfers timed out).

Clamp the mux target to a reachable ceiling (`initialForegroundMux` = 16) at the dial actuation layer, independent of what the policy asks. The foreground dial fills the pool once and the self-heal holds it — **held standby legs are never reaped**, and the active width stays bounded by the adaptive engine (`adaptCap`). This fixes the storm for any policy without depending on the preset's Mux value.

Note: the adaptive preset's own Mux value (and its embedded `bundle.wasm`) is stale/divergent from source — a separate pre-existing wasm-parity issue tracked apart from this actuation-layer fix.

### DIFF
--- a/pkg/router/router_dial.go
+++ b/pkg/router/router_dial.go
@@ -660,6 +660,21 @@ func (r *router) finishDial(
 			muxTarget = eff
 		}
 	}
+	// Clamp the warm-standby mux target to a REACHABLE ceiling, independent of what
+	// the policy asks. The adaptive preset requests a very large Mux (its
+	// "full disjoint pool" ≈ 513); on a real mesh no single exit offers hundreds of
+	// disjoint paths, so the background self-heal never reaches such a target and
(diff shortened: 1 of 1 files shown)


0pcom:
Wrong lever — clamping the whole mux target discards the warm-standby reserve, which should stay large. The real fix is to bound the ACTIVE width (adaptCap) and let #4319 (exit honors the mirrored active set) propagate to the fleet exits, keeping the standby pool large. Superseded.

PR #4469, opened 2026-09-04
Experimental Go/wasm browser (SkywireGoBrowser) in the wasm visor

0pcom:
Adds an **experimental** alternative to netscrape's `browse.js`: a browser written in Go/wasm ([0magnet/shipyard `cmd/browser`](https://github.com/0magnet/shipyard/tree/main/cmd/browser)). Its chrome (tabs, address bar, back/forward, reload), page transcoding (renders in a sandboxed `srcdoc`, inlines stylesheets as `<style>` and images as `data:` URIs via a parent/sandbox postMessage relay), and navigation are all Go + `syscall/js`. Only the **network** is delegated, through the same `skywireVisor.fetchDmsg` / `fetchClearnet` the JS engine already uses.

### Non-breaking
netscrape's `browse.js` is untouched and still the default. This only *adds* a launcher.

### What it wires
- `pkg/wasmhv/browseui/gobrowser.wasm.gz` — the module (632 KB gz), inflated by `GoBrowserWasm()`, served at `/gobrowser.wasm` (both the native HV page and `hv serve`).
- `pkg/wasmhv/browseui/gobrowser-loader.js` — concatenated into the bundle; defines `globalThis.SkywireGoBrowser.open()`, which opens a winbox window, sets `globalThis.__shipyardBrowserFetch` (mesh hosts → `fetchDmsg`, else `fetchClearnet`), and instantiates the module against that window.
- Re-exports in `pkg/wasmhv/embed.go`; serving cases in `pkg/visor/hypervisor_handlers_browse.go` and `pkg/visor/wasmserve.go`.

### How to test
Open the visor page, then in the console:
```js
SkywireGoBrowser.open()
```
A window should open; type `example.com` (clearnet) or a `*.dmsg` host (mesh) in the address bar.

### Scope
Built without a running visor to test against; the likely spots needing adjustment to the live API are in `gobrowser-loader.js` (the `fetchDmsg`/`fetchClearnet` signatures and return shapes, and `WinBox().body`). This *adds* the Go browser alongside netscrape; it does not reimplement netscrape's full mesh-browser API (`SkywireBrowse.mountPanel`, pk-routing `browseTo`, the resolver). Native + wasm packages compile.


0pcom:
Live-tested on the standalone visor page (hv serve --tls, Brave). Two loader bugs found and fixed: open() required a page-realm Go class that only exists in the SharedWorker realm there (wasm_exec now loads on demand), and fetchClearnet was called as (url, method, body) against the actual (exitPK, method, url, body) signature, sending a null url. After the fixes: example.com renders over the clearnet path and home.dmsg over the mesh path, chrome and transcoding working.

Note on provenance: the PR ships only the prebuilt gobrowser.wasm.gz — the source is 0magnet/shipyard cmd/browser and is not vendored here, so the blob can't be rebuilt from this repo alone.

PR #4479, opened 2026-09-04
feat(wasm-visor): true VPN client — netstack tunnel instance + JS API

0pcom:
Part 2 of the wasm VPN client (part 1: #4478). The in-tab instance dials the vpn-server's route group with the same DialRoutes machinery the skysocks-lite instances use (so the packet-level mux/SACK dataplane carries it), runs the standard VPN handshake, and brings up the gVisor netstack as the TUN. vpnDial originates connections inside the stack — TCP and UDP (DNS included, resolved through the tunnel) egress at the vpn-server, which cannot tell this client from a native one.

pkg/vpn grows the embedding entry: NewClientEmbedded (no app env, no app RPC), ServeConn, TunReady. JS surface: vpnStart(pk) / vpnStop() / vpnStatus() / vpnFetch(url).

Live-validating on the desk tab against a fleet vpn-server before merge.


0pcom:
Live-validated in the Brave desk-rig tab (standalone visor page over https): vpnStart against fleet vpn-server 02ea1b80411ef28514056466bf11d8a7b7fcdcf4a964ebd874f8bd9fd74e81eda2 → connected (handshake + netstack up), then vpnFetch("http://ip.skycoin.com/") → 200 with ip_address 172.105.110.46 — the vpn-server's egress IP. DNS resolved through the tunnel (UDP :53), TCP originated inside the netstack. Two handshake fixes landed during validation: nil-gateway hello rejection and absent NIC enumeration on js.

PR #4517, opened 2026-09-04
fix(cli): give the direct-DMSG fetch a persistent identity

### LINKED ISSUE #4512: cli: tp net-stats --no-rpc mints a throwaway dmsg identity per invocation
`skywire cli tp net-stats --no-rpc` registers a fresh dmsg client identity on every run:

```
DEBUG [skywire-cli]: Trying direct DMSG fetch: dmsg://02b307aee5c8ce1666c63891f8af25ad2f0a47a243914c963942b3ba35b9d095ae:80/all-transports/stats
DEBUG [skywire-cli]: direct DMSG identity: pk=03bc093c3c63882cc16bfb9c8bad17edd406cba4aef55877f254b24c6dda001f11 source=ephemeral
```

Each of those keys lands in dmsg discovery and is then abandoned. This is the same pattern removed from `hv serve` in #4501 and fixed in the reward server in #4502, where a host had seeded discovery with hundreds of single-use identities. Anything that polls this subcommand on a timer multiplies it — a 60-second sampler is 1440 abandoned keys a day.

The default RPC path is the right one and works; it is only the `--no-rpc` fallback that does this. Where a local visor is reachable the fallback should not mint an identity at all, and where it genuinely has to run standalone it should reuse a persisted key rather than generating one per invocation.

Noticed while standing up a transport time-series recorder, which was switched to read the visor's CXO subscription over RPC instead — identity-free, and a whole object or nothing rather than a possibly-short HTTP read.

0pcom:
Closes #4512.

`resolveFetchIdentity` ended in `cipher.GenerateKeyPair()`, so every `--no-rpc` fetch ran under a brand-new keypair and handed it to a direct client on each production dmsg server. A 60-second sampler was 1440 one-shot identities a day — and it is per fetch, not per invocation.

Same shape as #4501 and #4502, different remedy, because there is no client here to reuse: the RPC path (the default, and correct) opens none, and the visor's own key is off limits — dmsg-discovery holds one entry per PK and a dmsg server one session per PK. So the CLI gets an identity of its own, written once to `<home>/.skywire/cli.key` (`SKYWIRE_CLI_KEYFILE` overrides; 0600; `O_EXCL` so racing invocations converge on one key) and reused. It slots in as step 5 of the existing precedence, ahead of the random key, which now appears only when the file cannot be read or written. Because the key is stable, the direct step also takes a mutex — two overlapping fetches under one PK would evict each other on the shared servers.

One correction to the issue text: this path builds its clients over an in-memory `direct.Client`, whose writes no-op, so the abandoned keys never reached dmsg-discovery (a lookup of the PK in the report 404s). What they did reach was every production dmsg server, one fresh Noise handshake and session apiece, per fetch.

Verified on the reported command, three consecutive runs:

```
before  pk=03bc093c3c63882cc16bfb9c8bad17edd406cba4aef55877f254b24c6dda001f11 source=ephemeral
        pk=03341a3c773ea1298c71eac8039053d2ad86c104d12c43563aa201ba6dd67de2eb source=ephemeral
        pk=0352e04bb2b8aca1da22f07ee40b9224cd283227c6c70a5f73b45e76aebb2d89d2 source=ephemeral

after   pk=02c48f70745ae06c02a4244a63aea8dd36e7c0faaa8e60ff02e7f8a1f8d8b8be63 source=cli keyfile
        pk=02c48f70745ae06c02a4244a63aea8dd36e7c0faaa8e60ff02e7f8a1f8d8b8be63 source=cli keyfile
        pk=02c48f70745ae06c02a4244a63aea8dd36e7c0faaa8e60ff02e7f8a1f8d8b8be63 source=cli keyfile
```

`cli sd proxy --no-rpc` and `cli ut --no-rpc` report the same key — the 46 subcommands on `RegisterFetchFlags` share the helper. Without `--no-rpc` nothing is minted and no keyfile appears; the RPC path stays identity-free. The generated `--sk` help line is updated in `docs/skywire/`.

### DIFF
--- a/cmd/skywire-cli/cliutil/identity.go
+++ b/cmd/skywire-cli/cliutil/identity.go
@@ -1,0 +1,117 @@
+// Package cliutil cmd/skywire-cli/cliutil/identity.go c4-vis-cli
+package cliutil
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/skycoin/skywire/pkg/cipher"
+)
+
+// CLIKeyFileEnv names the environment variable that relocates the CLI's
+// own dmsg identity file. Unset, the identity lives at
+// <user home>/.skywire/cli.key.
+const CLIKeyFileEnv = "SKYWIRE_CLI_KEYFILE"
+
+// cliKeyFileName is the basename of the CLI identity under ~/.skywire —
+// the same user-space directory the visor's own userspace config and the
+// skysocks-client MITM root already use.
+const cliKeyFileName = "cli.key"
+
+// CLIKeyFilePath returns where the CLI's persistent dmsg identity lives.
+func CLIKeyFilePath() (string, error) {
+	if p := os.Getenv(CLIKeyFileEnv); p != "" {
+		return p, nil
+	}
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return "", fmt.Errorf("locate user home dir: %w", err)
+	}
+	return filepath.Join(home, ".skywire", cliKeyFileName), nil
+}
+
+// CLIIdentity returns the keypair the CLI runs as when it has to hold a
+// dmsg client of its own, generating and persisting one on first use.
+//
+// A CLI invocation that talks dmsg directly needs *an* identity, and the
+// only wrong answer is a fresh one per run: every invocation then costs a
+// new keypair, a new Noise handshake against every production dmsg server,
+// and — on the code paths that publish — a new abandoned dmsg-discovery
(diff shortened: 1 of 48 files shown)


0pcom:
Closing on the maintainer's call: `--sk` already lets an operator supply a stable key when they want one, and the keys this would replace never reach dmsg-discovery (verified — the direct fetch path uses an in-memory `direct.Client`; the sampled key 404s and is absent from all 923 entries). The real cost is a Noise handshake per fetch, which is not currently causing trouble at the volumes involved.

A persistent keyfile across 46 subcommands is more machinery than that justifies. The genuinely registering leak is #4518, which is where the effort should go.

PR #4566, opened 2026-09-05
fix(visor): resolve the DMSG-D CXO peer from dmsg.discovery too

0pcom:
Found by turning `dmsg.lookup_cxo` on for the first time. The flag had no effect: the clients-by-server feed never appeared in `cxo status`, and every peer entry lookup still went over HTTP.

`dmsgdCXOPeer` read only `dmsg.discovery_dmsg`. On a dmsg-only deployment that field is unset and the discovery PK lives in `dmsg.discovery`, already in `dmsg://<pk>:80` form — so the feed spec always failed with "no DMSG-D CXO peer (dmsg.discovery_dmsg unset)" and the resolver silently degraded to HTTP on every lookup. The feature has been unreachable in the deployment it ships against.

`tpdCXOPeer` already gets this right — it tries `DiscoveryDmsg` then `Discovery` — which is exactly why the TPD feeds work and this one never has. `dmsgdCXOPeer` now does the same. `parseDmsgPeer` rejects a non-dmsg scheme, so a clearnet `discovery` is still correctly ignored; a test covers that.

Live evidence from the visor this was found on: `dmsg.discovery` = `dmsg://022e607e0914d6e7ccda7587f95790c09e126bbd506cc476a1eda852325aadd1aa:80`, `dmsg.discovery_dmsg` = null, and with `lookup_cxo: true` the `dmsgd-clients-by-server` feed was absent from `cxo status` while `tpd-stats`, `tpd-all-transports` and `sd-services` were all syncing normally.

### DIFF
--- a/pkg/visor/cxo_subscription_manager.go
+++ b/pkg/visor/cxo_subscription_manager.go
@@ -175,13 +175,26 @@ func sdCXOPeer(v *Visor) (cipher.PubKey, bool) {
 	return parseDmsgPeer(v.conf.Launcher.ServiceDiscDmsg)
 }
 
-// dmsgdCXOPeer extracts the DMSG-D publisher PK from the visor's
-// dmsg.discovery_dmsg URL.
+// dmsgdCXOPeer extracts the DMSG-D publisher PK from the visor's dmsg
+// discovery configuration.
+//
+// Both discovery_dmsg AND discovery are tried, in that order, mirroring
+// tpdCXOPeer. On a dmsg-only deployment the PK lives in `discovery` (already
+// `dmsg://<pk>:80`) and discovery_dmsg is simply unset — reading only the
+// latter meant the feed spec always failed with "no DMSG-D CXO peer", so
+// enabling dmsg.lookup_cxo silently did nothing and every lookup fell back to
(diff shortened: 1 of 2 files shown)


0pcom:
**Correction to the evidence in the description.** The fix is right, but the live case I cited was not the one that blocked it.

I had edited the wrong config file. The visor under test reads its repo-local `skywire-config.json`, where `discovery_dmsg` **is** set — so `dmsgdCXOPeer` resolved fine there and the feed was absent only because `lookup_cxo` had not been set in that file. Setting it in the correct config brought the feed up: `dmsgd-clients-by-server`, 140 paths, 1,028,011 bytes, `last_err=-`, syncing against the publisher (which logs `CXO clients-by-server publisher running dmsg_port=54` and `CXO warm: pre-populated publisher tree from store entries=924`).

The change still stands on its own: a config carrying only `discovery` (already `dmsg://<pk>:80`) with `discovery_dmsg` unset cannot resolve the peer, which is a real shape — it is what `~/skywire-config.json` on this machine looks like — and `tpdCXOPeer` already handles both fields for exactly that reason. It is a latent bug that would bite the next deployment written that way, not the cause of the failure I was chasing.

PR #4977, opened 2026-09-17
fix(router): a repair is charged to the leg that lost the frame; the writer parks only when no leg is both un-shed and has window room

0pcom:
Repairs ride the fastest live leg, and `resendSeqs` charged the retransmit counter to that carrier. Measured on a two-leg exit: the fast leg (100 ms) finished a 50 MB download phase with 1813 retransmits (1463 SACK + 368 HoL) against the slow leg's 9 — every one a repair of the *slow* leg's losses — so the coupled controller's loss ratio judged the healthiest leg the worst and parked it to standby, leaving the bloated leg (first hop 154 -> 408 -> 2400 ms) to carry the next phase alone. A repair now counts against the leg whose gap it fills (the sequence's last-send leg, read before the resend re-tags it); the bytes still count against the carrier, where they went out. Re-measured: no leg demoted, 867 of 909 repairs on the bloated leg (was 872 on the healthy one), and the healthy leg's ack delay stayed 314-407 ms instead of climbing to 1420 ms.

The park gate and the spill target are now separate questions. The gate parks only when no ready leg is both un-shed (live RTT within `ecfCongestRttFactor` x its own baseline) and holding window room. The spill — `FirstUnsaturated` and the pick fallback in `selectTransport` — is the first ready leg that is un-shed AND has room, else any ready leg with room, and -1 only when nothing has headroom. So a frame never lands in a full window, and the RTT shed brakes placement rather than stalling a writer.

Dropping the shed from the gate entirely was tried on the rig and measured worse: 86 % of a 50 MB download landed on the bloated leg (44.1 vs 7.1 MB) at 5.86 MB/s against 7.62 MB/s with the brake, and parking rose from 12.4 to 22 waits per download because both windows then filled. #4962's lone-ready-leg rule is unchanged, no new events or counters, and per-leg `window_bytes`/`inflight_bytes` (already on `MuxLeg`) are now pinned by a test. `go test ./pkg/router/ ./pkg/visor/ -count=1` passes.


0pcom:
Closing: two variants of this change were measured live against develop and both regressed the two-leg group (50 MB down 5.86 and 6.54 MB/s vs 7.51 on develop; results in bench/2026-09-16/991c1981f-smoke and 0b03e012e-smoke). The repair-attribution half is sound on its own evidence (no false demotion of the carrier leg); the park-gate half interacts with the RTT shed in a way the smoke does not reward. Kept as a reference branch on the fork.

PR #5012, opened 2026-09-18
fix(router): shared-bottleneck demotion is off by default and recorded as a ruling; the detector samples per-leg newly-acked frames, not the frontier purge

0pcom:
Four rig runs on 2026-09-16/17 (`bench/2026-09-16/195b1094c-sbdsweep/{sbd-off,sbd-default}`, `0251e5da4-smoke`, `dfb0755c2-smoke`) put the shared-bottleneck detector at 0-for-8. Parking off (`--sbd-min-samples 1000000` both ends) gave two-leg 50 MB downloads 9.30 MB/s, x1.13 paired, 5 of 5 above x1.09; parking on at its defaults 6.80 MB/s, x0.86; parks decided while the group was idle were permanent, x0.81. With the evidence floor in, three of four compose-set parks were refuted by their own trial (two on idle ticks between bench rows, trial rate 2 B/s) and the one that stood cost 10.8%.

It could not have been right. The per-SACK delay series was taken from every entry a SACK purges below its contiguous frontier, all aged against one `sampleNow` — that is how long the frontier took to advance, one number shared by both legs, so the two series were two views of the group's own head-of-line coupling and correlated whatever the network did. Only the frames a SACK NEWLY acks are per-leg samples now: the frontier-edge entry and the bitmap bits set above it, each aged from its own send timestamp on the leg it rode. The evidence floor is read on the side those samples come from — the send path's SACK-acknowledged bytes per tick — because judged by bytes received, the exit that sends a download looks idle for the whole transfer.

So this ships with `--sbd-demote` (`sbd_demote`, default **false**) gating the demotion only: the grouping still reaches the mux, and every ruling the detector would have acted on is recorded as an `sbd_ruling` mux event naming the legs, their summary statistics and the rate behind the reading. A trial tick below `--sbd-min-evidence-rate` now returns no verdict and the trial stays open; the idle-park probation path goes with it — that path is what re-admitted legs mid-transfer on int-L (exit `retx_sent` 1729, against 215 with the parks standing and 183 with SBD off). The RACK per-leg delay basis is kept: `max(send→ack delay, end-to-end pong RTT)` can only raise a hole's threshold, never lower one.


0pcom:
A sweep on the frozen two-leg rig (2026-09-18, `bench/2026-09-16/195b1094c-sbdsweep`, 50 MB downloads paired against the best single route) showed this detector ruling wrongly and expensively: with it parked out (`--sbd-min-samples 1000000` on both ends) the group ran 9.30 MB/s, x1.13, every row above x1.09; with the default floor of 4 the park landed and the same group ran 6.80 MB/s, x0.86. The endpoint's downlink measured 9.55 MB/s over three concurrent single-route clients that night, so those two legs — distinct intermediates, distinct first hops — were never behind one pipe.

So a park is now a trial with goodput as the arbiter: the group's aggregate delivered-bytes rate before the park is compared with the rate measured with the leg out, and a fall past the limit unparks the leg (overriding `leg_park_min_hold`) and marks that PAIR verified-independent, vetoed out of the clustering for a backoff window. A park that costs no goodput stands unchanged, and each failed trial emits a `park_trial_failed` mux event naming both rates.

New live knobs in the existing style: `--sbd-trial-window` (3 s), `--sbd-trial-loss` (0.15), `--sbd-backoff` (5 min, doubling per repeat, capped 1 h); `--sbd-min-samples` and `--sbd-sample-interval` and their defaults are untouched. Both ends run it, which matters because a download is exit-sent and the ruling was made at the exit.

0pcom:
Rebased on develop and added a third commit. On `bench/2026-09-16/0251e5da4-smoke/mux-legs-2` the detector parked a leg at 01:06:30.615 and again at 01:06:35.615 with the group carrying nothing, before the first row moved a byte; since a park trial cannot fail against a pre-park rate of zero, that park was permanent and all 15 rows after it ran single-leg — x0.81 on 50 MB and x0.68 on 10 MB, where the same route pair with parking off gives x1.13.

A demotion is now withheld unless the group is carrying `--sbd-min-evidence-rate` (new live route setting, default 64 KiB/s), and a park made with no baseline stays on trial until traffic arrives rather than standing. The second half: the peer mirrors a promote right after a park, so the first verdict found the leg already active and skipped the whole block, dropping the suppression that would have blocked the re-park — the refutation is now recorded whether or not there is still a park to undo.

`go test ./pkg/router/ -count=1` green, `golangci-lint run ./pkg/router/...` 0 issues; not yet run on the rig.

0pcom:
Pushed e3d8fdb33: with sbd_demote off the detector was still a park — enforceBottleneckGroups handed its clustering to the mux before the evidence floor was read, and rebuildWeights gives every non-representative member of a group zero send weight, which drove mux-compose-T2xL2 to upload shares of 49175:100%,49176:0% (2 tunnels x 2 legs collapsing to 2 x 1, x0.06-0.12) across 118 sbd_ruling events in 670 s, one of them on a 6 B/s send path. The floor is now read at the top of the tick (below it: no clustering, no weight change, no event), the grouping reaches the mux only when demotion is on, and the observational ruling is recorded at most once per leg pair per sbd_backoff window.

PR #5014, opened 2026-09-18
feat(skysocks): a spread policy — capacity-proportional chunk assignment, max share per route, min routes, endgame

0pcom:
Criterion 10 of the mux goal asks for a spread policy: cap any one route's share
of an object, hold a minimum number of routes, and still beat 0.8x the best
single reference. This adds the policy and the design doc behind it
(`docs/design/route-spread-policy.md`).

Chunk placement was `pickSessionFor(pickRecv)` — a per-stream rule that knows
nothing about the object its streams belong to, which is why a 3 MB/s tunnel
beside an 8 MB/s one drags the pair below the fast tunnel alone. `spreadPlanner`
is the per-object ledger that was missing: weight each tunnel by its measured
capacity (or equally, at the privacy end of the spectrum), skip a tunnel at or
above `spread.max_share` while another is under its cap, and give the chunk to
the tunnel furthest below its weight. `spread.min_routes` promotes standby
tunnels through the existing promoter and never dials; `spread.endgame`
duplicates the tail chunks on the fastest idle tunnel, one duplicate per chunk,
closing the loser's stream when the winner lands.

All four knobs default to today's behaviour (`max_share` 1.0, `min_routes` 0,
`endgame` false, `weight` rate), so a client that sets nothing places chunks
byte for byte as it does now — the planner still keeps its ledger, which is
where the object's `shares=<port>:<pct>` completion line comes from, keyed so it
can be matched against the bench's per-transport deltas.

Tested: planner tables (proportional 8:3 -> 73/27, cap 0.4 over three uneven
tunnels, even weighting, unmeasured tunnel, tail-only duplication, promote
path) plus two end-to-end runs of 50 MB down and 50 MB up over three tunnels
under `max_share 0.4 / min_routes 3`, both landing at 34/34/32. `go test
./pkg/skysocks/... -count=1` and `-race -run 'Spread|Chunk|Upload|Stripe|Plan'`
green; lint clean on the changed files. Not measured live yet — the bench set
is in the design doc.


0pcom:
The first live run of this policy (spread.max_share=0.4 spread.min_routes=3) spread the downloads over 3-4 routes but held them at 5.13 MB/s against an 8.23 MB/s best single-route reference (0.62x), while its uploads reached 0.91x — chunk.concurrency is one object-wide admission gate, so three routes just shared the same 8 streams, and min_routes promoted on the failover RTT rank, leaving the fastest standby in the pool.

fb1b90dad sizes the download gate per active tunnel once the policy steers (chunk.tunnel_concurrency, new, default 4, times the active width, measured after min_routes has promoted; chunk.concurrency unchanged when nothing is set), and makes min_routes promote the standby with the highest measured capacity in the object's direction, falling back to the RTT rank when nothing is measured.

0pcom:
e22c03e04: the planner's pick and its reservation are now one critical section — the burst the admission gate releases (chunk.tunnel_concurrency × active width) was resolving every pick against the same all-zero ledger because the charge landed after Open(), which is how row 4 of mux-spread-3 put 48.0 of 50 MB on the direct tunnel under max_share 0.4. A failed open now un-charges its reservation, the tie-break prefers measured capacity over session order, and two tests cover both (the concurrency one reproduces row 4 exactly when the lock is split).

PR #5037, opened 2026-09-18
fix(router): a leg whose delay basis is a multiple of its sibling's is fed a probe, not a proportional share

0pcom:
The download direction is not scheduled by ECF. Under CapUniDir the reverse pick is `selectByDirection`, and its tier 1 reads `ts.schedule` — which `transportSelector.Rebuild` fills with the live legs unweighted for every predictive mode. So a leg kept taking every other frame of a download however far behind it was, and neither the ECF hold-back predicate nor the per-leg send window ever saw the decision. Measured on the rig 2026-09-18 (composition 2x2): one group paired a 7.4 MB/s leg with a route whose own reference ran at 0.06 MB/s, the exit placed 0.5–2.2 MB of each 10 MB download on the slow one, and the five trials took 8.3, 15.0, 24.0, 31.9 and 37.4 s — each trial's duration is the bytes placed on that leg divided by its ~60 KB/s. The group's own detector had the number in the same run: `mean 755.9 vs 9291.3 ms over 8/8 samples`.

The window refresh now rules a ready leg probe-only when its end-to-end delay basis (the same `max(first-hop RTT, send→ack delay)` ECF and RACK judge a leg by) exceeds the best ready leg's by more than `--leg-starve-ratio`, and is at least 250 ms in absolute terms. Such a leg carries at most `--leg-probe-bytes` per window — the window being its own basis, floored at 250 ms — and the rest of the transfer goes to a leg that can take it. It is deliberately not a park: the leg keeps its rules and its probe, its basis decays once we stop queueing on it, and the ruling lifts by itself. Both terms are live knobs (`skywire cli route settings --leg-starve-ratio --leg-probe-bytes`, defaults 6.0 and 64 KiB), and each transition is recorded as a `leg_probe_only` / `leg_full_share` mux event.

The default is chosen to leave today's healthy pairs alone: the pinned legs-2 pair reads 44 ms against 166 ms by first hop (3.8x), and the two shared-bottleneck rulings the campaigns recorded read 172.7 vs 214.6 ms and 336.3 vs 513.9 ms (1.24x and 1.53x) — all under 6.0, and the first pair under the absolute floor as well. Unit tests in `pkg/router` cover the collapse (the 12x pair stays inside one probe budget over a 16 MB transfer), the healthy pairs keeping a full share, the ruling being reported and lifting, and the knobs' validation. Not merged yet — it wants a before/after rig row first.


0pcom:
Revised after chain AL. The delay-only criterion also fired on healthy legs: on legs-2 the 166 ms leg beside the 44 ms one was cut five times and the 50 MB download fell to x0.854 from x1.0-1.08. A send→ack delay basis is window/rate, so the queue inflating it is our own — that leg was delivering 27-65 % of the bytes while reading 6-7x.

A leg is now ruled probe-only only when it fails BOTH readings: delay basis more than --leg-starve-ratio times the best ready leg's, AND proven delivery (SACK-acked bytes per second, EWMA'd per refresh) under 1/that of the best leg's. The one knob serves both ends and the margins are wide either way — the dead leg read 12.3x on delay and 0.9 % on goodput. A leg with no ack of its own is never ruled; a leg that goes silent folds zeroes into its delivery EWMA, so a real stall is still caught.

Still wants a rig row before merge.

PR #5050, opened 2026-09-18
feat(router): a BBR-style path model per mux leg, and a scheduler that reads it

0pcom:
Every reading the mux scheduler judged a leg by was its send→ack delay basis, which is queue depth in disguise — a basis is window/rate, so our own queue inflates it. A loaded 166 ms leg beside a 44 ms one reads 6-7x while it is delivering a third of the bytes, and a leg queued so deep that it never acknowledges anything reads as the group's *fastest*. Each leg now carries the two quantities separately: BtlBw, a windowed max of its SACK-proven delivery rate (app-limited samples may raise it, never lower it), and RTprop, a windowed min of its out-of-band path RTT — with BDP and the honest queueing delay `inflight/BtlBw` derived from them, reported per leg in `mux info` and cited in the rulings.

The consumers read them where they meant capacity or propagation delay: the outclassed-leg ruling judges `RTprop + queue` instead of the basis (which collapses the healthy pair's ratio from 6-7x to ~1.3x) and BtlBw instead of a delivery EWMA that an idle moment pulls down; the ECF/OTIAS/STMS arrival estimate becomes `RTprop/2 + inflight/BtlBw`; the per-leg send window takes `path.bdp_gain x BtlBw x RTprop` as a floor under the delivery-proven window; and the acceptor's download places with the same predictive scheduler the forward path uses instead of an unweighted round-robin. The congestion shed deliberately keeps its queue-tracking baseline — RTprop there is "how full is the pipe", and it shed every busy leg. Everything falls back to the exact pre-model predicate on a leg the model has not measured, and `path.model=false` restores it wholesale; all thirteen new tunables are catalog knobs, settable live.

A leg sitting on bytes it has sent with none confirmed for several RTprops is now ruled stalled whatever its basis says: it carries no data while a sibling can take it, its stranded window is flushed onto a healthy leg at once rather than aging out of RACK one backoff at a time, and the retransmit pick skips it. Against the emulated testbed (#5046) the live-scale dead-leg scenario goes from not completing in 90 s (6.5/8 MB, wire 1.94) to 1.37 s — x0.85 of the good leg alone, 0 % payload share, wire 1.03 — still short of that scenario's 0.9x bar, which the remaining ~200 ms of cold-start placement before any reading exists accounts for; the one-in-five cut-leg wedge on develop no longer reproduces; and the healthy 44/166 ms pair still aggregates at x1.55-1.75 with neither leg ruled. Stacks on #5044.


0pcom:
Held out of v1.3.95: on the rig the delivery-rate scheduler moved the forward stream off the lowest-latency leg and uploads did not improve over develop (chain rows in bench/2026-09-18/). Revisit for v1.3.96 with a before/after row.

PR #5054, opened 2026-09-18
feat(skysocks): the spread policy ships on — max_share 0.4 over at least 3 routes

0pcom:
Makes criterion 10's measured values the catalog defaults: a default proxy session spreads each object over at least three routes and no route carries more than 40 % of its bytes. Endgame and the even weight stay off. Both values remain live knobs; `spread.max_share=1 spread.min_routes=0` restores the old placement.

Tests of the non-steering path now set the policy off explicitly. Merges only after the rig's spread row on this tree holds the 0.8x bar; that row goes in bench/2026-09-18/.


0pcom:
Held out of v1.3.95. As a default this collapsed on the rig (chain AT, bench/2026-09-18/a0df50285-smoke): 50 MB down x0.605 and up x0.157 of the paired reference, while the same knobs applied live on develop gave x1.46 / x0.95 (chain AS). min_routes 3 forces at least 20 % of the bytes onto an unmeasured third route from the first byte. Candidate remedies before this ships: a quality floor for routes that count toward min_routes, or a size threshold below which the floor is off.

